PR #11, opened 2019-10-17
Service query options

dwhitney:
This is working! There are tests in the lib/tests folder. The one thing to note is that I found storing `QueryOptions` in `SimpleEvaluator` problematic because a lifetime variable needed to be threaded through all of the types. I moved it to be a parameter to `PreparedQuery::exec`, and returned `base_iri` as a parameter to `SimpleEvaluator::new`, but it can be overridden with the `QueryOptions` passed to `PreparedQuery::exec`. I also incorporated all of the feedback from your previous comments. Let me know if you have any feedback

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -27,8 +27,9 @@
 //! assert_eq!(vec![quad], results.unwrap());
 //!
 //! // SPARQL query
-//! let prepared_query = connection.prepare_query("SELECT ?s WHERE { ?s ?p ?o }", QueryOptions::default()).unwrap();
-//! let results = prepared_query.exec().unwrap();
+//! let prepared_query = connection.prepare_query("SELECT ?s WHERE { ?s ?p ?o }", None).unwrap();
+//! let options = QueryOptions::default();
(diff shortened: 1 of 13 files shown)


Tpt:
Thanks! It looks good! I'm going to try pushing the `QueryOptions` back into the `prepare` method in an other PR.

Tpt:
Sorry, I forgot. Please use `cargo fmt` to format your code to keep the library style consistent. You could maybe add it a pre commit hook to make sure all your change are using it (some IDEs also have an option to apply it on file save). I just made a change to reformat your code.

Tpt:
`cargo clippy` is also nice to get suggestions on how to improve your code style.

dwhitney:
cool yeah I'll do that in the future.

Also FWIW, I think `PreparedQuery::exec` is not a bad place for the `QueryOptions` struct. It would be where things would go when using something like a Reader monad or any sort of dependency injection framework. It is annoying that the base_uri is an argument to the `PrepareQuery::new` and not simply injected, but that could probably be fixed. Either way, I think this is good.

PR #13, opened 2019-10-17
Custom functions

dwhitney:
This is my attempt at supporting custom functions. It is current with your service-improvements branch. 

This allows a query like the one below, which introduces a function `ex:REVERSE` that reverses a string:

```sparql
PREFIX ex: <http://example.com#>
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
SELECT ?name ?reverse
WHERE
    { 
      ?s foaf:name ?name .
      BIND(ex:REVERSE(?name) as ?reverse)
    }
```

It adds a field to the `QueryOptions` that allows one to add their own `CustomFunctionsHandler`, which requires `fn handle(&self, node: &NamedNode, parameters: &[Term]) -> Option<Term>` to be implemented.

The appropriate plan, plan_builder and eval steps are all implemented.

Thanks again for the great library and your feedback! I feel like I'm starting to get the hang of Rust, and I really like its speed!


lib/src/sparql/mod.rs
```diff
@@ -121,70 +139,125 @@ impl<'a, S: StoreConnection + 'a> SimplePreparedQuery<S> {
...
+///   }
+/// ```
+pub trait CustomFunctionsHandler {
+    /// Handles all custom functions for a given query
+    ///
+    /// For the example above node would have the value http://example.com#REVERSE
+    /// and would recieve one parameter with the value mapped to ?name
+    ///
+    /// Returns `None` if there is an error or the given NamedNode isn't recognized
+    fn handle(&self, node: &NamedNode, parameters: &[Option<Term>]) -> Option<Term>;
```
Tpt: I would make the `parameters` type be `&[Term]`.

1. It would make implementations easier (just need to check the parameter count and not that all parameters are set).
2. Other implementations like [RDF4J](https://rdf4j.eclipse.org/documentation/custom-sparql-functions/) and [Jena](https://jena.apache.org/documentation/query/writing_functions.html) does not seem to allow failed evaluations as parameters.
3. We probably want expressions like `ex:fn(xsd:dateTime("foo"))` to fail instead of calling `ex:fn` on `[None]`.
4. Cases where an argument is in a maybe bound variable could be handled like `IF(BOUND(?optArg), ex:fn(?optArg), ex:fn())`.

To provide optional parameters, one could just set them at the end of the parameter list like what is done with the build in functions (e.g. [REGEX](https://www.w3.org/TR/sparql11-query/#func-regex)).

Having a signature: `fn handle(&self, node: &NamedNode) -> Option<(fn (parameters: &[Term]) -> Option<Term>)>` could be interesting to be consistent with the one for services, allow to distinguish the case where the function does not exists from the one where its evaluation failed and allow to call `CustomFunctionsHandler.handle` once per query evaluation if the function name is already known and just call the returned closure for each tuple. 

It could also make sense to change the return type from `Option<Term>` to `Result<Term>`. This way, if we want to implement a "debug" mode in the evaluator, the evaluator could raise meaningful errors from the custom functions without breaking change to the trait.

dwhitney:
> Thank you! It looks great!
> 
> Some nitpicking about the `handle` interface.
> 
> A side question (I have no real opinion on it) about naming (also affects the ServiceHandler): the `handle` method returns a function that could be meaningfully called an "handler" for the given service/function. What about renaming `ServiceHandler` to `ServiceHandlerFactory` and `handle` to `get_handler` or `build_handler` or just `build` and same for `CustomFunctionsHandler`? It's just to improve the API, we could have this discussion after merging this PR.

It's been a long time since I've done any object oriented programming, and words like "Factory" and "get" give me the heebie-geebies :) I like pure functions, which is why I preferred just a straight closure to a trait. I'm not going to make the PR to change it, but I certainly won't complain :)

lib/src/sparql/mod.rs
dwhitney: > I would make the `parameters` type be `&[Term]`.
> 
>     1. It would make implementations easier (just need to check the parameter count and not that all parameters are set).
> 
>     2. Other implementations like [RDF4J](https://rdf4j.eclipse.org/documentation/custom-sparql-functions/) and [Jena](https://jena.apache.org/documentation/query/writing_functions.html) does not seem to allow failed evaluations as parameters.
> 
>     3. We probably want expressions like `ex:fn(xsd:dateTime("foo"))` to fail instead of calling `ex:fn` on `[None]`.
> 
>     4. Cases where an argument is in a maybe bound variable could be handled like `IF(BOUND(?optArg), ex:fn(?optArg), ex:fn())`.
> 

How are unbound values handled in rudf? I haven't run into them yet, but I assume this was represented by `None`. You can get these values with `OPTIONAL` clauses. Also it says in the SPARQL spec: [If the evaluation of the expression produces an error, the variable remains unbound for that solution but the query evaluation continues.](https://www.w3.org/TR/sparql11-query/#assignment), so it would seem to me at a minimum the `handle` function needs to return `Option<Term>`. If we follow your fourth suggestion, function composition couldn't really be done in this way: `ex:fn1(ex:fn2(1234))`. If unbound values are handled some other way, I guess it would work.

That aside, I don't think it's that bad ergonomically. Since the `handle` function returns `Option<Term>`, you can easily unbox parameters you expect to be there:

```rust
let param0 = paramaters[0].flatten()?; //assuming flatten is function on Option, 
let param1 = parameters[1].and_then(|p| p)?; // if flatten is something I imagined
```
That's some additional boiler plate, but not much worse than what it is without `Option`:

```rust
let param0 = paramaters[0]?;
let param1 = parameters[1]?;
```

> 
> To provide optional parameters, one could just set them at the end of the parameter list like what is done with the build in functions (e.g. [REGEX](https://www.w3.org/TR/sparql11-query/#func-regex)).
> 
> Having a signature: `fn handle(&self, node: &NamedNode) -> Option<(fn (parameters: &[Term]) -> Option<Term>)>` could be interesting to be consistent with the one for services, allow to distinguish the case where the function does not exists from the one where its evaluation failed and allow to call `CustomFunctionsHandler.handle` once per query evaluation if the function name is already known and just call the returned closure for each tuple.
> 

Ok, I can change this.

> It could also make sense to change the return type from `Option<Term>` to `Result<Term>`. This way, if we want to implement a "debug" mode in the evaluator, the evaluator could raise meaningful errors from the custom functions without breaking change to the trait.

This would require changing `eval_plan_expression`. Not really a problem but just an aside - that said see my comment about failed valued being unbound but the query continuing.

Tpt: > How are unbound values handled in rudf?

With `Option<EncodedTerm>`. It's why `EncodedTuple` is defined as `type EncodedTuple = Vec<Option<EncodedTerm>>`.

> so it would seem to me at a minimum the handle function needs to return Option<Term>.

Yes, I am not arguing about it, the custom functions needs to be able to fail.

> If we follow your fourth suggestion, function composition couldn't really be done in this way: ex:fn1(ex:fn2(1234)).

It could easily. The Rust evaluation code would look kinda like `self.custom_functions_handler.handle("ex:fn1", &[self.custom_functions_handler.handle("ex:fn2", &[1234])?])` (note the `?` after ex:fn2 evaluation).

My goal is mostly to follow what other SPARQL implementations are doing and the behavior of built in functions: if a function parameter evaluation returns an error (possibly because a variable is unbound), the function is not called and an error is returned instead. This way, we forbid the creation of "error caching" functions but it's probably better for interoperability with other implementations that do not allow it either. 

>> It could also make sense to change the return type from Option<Term> to Result<Term>. This way, if we want to implement a "debug" mode in the evaluator, the evaluator could raise meaningful errors from the custom functions without breaking change to the trait.
>
> This would require changing eval_plan_expression. Not really a problem but just an aside - that said see my comment about failed valued being unbound but the query continuing.

If we just make custom function return results, we don't have to change much: One could just write `self.custom_functions_handler.handle(name, &parameters).ok()` to get back an `Option<T>` from a `Result<T>` just like it's done with the numerous calls to Rust std library functions that returns `Result`.

Tpt:
> I like pure functions, which is why I preferred just a straight closure to a trait. I'm not going to make the PR to change it, but I certainly won't complain :)

Yes, functions are nicer. I was thinking of making clearly named traits and then having some wilcard implementation for functions.

For example we could write alongside `ServiceHandler` definition something like:
```rust
impl<F: FnMut(node: NamedNode) -> Option<(fn(GraphPattern) -> Result<BindingsIterator<'a>>)>> ServiceHandler for F {
  fn handle<'a>(
        &'a self,
        node: &NamedNode,
    ) -> Option<(fn(GraphPattern) -> Result<BindingsIterator<'a>>)> {
       self(node)
    }
}
```

This way people could implement `ServiceHandler` with a simple functions that will automatically implement the trait or build more complex implementations with a struct implementing the trait.

dwhitney:
Last night i sat down to use the `SERVICE` clause with reqwest as you suggested, and I ran into some issues that I think suggest maybe we ought to go with a simpler type signature on the handler functions. 

struct WebHandler;
        impl ServiceHandler for WebHandler {
            fn handle<'a>( &'a self, named_node: &NamedNode,) -> Option<(fn(GraphPattern) -> Result<BindingsIterator<'a>>)> {
                fn pattern_handler<'a>(graph_pattern: GraphPattern) -> Result<BindingsIterator<'a>> {
                    let iri = named_node.into_string();    
                    unimplemented!()
                }
            
                Some(pattern_handler)
            }
        }
```

error:

```
error[E0434]: can't capture dynamic environment in a fn item
   --> src/lib.rs:120:31
    |
120 |                     let iri = named_node.into_string();    
    |                               ^^^^^^^^^^
    |
    = help: use the `|| { ... }` closure form instead
```

It looks like fn pointers can't act as closures because they can't capture their environment. But using a closure instead requires that it be wrapped in a `Box` and cast to a fn pointer (which I have not been able to do successfully). We'd either need to change the type of the function in the handler to accept a closure instead of a function, or just do something simpler like make the signature:

```rust
    fn handle<'a>( &'a self, node: &NamedNode, pattern: GraphPattern) -> Option<Result<BindingsIterator<'a>>>; }
```

I like currying, but not if it's going to be painful. What do you think?

Tpt:
Hi! I am actually encountering the same problem right now while implementing the evaluation of the SERVICE test cases from the official SPARQL test suite.

You could not indeed cast to a `fn` pointer but it's possible to use the `Fn` trait with dynamic dispatch. It would look like `Box<dyn Fn(GraphPattern) -> Result<BindingsIterator<'a>> + 'a>`.
Note the second `'a` to state that the box content could not outlive `'a` (by default boxes are `'static`).

Using `fn handle<'a>( &'a self, node: &NamedNode, pattern: &'a GraphPattern) -> Option<Result<BindingsIterator<'a>>>; }` looks like a good simple solution to me. I would also take `GraphPattern` as a reference to avoid a clone during evaluation. Do you want me to write the change?

dwhitney:
heh I'm in the middle of writing the change, but I'm having trouble with `eval.rs` - probably better if you do it. I had to make node a value instead of a reference. FWIW, I'm not sure this would be much of an issue for me personally. When I use `SERVICE` clauses it's usually to load a fairly large amount of data once, so the performance impact would be near zero. I'd expect this to be fairly common since making many calls to `SERVICE` would likely have IO be the bottleneck. Either way - it's probably better if you handle the change

Tpt:
I just pushed the change on master.

I also fixed the `SILENT` behaviour: [the current tuple should not be discarded](https://www.w3.org/TR/sparql11-federated-query/#serviceFailure)

I also made the SPARQL Federation tests run: all passes but [one where the service name is retrieved from an other part of the query](https://w3c.github.io/rdf-tests/sparql11/data-sparql11/service/index.html#http://www.w3.org/2009/sparql/docs/tests/data-sparql11/service/manifest#service5)

dwhitney:
Hey I was out of town this weekend with my family. I'll address the feedback tonight so we can get this merged.

ktk:
@dwhitney any plans to revisit this so it could be merged in master?

saminahbab:
Hello, I was just wondering if this made it to master? I am not seeing any of the code for custom functions in master. It would be really helpful for me to have custom functions implemented, so was wondering the state of this issue.

Tpt:
@saminahbab Hi! Custom functions are not supported yet sadly. I should definitely do that.

saminahbab:
> 

Ah @Tpt  is it fair to say that the steps to add custom functions are defined in this issue by @dwhitney  and are relevant to the code currently in master? It looks to me from the files changed that a lot of the work seems done? 

Thank you for getting back to me

Tpt:
@saminahbab Yes, most of the work is done. But Oxigraph internals have changed quite a bit and I am not fully happy with the code of this PR. I plan to rewrite my own implementation.

Tpt:
I have just pushed an other attempt at custom functions: cfb3be6d3470f32be6c1b24ba42c017a2ead6945

Here is an example on how they work: https://github.com/oxigraph/oxigraph/blob/cfb3be6d3470f32be6c1b24ba42c017a2ead6945/lib/src/sparql/mod.rs#L164-L174

PR #23, opened 2020-04-02
Open some inners of the library

pchampin:
The goal is to provide other crates with access to low-level term creation.

This is required, e.g., when binding to other languages (e.g. Python, JS) or other APIs (e.g. Sophia).


Tpt:
Thank you!

PR #26, opened 2020-04-05
making unchecked constructor of Literal public

pchampin:
for the same reason as before: bindings to other languages/APIs can benefit from this


Tpt:
Thank you! Could you amend the method documentation to state that the language code must be in lower case? Oxigraph uses internally only lower cased language tags to make easier/faster SPARQL queries evaluation.

Tpt:
Thank you!

Issue #40, opened 2020-07-15
docker image


maxlath:
Dear Thomas,

To make it easier for people unfamiliar with the project dependencies to try oxigraph, I wrote a [Dockerfile](https://github.com/maxlath/docker-oxigraph/blob/master/Dockerfile) to generate images, and published those as [maxlath/oxigraph](https://hub.docker.com/r/maxlath/oxigraph) and [maxlath/oxigraph-wikibase](https://hub.docker.com/r/maxlath/oxigraph-wikibase).

If you're interested, I could give ownership of the [repo](https://github.com/maxlath/docker-oxigraph) to the oxigraph organization and republish the images as `oxigraph/oxigraph` and `oxigraph/oxigraph-wikibase`, what do you think?

Au plaisir

Tpt:
Hi Maxime!

It's a great idea. Instead of having a different repository for the Docker images, we could maybe just have them in this main git repository. This way, we could easily release new Docker images when a new release is done (maybe using GitHub CI).
What do you think about it?

maxlath:
Sounds great! I opened a PR in this sense #41

I'm not very familiar with GitHub CI: could it just run the `publish_docker_images.sh` script as such, given some credentials?

I didn't try Oxigraph for JavaScript yet, but will do ;)

Tpt:
I have published the Docker images to [Docker hub](https://hub.docker.com/orgs/oxigraph/repositories). I have tweaked the `Dockerfile` to build from the local code instead of doing a `git clone`. I have just to setup GitHub Actions to[ publish the images automatically](https://docs.github.com/en/actions/language-and-framework-guides/publishing-docker-images) when a release is done.

Tpt:
The publication is now automated with Github actions. Let's close this issue

PR #41, opened 2020-07-16
Adds docker-oxigraph

maxlath:
As suggested in #40

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,13 @@
+FROM rust:1.44 as builder
+
+RUN apt-get update && \
+  apt-get install clang -y && \
+  git clone https://github.com/oxigraph/oxigraph --depth 1 && \
+  cd oxigraph && \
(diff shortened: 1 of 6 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM rust:1.44 as builder
```
Tpt, line 1: It might be better to target the version `rust:1-buster` to get the latest stable version of Rust while ensuring we build on the same system as the exec image.

maxlath: done in 94f2bc4

maxlath:
I added a Dockerfile in both `server` and `wikibase` directory. As for the modifications to the readme, I'm not sure to get what you have in mind, but feel welcome to get the piece you want in the main readme and push on this branch

Tpt:
Thank you! I'll directly push my README changes into MASTER

PR #60, opened 2020-09-18
Server: Allow choice of storage backend with feature

theduke:


server/src/main.rs
```diff
@@ -441,7 +442,7 @@ mod tests {
     use crate::handle_request;
     use async_std::task::block_on;
     use http_types::{Method, Request, StatusCode, Url};
-    use oxigraph::RocksDbStore;
+    use oxigraph::Store;
```
Tpt, line 445: Sadly, it does not work because Rust is looking for `Store` in the `oxigraph` crate directly.

I believe this works:
```suggestion
    use super::Store;
```

Tpt (changes requested):
Thank you so much for this pull request!

There is just a problem with a `use` directive in a test.

server/src/main.rs
theduke: Note to self: always run the tests yourself. ;)

Tpt:
Thank you!

theduke:
As a side note: it would have been cleaner to make the code generic over `MutableDatastore` instead, but the `sophia` type signatures are pretty complex so I took the fast path.

Tpt:
I should maybe introduce a `MutableDatastore` like trait in Oxigraph that would allow to do that easily. I have not done it yet because I did not not figured out a way to do it in a nice way without [GAT](https://github.com/rust-lang/rust/issues/44265). But  I have simplified quite a lot the Oxigraph APIs in the past few month so it's maybe doable now. Having this trait might also help remove the code duplication between the different stores.

PR #171, opened 2021-11-26
Feture debug etiennept

etiennept:


Tpt:
Thank you very much for this pull request!

I am not planning to merge the rocksdb and v0.3 branches just yet. I would prefer for eeverything to be ready before that. Keeping master on the "0.2" line also allows me to easily do new 0.2 patch versions.

Issue #88, opened 2021-03-24
Authentication support


edmondchuc:
Hi, fantastic project!

It would be awesome to have authentication (and perhaps authorisation) support for the SPARQL and graph store endpoints in the Oxigraph server. For example, you can have the store as a public facing read-only SPARQL endpoint while the SPARQL update endpoint is protected with basic auth.

I would love to contribute to this project. I don't have any experience in Rust (but this project has inspired me to learn it). I mainly work in Python. 

Some options to implement basic auth:
- NGINX (easiest but probably not user-friendly in terms of deployment)
- Bundle a Python Flask/FastAPI proxy with a Next.js frontend in the Docker image of Oxigraph to handle and manage basic auth 
  - I can help with this, and implement role-based access control and create an admin dashboard similar to RDF4J and GraphDB
- Implement auth in Oxigraph server (in Rust) and use a Next.js frontend for the admin dashboard.

The chosen option will obviously affect future feature implementations such as authorisation. 

If we go with the latter two options, we can use a file-based JSON (or RDF) store. This makes it very easy to deploy and modify things for sys admins (following GraphDB's approach). 

Let me know your thoughts.

ansarizafar:
Authentication and authorization should be built in Oxigraph. I really like how Flureedb is doing authorization with smart functions (currently only clojure is supported but JavaScript support is planned)

Tpt:
Hi! Thank you so much for stepping in @edmondchuc @ansarizafar 

I see two different points here:
1. Authentication/authorization.
2. A nice admin dashboard frontend. It would be much welcome and a nice addition compared to [the current very limited JS interface](https://github.com/oxigraph/oxigraph/blob/master/server/templates/query.html).

About authentication/authorization, I currently use Nginx. This way, I could use all the Nginx plugins to do authentication, filtering, load balancing...
Personally, the feature I miss the most at the moment is the ability to limit the execution time of SPARQL queries to avoid overloading the server with resource-consuming queries.
I fear that adding authentication to Oxigraph means having to add a significant set of authentication providers (static user/password, LDAP...) even if there is currently with Oxigraph only three possible authorization levels: nothing, read and read-write.
I would be very curious about the practical use cases you have for authentication/authorization to see how well we could support them without having to build yet another authentication/authorization system. But if there are use cases where it makes sense to have authentication/authorization inside of Oxigraph server, I am definitely fine to have them integrated there.

[...]

PR #178, opened 2021-12-06
Fixed the Dockerfile and added basic authentication using nginx and docker-compose

vemonet:
Hi @Tpt, thanks a lot for this nice implementation of SPARQL in Rust! 

I met some issues when trying to run the latest source using Docker, I fixed them and added a few improvements to the docker deployment

- [x] I fixed the `Dockerfile` that was in the `server` folder, some dependencies were missing, and it needs to be in the root folder to be able to install the `oxigraph` crate from the local directory

- [x] Added a `docker-compose.yml` and `server/nginx.conf` to deploy with nginx authentication on `/update`, it has been mentioned in this issue: https://github.com/oxigraph/oxigraph/issues/88

It would be better if authentication was implemented directly in Rust, but in the meantime using a nginx proxy to add authentication only on `/update` can be convenient if someone wants to deploy the SPARQL endpoint publicly

- [x] Added a GitHub Action workflow that will build and publish the docker image to the GitHub Container registry at `ghcr.io/oxigraph/oxigraph` 

DockerHub now has pull rates limits that can be a pain when working in cluster, and it's convenient to have everything in the same codebase. The workflow will publish every push to the master branch as the `latest` tag, and every new release as the corresponding tag (`v0.1.2` will be published as tag `0.1.2`)

- [x] Updated the `server/README.md` to reflect the mentioned changes

Also, most commands where instructing to use `oxigraph_server load --file my_file.nq --location my_data_storage_directory` but the right command is actually `oxigraph_server --location my_data_storage_directory load --file my_file.nq` so I changed it

- [x] Added the volumes shared by docker in the `.gitignore`

Let me know if there are any changes you would like to do!


server/nginx.conf
```diff
@@ -0,0 +1,38 @@
...
+
+        location ~ ^(/|/query)$ {
+            # limit_except GET POST OPTIONS {
+            #         deny  all;
+            # }
+            proxy_pass http://oxigraph:7878;
+            proxy_pass_request_headers on;
+        }
+
+        location /update {
```
Tpt, line 27: You also might want to expose the `/store` path providing [the SPARQL graph store protocol](https://www.w3.org/TR/sparql11-http-rdf-update/)/

server/nginx.conf
```diff
@@ -0,0 +1,38 @@
...
+            # Auth: https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/
+            limit_except GET POST {
+                    deny  all;
+            }
+            auth_basic           "Oxygraph Administrator's Area";
```
Tpt, line 32: `Oxigraph` not `Oxygraph` (it's tricky for French native speaker ;-)).

Tpt (changes requested):
Thank you for the pull request! It's very nice to see contributors to Oxigraph.

> I fixed the Dockerfile that was in the server folder, some dependencies were missing, and it needs to be in the root folder to be able to install the oxigraph crate from the local directory

I don't think having the Dockerfile for Oxigraph server in the root directory is required to build the docker image. `docker build . -f server/Dockerfile` should work. Oxigraph git repository contains a lot of projects (the database as a Rust library, the python bindings, the JS bindings, the server...) so keeping build file in the matching subdirectory seems cleaner to me.

> Added a docker-compose.yml and server/nginx.conf to deploy with nginx authentication on /update, it has been mentioned in this issue:

Thank you! I am not sure providing them "built-in" and ready to use is the best thing to do. Oxigraph reusers are likely to want to customize them (e.g. change which path are allowed, which are not, add more containers to the Docker Compose file...). I would feel more confortable providing them as "examples" in the server README and have the compose file pulling a released version of the container to avoid people naively rebuilding the container on `docker-compose up`.

> Added a GitHub Action workflow that will build and publish the docker image to the GitHub Container registry at ghcr.io/oxigraph/oxigraph.

Thank you! It would be great to have the images pushed to the GitHub Container registry "nightly" indeed.

Thank you again!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,53 @@
+name: Publish Docker image
```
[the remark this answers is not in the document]
vemonet: I was hesitating to use buildx, but I got some issues recently with it that were resolved with the built-in docker in other projects. If it was already working well with buildx I'll put this though, it has a nice caching system built-in now

vemonet:
For the `Dockerfile` location I'll move it, I had some weird behavior with calling the parent folder even when in the context path just in GitHub Actions in the past, but I think it was not with buildx :)

> Thank you! I am not sure providing them "built-in" and ready to use is the best thing to do. Oxigraph reusers are likely to want to customize them (e.g. change which path are allowed, which are not, add more containers to the Docker Compose file...). I would feel more confortable providing them as "examples" in the server README and have the compose file pulling a released version of the container to avoid people naively rebuilding the container on docker-compose up.

My approach focuses on not needing to rebuild the existing `nginx` container, and just configure it at runtime directly with the `server/nginx.conf`, so there is no need to build, host and maintain an extra image (environmentally friendly!) 

It is also easier for the users to custom, as can they can edit the `nginx.conf` or `docker-compose.yml` before starting the docker-compose without needing to rebuild the nginx proxy image. And they don't need to fiddle with the different files paths (calling `nginx.conf` from the `docker-compose.yml`)

But I probably see things this way because I usually clone the repository

If you plan on users not needing to clone the repo then putting the `docker-compose.yml` file in the README makes sense, but we will need to publish also this image: add a `Dockerfile` and a specific workflow to only republish this image when the `nginx.conf` or `Dockerfile` change. Or just ask the users to copy the `nginx.conf` themselves alongside the `docker-compose.yml`, we have seen more dangerous operations but you never know with the users nowadays

Tpt:
> For the `Dockerfile` location I'll move it, I had some weird behavior with calling the parent folder even when in the context path just in GitHub Actions in the past, but I think it was not with buildx :)

Thank you! 

> My approach focuses on not needing to rebuild the existing `nginx` container, and just configure it at runtime directly with the `server/nginx.conf`, so there is no need to build, host and maintain an extra image (environmentally friendly!) 

That's great! Thank you! About <q>have the compose file pulling a released version of the container to avoid people naively rebuilding the container on docker-compose up.</q> I was talking about the Oxigraph container that is now rebuilt by compose, not nginx. Sorry.

> If you plan on users not needing to clone the repo then putting the `docker-compose.yml` file in the README makes sense, but we will need to publish also this image: add a `Dockerfile` and a specific workflow to only republish this image when the `nginx.conf` or `Dockerfile` change. Or just ask the users to copy the `nginx.conf` themselves alongside the `docker-compose.yml`, we have seen more dangerous operations but you never know with the users nowadays (also it makes it easier to custom, but you'll have a really long README)
> Let me know what would be your favorite approach for the docker-compose: publishing a new image or the user copy/paste everything from the readme?

I am more in favor of asking the users to copy/paste everything from the readme, including the `nginx.conf`. They should take time to customize the user and password anyway and think which piece of the server they want to expose externaly or not. They are also likely to want to enable https. So, making this process manual is fine imho.

> I'll make the changes when I find some times.

Amazing! Thank you!

.github/workflows/doc.yml
```diff
@@ -1,9 +1,11 @@
-name: doc
+name: doc and docker
 
 on:
   push:
     branches:
       - master
+  release:
+    types: [published]
```
Tpt, line 8: We are already building a Docker image on release (in the `release.yml` script). It might be better and more consistent to upload the same image on both Docker Hub and GitHub. Also one might not want to push to the "nightly docker track" images for release that might get built from some `vX` branch and not from the master branch. If you want I can add the upload to ghcr.io of "release images" to the `release.yml` script myself.

server/README.md
```diff
@@ -93,7 +94,101 @@ curl -X POST -H 'Accept: application/sparql-results+json' -H 'Content-Type: appl
...
+}
+```
+
+Then a `docker-compose.yml` in the same folder, you can change the default user and password in the `environment` section:
+
+```yaml
+version: "3"
+services:
+  oxigraph:
+    image: ghcr.io/oxigraph/oxigraph
```
Tpt, line 138: I would maybe use Docker Hub images that use the stable release instead of pushing people to use maybe broken nightly images.

server/README.md
```diff
@@ -93,7 +94,101 @@ curl -X POST -H 'Accept: application/sparql-results+json' -H 'Content-Type: appl
...
+    entrypoint: "bash -c 'echo -n $OXIGRAPH_USER: >> /etc/nginx/.htpasswd && echo $OXIGRAPH_PASSWORD | openssl passwd -stdin -apr1 >> /etc/nginx/.htpasswd && /docker-entrypoint.sh nginx'"
+```
+
+Once the `docker-compose.yaml` and `nginx.conf` are ready, start the Oxigraph server and nginx proxy for authentication on http://localhost:7878:
+
+```sh
+docker-compose up
+```
+
+To make an update to the graph, first change `$OXIGRAPH_USER` and `$OXIGRAPH_PASSWORD`, or set the environment variables, then run:
```
Tpt, line 166: Maybe make clear that the `curl` command is an example to show how auth work, not a mandatory step to make auth work?

.github/workflows/doc.yml
```diff
@@ -37,3 +39,47 @@ jobs:
...
+          # Change all uppercase to lowercase
+          IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+          # Use Docker `latest` tag convention
+          [ "$VERSION" == "master" ] && VERSION=latest
+          echo "IMAGE_ID=$IMAGE_ID" >> $GITHUB_ENV
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
```
Tpt, line 73: Is it really useful to have a big bash script for that? It might be more convenient to just use `ghcr.io/oxigraph/oxigraph:latest` in the `docker/build-push-action` action call.

Tpt (commented):
Thank you! I have some minor concernsbut it's looking great!

.github/workflows/doc.yml
Tpt: > The only thing that I am not sure and didn't really think of, is that in the end you will have 2 different images on 2 different repository: release on DockerHub and nightly on GitHub Container Registry, which could be slightly confusing (ideally you should just have 1 repository with 1 latest tag and all the release tags)

An option: have on DockerHub the release (`vX.Y.Z` tags and `latest` tag for the latest release) and in GitHub Container Registry the release (`vX.Y.Z` tags and `latest` tag for the latest release) and the nightly builds (`nightly` tag).
Using the `latest` tag only for release would make the users pull by default the latest stable release, it seems a bit safer to me than pulling the lastest nightly build.

vemonet: It makes perfect sense @Tpt , the user should get a stable release when pulling without providing a specific tag

I moved the gh release publish to `release.yml` (I re-used the step you had for the dockerhub build)

And I set the `doc.yml` to publish to the nightly tag, note that by default the [docker-metadata action](https://github.com/docker/metadata-action#basic) would use the `master` tag, that might be more semantically accurate, but I am not sure what is the usual convention for this

I currently enabled the github action caching  for the `nightly` tag action, it can be quite good to reduce regular builds time, but it's still experimental, so you might want to disable it if you are facing issues. I added a comment with a link to the doc about it in the action file

Tpt:
Thank you!

PR #190, opened 2022-02-26
Adds a web build to Oxigraph JS

Tpt:
Builds an hybrid web/node NPM module able to be used both in Node.JS and on modern web browsers.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -136,6 +136,19 @@ jobs:
           name: pyoxigraph_wheel_x86_64_windows
           path: target/wheels/*.whl
 
+  npm_tarball:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: rustup update
+      - run: cargo install wasm-pack
+      - run: npm run pack
+        working-directory: ./js
+      - uses: actions/upload-artifact@v2
+        with:
+          name: oxigraph_wasm_npm
+          path: js/*.tgz
+
   docker:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


bollwyvl:
Awesome!

Selfishly: I know there are 10 million web build toolchains, but It would be lovely to have a minimum-viable webpack 5 example in the test suite, as these aren't fun to troubleshoot without something canonical that works. 

A longer, and even-crazier, con would be to get `pyoxigraph` into the game as well: having it plug into the [pyodide](https://github.com/pyodide/pyodide/) ecosystem would make for a very interesting experience, a la the [numpy website](https://numpy.org)... it looks like all of the c-native parts of the `rdflib` stack already work today, but  [rust is another story](https://github.com/pyodide/pyodide/issues/1973).

Nearer-term, it would probably be possible to wire up oxigraph.js as a custom jupyterlite kernel, if you were so inclined...

Tpt:
@bollwyvl Thank you!

> Selfishly: I know there are 10 million web build toolchains, but It would be lovely to have a minimum-viable webpack 5 example in the test suite, as these aren't fun to troubleshoot without something canonical that works.

I just tried WebPack support. It works well and I have updated the README about it. I need to figure out a proper way to run web browser tests in GitHub CI.

> A longer, and even-crazier, con would be to get pyoxigraph into the game as well: having it plug into the [pyodide](https://github.com/pyodide/pyodide/) ecosystem would make for a very interesting experience, a la the [numpy website](https://numpy.org/)... it looks like all of the c-native parts of the rdflib stack already work today, but https://github.com/pyodide/pyodide/issues/1973.

It would be amazing but that's gonna be "fun". pyoxigraph depends on RocksDB that depends on a file system being available and is likely to not be easily compatible with WASM. So, it won't be easy but it would be amazing if it works indeed!

> Nearer-term, it would probably be possible to wire up oxigraph.js as a custom jupyterlite kernel, if you were so inclined...

I believe someone already started to work on it but I don't remind who/where. Sorry.

PR #223, opened 2022-07-28
Add tests for pyoxigraph parse and serialize functions

edmondchuc:


python/tests/test_parse.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pyoxigraph import *
+
+
+class TestParse(unittest.TestCase):
+    def test_parse(self):
+        input = io.BytesIO(b'<foo> <p> "1" .')
+        result = list(parse(input, "text/turtle", base_iri="http://example.com/"))
+
+        assert result == [
```
Tpt, line 12: I believe it's better to use `self.assertEqual` insert of `assert` in order to get a better integration with `unittest`.

python/tests/test_serialize.py
```diff
@@ -0,0 +1,24 @@
...
+                    Literal("1"),
+                )
+            ],
+            output,
+            "text/turtle",
+        )
+
+        assert (
+            output.getvalue() == b'<http://example.com> <http://example.com/p> "1" .\n'
+        )
```
Tpt, line 24: Same here

Tpt (commented):
Thank you so much for your PR.

I would maybe merge the test_parse and test_serialize files into a test_io file to avoid very small files. But it's a very personal preference so feel free to ignore it.

python/tests/test_parse.py
edmondchuc: Ah yes, thank you. I'm too used to writing pytest tests which uses `assert`, oops.

edmondchuc: I've moved the tests into a test_io file as well.

Tpt (approved):
Thank you!

Indeed pytest is much nicer and convenient. Because of the small number of tests I have been lazy and used unittest in order to avoid extra dependencies but moving to pytest might make a lot of sense in the future.

PR #224, opened 2022-07-28
Add pyoxigraph type annotations

edmondchuc:


edmondchuc:
I just tested pyoxigraph again in vscode and it is now autocompleting. I think when I previously tested it last week, my python interpreter must have been misconfigured with vscode. Now that I know that both autocompletion and docstrings are showing for both vscode and pycharm, I'd say we should hold off merging the type annotations until we fix stubdoc's issues with pulling all the docstrings out of pyoxigraph.
Reason for holding off is, despite type annotations being very useful, I think having the docstrings show in IDE is the better tradeoff.

Tpt:
I believe we can close this PR now that type annotation are automatically built and push

PR #225, opened 2022-07-31
Upgrades RocksDB to v7.4.5

Tpt:


Tpt:
The Windows CI failure is caused by https://github.com/facebook/rocksdb/issues/10455 that does not affter RocksDB 7.3 for some unknown reasons.

PR #226, opened 2022-08-01
Fixes ReadTheDocs build

Tpt:
Adds a shared requirements.txt file for the documentation dependencies


edmondchuc:
The devcontainer's `Dockerfile` can probably be updated as well.

https://github.com/oxigraph/oxigraph/blob/4902bac3a788a0acf40587316859da5439392f9e/.devcontainer/Dockerfile#L18-L22

Issue #218, opened 2022-07-20
Load/parse RDF string data directly into store


milvala:
Hi

I was wondering whether there's a way to directly load/parse RDF data in Python `str` format (cf. the [rdflib parse method](https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.html?highlight=add#rdflib.graph.Graph.parse), which has multiple input options)? 

I'm currently using the following method, but it seems a bit convoluted:

```
rdf_data_io = io.BytesIO(rdf_data.encode())
store1.bulk_load(input=rdf_data_io, mime_type="application/n-triples")
```

It would be nice to be able to just provide `rdf_data` in `str` format, without using an intermediate conversion to an io object.

Tpt:
Hi! Thank you for trying Oxigraph. It's now possible starting from Oxigraph 0.3.0. I have not added an example in the documentation yet, it's something I should fix.

```python
store = Store()
store.bulk_load(rdf_data.encode(), "text/turtle)
```

milvala:
Hi @Tpt 

Thanks for your quick reply and no problem! I wasn't sure whether I was overlooking something or whether it was not possible yet :)

Tpt:
A maybe bad idea: deprecate the `load` method and add two new methods `load_from_data` and `load_from_file`, the first one taking for input the serialized data (`io.BytesIO`, `str`, `bytes`...) and the second one the file path. What do you think about it?

PR #227, opened 2022-08-01
Add `Store.load_from_stream` and `Store.load_from_data` methods

edmondchuc:
Fixes https://github.com/oxigraph/oxigraph/issues/218

This PR implements two new methods to the `Store` class as suggested in: https://github.com/oxigraph/oxigraph/issues/218#issuecomment-1193384035 along with tests.

### DIFF
--- a/python/src/store.rs
+++ b/python/src/store.rs
@@ -7,9 +7,34 @@ use oxigraph::io::{DatasetFormat, GraphFormat};
 use oxigraph::model::{GraphName, GraphNameRef};
 use oxigraph::sparql::Update;
 use oxigraph::store::{self, LoaderError, SerializerError, StorageError, Store};
-use pyo3::exceptions::{PyIOError, PyRuntimeError, PyValueError};
-use pyo3::prelude::*;
+use pyo3::exceptions::{self, PyIOError, PyRuntimeError, PyValueError};
+use pyo3::types::{IntoPyDict, PyBytes, PyString};
+use pyo3::{prelude::*, PyTypeInfo};
 use pyo3::{Py, PyRef};
+
+struct InputValue(String);
+
+impl<'source> FromPyObject<'source> for InputValue {
+    fn extract(ob: &'source pyo3::PyAny) -> PyResult<Self> {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        if PyString::is_type_of(ob) {
+            Ok(InputValue(ob.to_string()))
+        } else if PyBytes::is_type_of(ob) {
+            let kwargs = vec![("encoding", "utf-8")].into_py_dict(py);
+            Ok(InputValue(
+                ob.call_method("decode", (), Some(kwargs))
+                    .unwrap()
+                    .to_string(),
+            ))
+        } else {
+            Err(exceptions::PyTypeError::new_err(format!(
+                "'{}' type is unsupported.",
+                ob.get_type().name().unwrap()
+            )))
+        }
+    }
+}
 
 /// RDF store.
 ///
@@ -322,7 +347,7 @@ impl PyStore {
         })
     }
 
-    /// Loads an RDF serialization into the store from a stream.
+    /// Loads an RDF serialization into the store from a file.
     ///
     /// Loads are applied in a transactional manner: either the full operation succeeds or nothing is written to the database.
     /// The :py:func:`bulk_load` method is also available for much faster loading of big files but without transactional guarantees.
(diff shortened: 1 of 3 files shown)


python/src/store.rs
```diff
@@ -322,6 +322,163 @@ impl PyStore {
...
+    /// :type input: str
+    /// :param mime_type: the MIME type of the RDF serialization.
+    /// :type mime_type: str
+    /// :param base_iri: the base IRI used to resolve the relative IRIs in the file or :py:const:`None` if relative IRI resolution should not be done.
+    /// :type base_iri: str or None, optional
+    /// :param to_graph: if it is a file composed of triples, the graph in which the triples should be stored. By default, the default graph is used.
+    /// :type to_graph: NamedNode or BlankNode or DefaultGraph or None, optional
+    /// :raises ValueError: if the MIME type is not supported or the `to_graph` parameter is given with a quad file.
+    /// :raises SyntaxError: if the provided data is invalid.
+    /// :raises IOError: if an I/O error happens during a quad insertion.
```
edmondchuc, line 433: @Tpt would this still raise an `IOError` anywhere?

edmondchuc:
@Tpt in https://github.com/oxigraph/oxigraph/issues/218#issuecomment-1193384035 you mentioned deprecating `Store.load()`. What should this entail? Log some warnings of the deprecation?

edmondchuc:
**Action:** 
- [ ] Instead of `load_from_stream`, create a method named `load_from_file` which accepts a file path. 
- [ ] Change `load_from_data` to accept a python I/O object or a `str`/`bytes` buffer.

Tpt:
An other naming idea: the shorter `load_file` and `load_data`

Tpt:
An idea to avoid code duplication: create a shared method not exposed with pyo3 that takes for input type `impl BufRead + Send` and use it in the three implemented methods. But it's likely to be a bit tricky so feel free to ignore this idea and let me attempt it in a follow-up change.

edmondchuc:
Hi @Tpt I've added my implementation of `load_file` and `load_data`. The former is the same code as the current `load` method and the latter accepts input as either a `str` or `bytes`. 

I am not sure if what I have implemented is idiomatic rust/pyo3. Please let me know if it should be done differently.

edmondchuc:
> An idea to avoid code duplication: create a shared method not exposed with pyo3 that takes for input type impl BufRead + Send and use it in the three implemented methods.

I am keen to attempt this if what I've done so far makes sense.

python/src/store.rs
Tpt: No, it won't.

Tpt:
> @Tpt in https://github.com/oxigraph/oxigraph/issues/218#issuecomment-1193384035 you mentioned deprecating Store.load(). What should this entail? Log some warnings of the deprecation?

I believe we want to use Python deprecation warning system. The PyO3 library seems to expose it using [`PyErr::warn`](https://docs.rs/pyo3/latest/pyo3/prelude/struct.PyErr.html#method.warn) but the documentation is a bit lacking in this area. It's something we maybe should investigate in a follow up PR.

> Hi @Tpt I've added my implementation of load_file and load_data. The former is the same code as the current load method and the latter accepts input as either a str or bytes.
> I am not sure if what I have implemented is idiomatic rust/pyo3. Please let me know if it should be done differently.

Thank you so much! The code looks good. I would maybe move the Python I/O objects support to the `load_data` method because it's more "data" than a "file".

> I am keen to attempt this if what I've done so far makes sense.

That would be amazing. It might also help with the moving of I/O objects to `load_data` easier by allowing to write code like `if is_string { self.load_inner(the str) } else { self.load_inner(the_io_object) }` so I would suggest to do this refactoring first. But if you find it too hard, I can easily take some time to make an attempt. Rust generics are not the easiest thing (it does not have the C++ templates footguns but making the compiler happy with what ones write often takes some efforts).

edmondchuc:
> Thank you so much! The code looks good. I would maybe move the Python I/O objects support to the load_data method because it's more "data" than a "file".

Awesome, I've added support for Python I/O objects to `Store.load_data`. This includes `RawIOBase`, `BufferedIOBase` and `TextIOBase`. `TextIOBase` is a new addition. I've also added tests for these I/O object types.

> That would be amazing. It might also help with the moving of I/O objects to load_data easier by allowing to write code like if is_string { self.load_inner(the str) } else { self.load_inner(the_io_object) } so I would suggest to do this refactoring first. But if you find it too hard, I can easily take some time to make an attempt. Rust generics are not the easiest thing (it does not have the C++ templates footguns but making the compiler happy with what ones write often takes some efforts).

I fumbled around with the code, still a bit uncertain what I'm doing in rust, but I had a go. It deviates a bit from what you suggested. Let me know if there are any issues with it.

> An idea to avoid code duplication: create a shared method not exposed with pyo3 that takes for input type impl BufRead + Send and use it in the three implemented methods.

I wasn't able to use `impl BufRead + Send`. Apparently python functions can't have generics, unless I misunderstood something. 

The result was a unified method called `Store.load_inner` which accepts `input` as `&[u8]`.

Tpt:
>>I wasn't able to use `impl BufRead + Send`. Apparently python functions can't have generics, unless I misunderstood something.
>
> The result was a unified method called `Store.load_inner` which accepts input as &[u8].

Thank you so much for giving it a try. We are moving forward! Indeed, python functions can't have generics, my suggestion was to build a function not exposed to python (i.e. not in the `#[pymethods]` impl block). I am also not sure converting the input to string is a good approach in all cases: what if the user wants to parse a multiple-gigabytes dump? The new code loads it all into memory while the old one was able to parse it in streaming.

If you want you could revert a bit your change and I can give it a try.
Thank you again!

edmondchuc:
> I am also not sure converting the input to string is a good approach in all cases: what if the user wants to parse a multiple-gigabytes dump? The new code loads it all into memory while the old one was able to parse it in streaming.

Yes, I agree completely!

> If you want you could revert a bit your change and I can give it a try.


Yeah sure, go for it! I'm keen to learn the correct way. I believe you can push directly to this branch, feel free.

Tpt:
The new 0.4.0-alpha.1 version provides this feature: `load("my_string")` will parse `"my_string"` as the file content whereas `load(path="file.ttl")` will parse the `file.ttl` file.

Issue #268, opened 2022-10-25
SPARQL AST editing in spargebra


hobofan:
We have internally implemented a few extensions to `spargebra`, to allow us in building and composing SPARQL queries, that we could possibly upstream.

@Tpt I'm wondering if that is something that you would like to be a part of `spargebra`? If so, where would be a good place to put it? (A new module`spargebra::algebra::edit`?). I could also understand if you feel that it's best kept separate as that is possibly an endless topic.

One example would be the `RenameVariable` trait:
```rust
/// Trait that allows to recursively rename all occurrences of a variable.
pub trait RenameVariable {
    /// Replaces all nested occurrences of the variable `old_name` with the variable `new_name`.
    fn rename_variable(&mut self, old_name: &str, new_name: &str);
}
```

Tpt:
Nice! Thank you!

Yes to include well defined operations in spargebra. If it's general use like variable renaming it might even be nice to include them as inherent method instead of traits. I am planning to implement a query optimiser inside of spargebra or an other related crate and having manipulation methods would be very convenient.

PR #274, opened 2022-10-31
Adds custom SPARQL operator OX_LATERAL

### LINKED ISSUE #267: Implement Jena-like `loop:` "service" to support lateral joins
We are currently trying to use oxigraph for a project where GraphQL-like query structures are transformed into SPARQL queries. 

E.g. a typical high-level query for that would be "Query 10 of A (filtered by X), and for each of them query 5 of B"

However we are running into the problem that SPARQL currently doesn't support "lateral joins" (see this thread for deeper explanation: https://github.com/w3c/sparql-12/issues/100), which means that the required "10 of" isn't correctly expressable as a LIMIT in a SPARQL query.

I think out of the solutions proposed in the linked thread, the `loop:` service enhancer that Jena recently added (https://jena.apache.org/documentation/query/service_enhancer.html), is one of the better options.

I did already try to implement that functionality for oxigraph for some time now, but after trying out a few different things, I'm not sure how it would be best implemented or if that would even be possible with oxigraph's current architecture. Any pointers on the topic would be greatly appreciated! :)

Tpt:
"FOO OX_LATERAL(?v1 ?vn) SUBSELECT" means that for all bindings emitted from FOO SUBSELECT is going to be called with ?v1 ?vn already set from the binding from FOO

For example:
```sparql
SELECT ?s ?o WHERE {
    ?s a ex:T .
    OX_LATERAL(?s) {SELECT ?o WHERE { ?s ex:p ?o } ORDER BY ?o LIMIT 2}
}
```

Closes #267


hobofan:
I'm not sure what the currently assumed state of stability of a query structure is when serializing and deserializing it.

Until now we've constructed the query as a `spargebra` `Query` during our higher-level query planning layer, serialized it as a `String` in the end, and then parsed the query again when passing it into `Store::query`. So essentially a roundtrip of query->string->query.

When testing `OX_LATERAL` with a query where I constructed a `GraphPattern::Sequence`, that is gone and replaced by a `GraphPattern::Join` after the roundtrip from what I can tell, resulting in a different query result. If you want I can try to produce a reduced test for that.

As a "workaround"(/ more desirable solution 😅 ) I've eliminated the roundtrip (#276).

----

With that workaround in place, I was able to integrate the feature in the first places, and it's looking good so far! :)

Tpt:
> I'm not sure what the currently assumed state of stability of a query structure is when serializing and deserializing it.

It's bad. I have just merged #270 that makes sure the serialization output can be parsed. But in the future we should definitely ensure that serialization -> deserialization ends up with the exact same algebra, at least in common cases.

> As a "workaround"(/ more desirable solution :sweat_smile: ) I've eliminated the roundtrip (https://github.com/oxigraph/oxigraph/pull/276).

Thank you!

> With that workaround in place, I was able to integrate the feature in the first places, and it's looking good so far! :)

Amazing!

hobofan:
As promised, further feedback:

We've now integrated it into all places we wanted for now in our codebase, and seems to be working without issues! :)

Generally, we feel like composing lateral queries via spargebra puts quite a bit more stress on the tools available for composing queries, and required us to write a few utilities for that (e.g. for correctly creating an intermediate `GraphPattern::Projection` for any `GraphPattern`, so that we can convert that into a lateral join). I'll try to see that I can generalize some of those into generic utilities for query building to upstream them (see #268).

Tpt:
Hi! Thank you very much for the feedback. It's great!

About the creation of intermediate `GraphPattern::Projection`, if you know that the composition is safe (i.e., no possible variable conflict), you might just use `GraphPattern::Sequence` without a nested `Projection`. It will avoid you the hassle of generating the projection. But beware, it is likely that the serialization back to SPARQL of such queries does not work properly.

hobofan:
One thing that I've been stubbing my toe on now is the join behaviour of `GraphPattern::Sequence`. As far as I can tell it will internally generate a `ForLoopJoin`, which joins the resulting records via a normal join, with no option of turning it into a left join. Is that correct?

In most situations there was a way to work around that (or the alternative way of constructing the query was more optimized anyway), but I seem to be encountering situation where a `LeftSequence` would be the only way to get the logic to work correctly AFAIK.

Tpt:
> One thing that I've been stubbing my toe on now is the join behaviour of GraphPattern::Sequence. As far as I can tell it will internally generate a ForLoopJoin, which joins the resulting records via a normal join, with no option of turning it into a left join. Is that correct?

Yes.

> In most situations there was a way to work around that (or the alternative way of constructing the query was more optimized anyway), but I seem to be encountering situation where a LeftSequence would be the only way to get the logic to work correctly AFAIK.

That's a great point. This make me realise I have made a huge mistake in the current implementation of `OX_LATERAL` with optionals: if we write
```sparql
SELECT * WHERE {
  VALUES ?s { ex:s }
  OPTIONAL {
    FILTER(!BOUND(?s))
    VALUES ?o { ex:o }
  }
}
```
it returns properly (?s = `ex:s`, ?o = `ex:o`) because in the left joins the two sides should be evaluated independently.
But in 
```sparql
SELECT * WHERE {
  VALUES ?s { ex:s }
  OPTIONAL {
    OX_LATERAL(?s) {SELECT * WHERE {
    FILTER(!BOUND(?s))
    VALUES ?o { ex:o }
  }}
}
```
it returns (?s = `ex:s`, ?o = `UND`) because `?s` is wrongly captured inside the optional, breaking the independence.

To fix that properly I believe we should introduce as you said a `LeftSequence` that would properly support the semantic of running the optional for each tuple outside of it and surface it into the SPARQL syntax with something like `OX_LATERAL_OPTIONAL { O }` or `OX_LATERAL OPTIONAL { O }`.

This way:
```sparql
SELECT * WHERE {
  VALUES ?s { ex:s }
  OX_LATERAL OPTIONAL {
    FILTER(!BOUND(?s))
    VALUES ?o { ex:o }
  }
}
```
would return (?s = `ex:s`, ?o = `UND`) as expected and would allow the usage of LATERAL subqueries inside it.

Side note: it is likely we would also need "LATERAL" support for the "GRAPH" and "SERVICE" keywords.

What do you think about it?

Tpt:
@hobofan I have updated this pull request. The `OX_LATERAL` subqueries inside `OPTIONAL` are now properly independant from the outside code. I have introduced `OX_LATERAL OPTIONAL`  and a `LeftSequence` operator that should provide the behavior you are looking for.

Side note: The python failure in the CI is independant of this PR.

Tpt:
@hobofan Thank you! Great to hear it works for you and you avoided the algebra hacks. Indeed having two nested "LATERAL" is quite verbose but not having them would mean that everything inside "LATERAL" is going to be "LATERAL" so we would have to build a way to "escape" LATERAL to express some queries...

> What I can do is giving feedback from the "consumer" perspective, and from that point of view, the feature feels great! :)

Thank you! "Customer" feedbacks are amazing to check that the features are actually usable and understandable.

Tpt:
I have opened #336 with an implementation following [SPARQL 1.2 SEP-0006](https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0006/sep-0006.md) and compatible with Apache Jena.

Tpt:
The other implementation is now merged and released

Issue #269, opened 2022-10-26
Server: Allows SPARQL querying from the command line


Tpt:
It might be nice to be able to execute SPARQL queries from the cli without having to spawn a server.
For example (not sure it's the best interface):
`cat my-query.rq | oxigraph_server --location XXXXX query --output-format=csv > result.csv`

RichDijk:
Would be convenient indeed.
In this use case the server executes the sparql. To me the server looks heavy for this purpose because only a store.query with output formatting is required.

joepie91:
As there is currently an `oxigraph_server query` command, is this issue resolved, or is there still some detail left to implement?

Tpt:
This issue should definitely be closed. Thank you for spotting it!

PR #326, opened 2022-12-21
Server: Open readonly and secondary

BenediktSeidl:
i saw #88 and thought i give it a try. this is of course not finished, it's more a sketch how i would implement this feature.  it should be enough to start a discussion, and exchange ideas, i've written some code comments when i was unsure how to do things.
i'm open for suggestions, still new to rust.

if you open the query page, and press the play button you should see a password prompt. the credentials are currently hard coded.

### DIFF
--- a/lib/src/storage/backend/rocksdb.rs
+++ b/lib/src/storage/backend/rocksdb.rs
@@ -3,7 +3,7 @@
 #![allow(unsafe_code, trivial_casts)]
 
 use crate::storage::error::StorageError;
-use crate::store::CorruptionError;
+use crate::store::{CorruptionError, StoreOpenOptions};
 use lazy_static::lazy_static;
 use libc::{self, c_char, c_void, free};
 use oxrocksdb_sys::*;
@@ -24,6 +24,13 @@ use std::rc::{Rc, Weak};
 use std::sync::Arc;
 use std::thread::{available_parallelism, yield_now};
 use std::{ptr, slice};
+
+#[derive(PartialEq, Eq)]
+enum OpeningMode {
+    InMemory,
+    Primary,
+    Secondary(PathBuf),
+}
 
 macro_rules! ffi_result {
     ( $($function:ident)::*() ) => {
@@ -69,6 +76,28 @@ lazy_static! {
     };
 }
 
+fn primary_db_handler_or_error(
+    db: &InnerDbHandler,
+) -> Result<&InnerDbHandlerPrimary, StorageError> {
+    match db {
+        InnerDbHandler::Primary(inner_db) | InnerDbHandler::InMemory(inner_db) => Ok(inner_db),
+        _ => Err(StorageError::Other(
+            "Database is opened as secondary, can not execute this operation.".into(),
+        )),
+    }
+}
+
+fn secondary_db_handler_or_error(
+    db: &InnerDbHandler,
+) -> Result<&InnerDbHandlerSecondary, StorageError> {
+    match db {
+        InnerDbHandler::Secondary(inner_db) => Ok(inner_db),
+        _ => Err(StorageError::Other(Box::<dyn Error + Send + Sync>::from(
+            "Expected secondary InnerDbHandler, got primary/in memory one",
+        ))),
+    }
+}
+
 pub struct ColumnFamilyDefinition {
     pub name: &'static str,
     pub use_iter: bool,
@@ -84,11 +113,44 @@ unsafe impl Send for Db {}
 
 unsafe impl Sync for Db {}
 
-struct DbHandler {
+/// Handler that has read only access to the database
+struct InnerDbHandlerSecondary {
+    db: *mut rocksdb_t,
+    primary_path: PathBuf,
+}
+
+/// Handler that has read and write access to the database
+struct InnerDbHandlerPrimary {
     db: *mut rocksdb_transactiondb_t,
-    options: *mut rocksdb_options_t,
     transaction_options: *mut rocksdb_transaction_options_t,
     transactiondb_options: *mut rocksdb_transactiondb_options_t,
+    path: PathBuf,
+}
+
+enum InnerDbHandler {
+    Primary(InnerDbHandlerPrimary),
+    InMemory(InnerDbHandlerPrimary),
+    Secondary(InnerDbHandlerSecondary),
+}
+
+impl InnerDbHandler {
+    fn is_null(&self) -> bool {
+        match self {
+            Self::Primary(s) | Self::InMemory(s) => s.db.is_null(),
+            Self::Secondary(s) => s.db.is_null(),
+        }
+    }
+    fn primary_path(&self) -> &PathBuf {
+        match self {
+            Self::Primary(s) | Self::InMemory(s) => &s.path,
+            Self::Secondary(s) => &s.primary_path,
+        }
+    }
+}
+
+struct DbHandler {
+    db: InnerDbHandler,
+    options: *mut rocksdb_options_t,
     read_options: *mut rocksdb_readoptions_t,
     write_options: *mut rocksdb_writeoptions_t,
     flush_options: *mut rocksdb_flushoptions_t,
@@ -99,8 +161,6 @@ struct DbHandler {
     column_family_names: Vec<&'static str>,
     cf_handles: Vec<*mut rocksdb_column_family_handle_t>,
     cf_options: Vec<*mut rocksdb_options_t>,
-    path: PathBuf,
-    in_memory: bool,
 }
 
 impl Drop for DbHandler {
@@ -109,7 +169,14 @@ impl Drop for DbHandler {
             for cf_handle in &self.cf_handles {
                 rocksdb_column_family_handle_destroy(*cf_handle);
             }
-            rocksdb_transactiondb_close(self.db);
+            match &self.db {
+                InnerDbHandler::Primary(s) | InnerDbHandler::InMemory(s) => {
+                    rocksdb_transactiondb_close(s.db);
+                }
+                InnerDbHandler::Secondary(s) => {
+                    rocksdb_close(s.db);
+                }
+            }
             for cf_option in &self.cf_options {
                 rocksdb_options_destroy(*cf_option);
             }
@@ -119,13 +186,19 @@ impl Drop for DbHandler {
             rocksdb_envoptions_destroy(self.env_options);
             rocksdb_ingestexternalfileoptions_destroy(self.ingest_external_file_options);
             rocksdb_compactoptions_destroy(self.compaction_options);
-            rocksdb_transaction_options_destroy(self.transaction_options);
-            rocksdb_transactiondb_options_destroy(self.transactiondb_options);
+            if let InnerDbHandler::Primary(inner_db) | InnerDbHandler::InMemory(inner_db) = &self.db
+            {
+                rocksdb_transaction_options_destroy(inner_db.transaction_options);
+                rocksdb_transactiondb_options_destroy(inner_db.transactiondb_options);
+            }
             rocksdb_options_destroy(self.options);
             rocksdb_block_based_options_destroy(self.block_based_table_options);
         }
-        if self.in_memory && self.path.exists() {
-            remove_dir_all(&self.path).unwrap();
+
+        if let InnerDbHandler::InMemory(db) = &self.db {
+            if db.path.exists() {
+                remove_dir_all(&db.path).unwrap();
+            }
         }
     }
 }
@@ -138,7 +211,11 @@ impl Db {
             temp_dir()
         }
         .join(format!("oxigraph-rocksdb-{}", random::<u128>()));
-        Ok(Self(Arc::new(Self::do_open(path, column_families, true)?)))
+        Ok(Self(Arc::new(Self::do_open(
+            path,
+            column_families,
+            &OpeningMode::InMemory,
+        )?)))
     }
 
     pub fn open(
@@ -148,17 +225,29 @@ impl Db {
         Ok(Self(Arc::new(Self::do_open(
             path.to_owned(),
             column_families,
-            false,
+            &OpeningMode::Primary,
         )?)))
+    }
+
+    pub fn open_with_options(
+        options: StoreOpenOptions,
+        column_families: Vec<ColumnFamilyDefinition>,
+    ) -> Result<Self, StorageError> {
+        match options {
+            StoreOpenOptions::OpenAsSecondary(options) => Ok(Self(Arc::new(Self::do_open(
+                options.path,
+                column_families,
+                &OpeningMode::Secondary(options.secondary_path),
+            )?))),
+        }
     }
 
     fn do_open(
         path: PathBuf,
         mut column_families: Vec<ColumnFamilyDefinition>,
-        in_memory: bool,
+        open_mode: &OpeningMode,
     ) -> Result<DbHandler, StorageError> {
         let c_path = path_to_cstring(&path)?;
-
         unsafe {
             let options = rocksdb_options_create();
             assert!(!options.is_null(), "rocksdb_options_create returned null");
@@ -170,28 +259,32 @@ impl Db {
                 available_parallelism()?.get().try_into().unwrap(),
             );
             if let Some(available_fd) = available_file_descriptors()? {
-                if available_fd < 96 {
-                    rocksdb_options_destroy(options);
-                    return Err(io::Error::new(
-                        io::ErrorKind::Other,
-                        format!(
-                            "Oxigraph needs at least 96 file descriptors, only {available_fd} allowed. Run e.g. `ulimit -n 512` to allow 512 opened files"
-
-                        ),
-                    )
-                    .into());
-                }
-                rocksdb_options_set_max_open_files(
-                    options,
-                    (available_fd - 48).try_into().unwrap(),
-                );
+                let max_open_files = match &open_mode {
+                    OpeningMode::Primary | OpeningMode::InMemory => {
+                        if available_fd < 96 {
+                            rocksdb_options_destroy(options);
+                            return Err(io::Error::new(
+                                io::ErrorKind::Other,
+                                format!(
+                                    "Oxigraph needs at least 96 file descriptors, \
+                                    only {available_fd} allowed. \
+                                    Run e.g. `ulimit -n 512` to allow 512 opened files"
+                                ),
+                            )
+                            .into());
+                        }
+                        (available_fd - 48).try_into().unwrap()
+                    }
+                    OpeningMode::Secondary(_) => -1,
+                };
+                rocksdb_options_set_max_open_files(options, max_open_files);
             }
             rocksdb_options_set_info_log_level(options, 2); // We only log warnings
             rocksdb_options_set_max_log_file_size(options, 1024 * 1024); // Only 1MB log size
             rocksdb_options_set_recycle_log_file_num(options, 10); // We do not keep more than 10 log files
             rocksdb_options_set_compression(
                 options,
-                if in_memory {
+                if open_mode == &OpeningMode::InMemory {
                     rocksdb_no_compression
                 } else {
                     rocksdb_lz4_compression
@@ -201,7 +294,7 @@ impl Db {
             );
             rocksdb_options_set_env(
                 options,
-                if in_memory {
+                if open_mode == &OpeningMode::InMemory {
                     ROCKSDB_MEM_ENV.0
                 } else {
                     ROCKSDB_ENV.0
@@ -218,12 +311,6 @@ impl Db {
                 16,
             );
             rocksdb_options_set_block_based_table_factory(options, block_based_table_options);
-
-            let transactiondb_options = rocksdb_transactiondb_options_create();
-            assert!(
-                !transactiondb_options.is_null(),
-                "rocksdb_transactiondb_options_create returned null"
-            );
 
             if !column_families.iter().any(|c| c.name == "default") {
                 column_families.push(ColumnFamilyDefinition {
@@ -261,24 +348,76 @@ impl Db {
 
             let mut cf_handles: Vec<*mut rocksdb_column_family_handle_t> =
                 vec![ptr::null_mut(); column_family_names.len()];
-            let db = ffi_result!(rocksdb_transactiondb_open_column_families_with_status(
-                options,
-                transactiondb_options,
-                c_path.as_ptr(),
-                c_column_families.len().try_into().unwrap(),
-                c_column_families
-                    .iter()
-                    .map(|cf| cf.as_ptr())
-                    .collect::<Vec<_>>()
-                    .as_ptr(),
-                cf_options.as_ptr() as *const *const rocksdb_options_t,
-                cf_handles.as_mut_ptr(),
-            ))
+
+            let c_num_column_families = c_column_families.len().try_into().unwrap();
+            let c_column_family_names = c_column_families
+                .iter()
+                .map(|cf| cf.as_ptr())
+                .collect::<Vec<_>>();
+
+            let db = match &open_mode {
+                OpeningMode::Primary | OpeningMode::InMemory => {
+                    let transactiondb_options = rocksdb_transactiondb_options_create();
+                    assert!(
+                        !transactiondb_options.is_null(),
+                        "rocksdb_transactiondb_options_create returned null"
+                    );
+
+                    ffi_result!(rocksdb_transactiondb_open_column_families_with_status(
+                        options,
+                        transactiondb_options,
+                        c_path.as_ptr(),
+                        c_num_column_families,
+                        c_column_family_names.as_ptr(),
+                        cf_options.as_ptr() as *const *const rocksdb_options_t,
+                        cf_handles.as_mut_ptr(),
+                    ))
+                    .map(|db| {
+                        let transaction_options = rocksdb_transaction_options_create();
+                        assert!(
+                            !transaction_options.is_null(),
+                            "rocksdb_transaction_options_create returned null"
+                        );
+                        rocksdb_transaction_options_set_set_snapshot(transaction_options, 1);
+                        let primary_db_handler = InnerDbHandlerPrimary {
+                            db,
+                            transaction_options,
+                            transactiondb_options,
+                            path,
+                        };
+                        match open_mode {
+                            OpeningMode::InMemory => InnerDbHandler::InMemory(primary_db_handler),
+                            OpeningMode::Primary => InnerDbHandler::Primary(primary_db_handler),
+                            _ => unreachable!(),
+                        }
+                    })
+                    .map_err(|e| {
+                        rocksdb_transactiondb_options_destroy(transactiondb_options);
+                        e
+                    })
+                }
+                OpeningMode::Secondary(secondary_path) => {
+                    ffi_result!(rocksdb_open_as_secondary_column_families_with_status(
+                        options,
+                        c_path.as_ptr(),
+                        path_to_cstring(secondary_path)?.as_ptr(),
+                        c_num_column_families,
+                        c_column_family_names.as_ptr(),
+                        cf_options.as_ptr() as *const *const rocksdb_options_t,
+                        cf_handles.as_mut_ptr(),
+                    ))
+                    .map(|db| {
+                        InnerDbHandler::Secondary(InnerDbHandlerSecondary {
+                            db,
+                            primary_path: path,
+                        })
+                    })
+                }
+            }
             .map_err(|e| {
                 for cf_option in &cf_options {
                     rocksdb_options_destroy(*cf_option);
                 }
-                rocksdb_transactiondb_options_destroy(transactiondb_options);
                 rocksdb_options_destroy(options);
                 rocksdb_block_based_options_destroy(block_based_table_options);
                 e
@@ -302,7 +441,7 @@ impl Db {
                 !write_options.is_null(),
                 "rocksdb_writeoptions_create returned null"
             );
-            if in_memory {
+            if open_mode == &OpeningMode::InMemory {
                 rocksdb_writeoptions_disable_WAL(write_options, 1); // No need for WAL
             }
 
@@ -330,18 +469,9 @@ impl Db {
                 "rocksdb_compactoptions_create returned null"
             );
 
-            let transaction_options = rocksdb_transaction_options_create();
-            assert!(
-                !transaction_options.is_null(),
-                "rocksdb_transaction_options_create returned null"
-            );
-            rocksdb_transaction_options_set_set_snapshot(transaction_options, 1);
-
             Ok(DbHandler {
                 db,
                 options,
-                transaction_options,
-                transactiondb_options,
                 read_options,
                 write_options,
                 flush_options,
@@ -352,8 +482,6 @@ impl Db {
                 column_family_names,
                 cf_handles,
                 cf_options,
-                path,
-                in_memory,
             })
         }
     }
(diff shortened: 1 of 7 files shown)


server/src/auth.rs
```diff
@@ -0,0 +1,144 @@
...
+            User::Anonymous => false,
+            User::Named(_) => true,
+        },
+        _ => true,
+    }
+}
+
+// a similar function could be moved to oxhttp?
+fn read_user_from_request(request: &Request) -> anyhow::Result<User> {
+    match request.header(&HeaderName::AUTHORIZATION) {
```
Tpt, line 58: Rust tip (no need to change the code):
```rust
match option {
    Some(v) => SOME,
    None => NONE
}
```

can be also written
```rust
if let Some(v) = option {
    SOME
} else {
    NONE
}
```

server/src/auth.rs
```diff
@@ -0,0 +1,144 @@
...
+        None => Ok(User::Anonymous),
+        Some(header) => match header.to_str().unwrap().split_once(" ") {
+            None => Err(anyhow::anyhow!("Could not read Header")),
+            Some((prefix, payload)) => {
+                if prefix != "Basic" {
+                    return Err(anyhow::anyhow!("Only support Basic Auth"));
+                }
+                match decode(payload) {
+                    Err(_) => Err(anyhow::anyhow!("Could not base64 decode header")),
```
Tpt, line 67: anyhow has the great [`context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context) method that allows to add context to errors.

For example you could write
```rust
let decoded = decode(payload).context("Could not base64 decode header")?
```
and get the same behavior + the original error is tracked and printed after the new message.

server/src/auth.rs
```diff
@@ -0,0 +1,144 @@
...
+
+// a similar function could be moved to oxhttp?
+fn read_user_from_request(request: &Request) -> anyhow::Result<User> {
+    match request.header(&HeaderName::AUTHORIZATION) {
+        None => Ok(User::Anonymous),
+        Some(header) => match header.to_str().unwrap().split_once(" ") {
+            None => Err(anyhow::anyhow!("Could not read Header")),
+            Some((prefix, payload)) => {
+                if prefix != "Basic" {
+                    return Err(anyhow::anyhow!("Only support Basic Auth"));
```
Tpt, line 64: you can just use `anyhow::bail!(` instead of `return Err(anyhow::anyhow!(`.

server/src/auth.rs
```diff
@@ -0,0 +1,144 @@
...
+            // only valid users and anonymous users left
+            if !can_user_access(&user, required_permissions) {
+                return match user {
+                    User::Anonymous => Err((Status::UNAUTHORIZED, "unauthorized".into())),
+                    User::Named(_) => Err((Status::FORBIDDEN, "forbidden".into())),
+                };
+            }
+            return Ok(());
+        }
+        Err(e) => Err((Status::BAD_REQUEST, format!("Bad request: {:?}", e))),
```
Tpt, line 101: Instead of a `match` you can use 
```rust
let user = read_user_from_request(request).map_err(|e|  (Status::BAD_REQUEST, format!("Bad request: {:?}", e)))?
```

server/src/auth.rs
```diff
@@ -0,0 +1,144 @@
+use base64::decode;
+use oxhttp::model::{HeaderName, Request, Status};
+
+pub enum Store {
+    All,
+    // this would allow specifying different permissions for each store,
+    // not 100% sure if i understood this correctly,
+    // or if it's easier to leave this out.
+    Defined(String),
+}
```
Tpt, line 10: Currently, `sparql_server` exposes a single store

Tpt (commented):
Thank you very much for this PR! It's nice to see interest in Oxigraph.

The code looks great. I allowed myself some suggestions around syntactic sugars but nothing important.

On the feature itself, to say the truth, I am a big reluctant to add access control to Oxigraph, it's a rabbit hole and properly implementing it is not easy. But if you are willing to do the work, I would be more than happy to review and merge it.

As you wrote in the comment, the major question is how to represents users/tokens/rights, how to store them and which expressivity do we want.
I believe for now we want some "simple rights" (maybe only "read" or "write", allowing SPARQL UPDATE means allowing nearly any change, even load from the web).
To store the (users, token, rights) tuples maybe a plain text file could do the job, having a database for that seems a bit over engineered but I would be happy to be convinced otherwise. What do you think about it?

Thank you again!

nyurik:
A fairly simple approach I saw with Blazegraph is to separate access with nginx or a similar proxy - i.e. a `GET` request is public, but the `POST / PUT / DELETE / ...` requests must only come from the internal network.  This of course allow bad configurations to give access to the data, but that's no different than misconfiguring the database itself.

A simple solution would be to add a recomended nginx and possibly other proxy configurations as examples to this repo, but I think access control is a bit premature for Oxigraph, and may distract from making it the best possible SPARQL engine on the planet :)

Tpt:
@nyurik Yes for nginx. There is already [a section about it in the server readme](https://github.com/oxigraph/oxigraph/tree/main/server#run-the-web-server-with-basic-authentication) (feel free to improve it if you feel so). It's what I am doing for my own Oxigraph deployments. About the priority of access control, I agree it's not a priority. It's why I don't plan to spend my own time on it. But it's a regularly asked question and if someone wants to take the time to implement it properly, I am not against it.

BenediktSeidl:
> I am a big reluctant to add access control to Oxigraph

This was probably a Freudian slip? ;-) I don't want to force this on you. In worst case you will have to support this feature, so i won't be angry if you don't want to include this. I like to browse through old open tickets in Github projects I find interesting, and I thought that this could be a problem I could tackle.

Regarding the auth backend: I was probably taken away by "just imagine the possibilities" and thought it would be fun to use the database as backend, but i'm totally fine with a simpler solution.

I would suggest to use a [htpasswd](https://httpd.apache.org/docs/current/programs/htpasswd.html) file, together with a [auth group file](https://httpd.apache.org/docs/2.4/mod/mod_authz_groupfile.html#authgroupfile)? Perhaps restricting the htpasswd file to only allow bcrypt passwords and refusing to start when other (insecure) hashes are used?
* ✅ use existing technology and known file formats
* ❌ need to point to two different files (or explain that a `<filename>.htpasswd` and `<filename>.authgroup` will be used)
otherwise we could also come up with some schema for a toml file:
* ❌ there is no easy way to add password hashes to the toml file, so one would have to explain how to hash the password

I'm also fine to reduce the granularity of the groups to just "read" and "write".

Have you thought about how to use the error mechanism to transport the additional header?

nyurik:
> > I am a big reluctant to add access control to Oxigraph
> 
> This was probably a Freudian slip? ;-) I don't want to force this on you. In worst case you will have to support this feature, so i won't be angry if you don't want to include this. I like to browse through old open tickets in Github projects I find interesting, and I thought that this could be a problem I could tackle.
> 
> Regarding the auth backend: I was probably taken away by "just imagine the possibilities" and thought it would be fun to use the database as backend, but i'm totally fine with a simpler solution.
> 
> I would suggest to use a [htpasswd](https://httpd.apache.org/docs/current/programs/htpasswd.html) file, together with a [auth group file](https://httpd.apache.org/docs/2.4/mod/mod_authz_groupfile.html#authgroupfile)? Perhaps restricting the htpasswd file to only allow bcrypt passwords and refusing to start when other (insecure) hashes are used?
> 
> * white_check_mark use existing technology and known file formats
> * x need to point to two different files (or explain that a `<filename>.htpasswd` and `<filename>.authgroup` will be used)
>   otherwise we could also come up with some schema for a toml file:
> * x there is no easy way to add password hashes to the toml file, so one would have to explain how to hash the password
> 
> I'm also fine to reduce the granularity of the groups to just "read" and "write".
> 
> Have you thought about how to use the error mechanism to transport the additional header?

hehe, must have been :)

The biggest issue with auth stuff is multi-platform support - the moment you introduce password backends, you will have to handle them all, which is obviously a massive pain. So either store it in the DB itself (i.e. introduce a separate password-only store inaccessible to the users, but available to the server), or have backend support for all the different OSes - meh in all sorts of ways

Tpt:
> This was probably a Freudian slip? ;-) I don't want to force this on you. In worst case you will have to support this feature, so i won't be angry if you don't want to include this.

Maybe XD. I am fine having something in Oxigraph Server, as soon as it stays easy to maintain (no big ACL system). But having users, tokens and read/write (or maybe a bit more fine-grain) authorizations is fine for me.

> I like to browse through old open tickets in Github projects I find interesting, and I thought that this could be a problem I could tackle.

Thank you, it's very kind :).


> I would suggest to use a [htpasswd](https://httpd.apache.org/docs/current/programs/htpasswd.html) file, together with a [auth group file](https://httpd.apache.org/docs/2.4/mod/mod_authz_groupfile.html#authgroupfile)? Perhaps restricting the htpasswd file to only allow bcrypt passwords and refusing to start when other (insecure) hashes are used?

That sounds like a good idea!

A bold proposal (not sure if it's a good idea): Instead of having a full auth systems with users, passwords and rights what about just allowing to restrict write access to nobody/from a given set of IPs/from anywhere? This would cover the following use cases:
- I want a read only server. In this case, it is enough to just close write access
- I want a read only server from anywhere but my localhost and I serve Oxigraph publicly without reverse proxy (bad idea but why not...): In this case, it is enough to just close write access to anything but localhost.
- I am using a reverse proxy: I manage routes there using my reverse proxy features so I don't need access control anyway (for read-only whitelist /query and /store with OPTIONS and GET methods).
What do you think about it?

> So either store it in the DB itself (i.e. introduce a separate password-only store inaccessible to the users, but available to the server), or have backend support for all the different OSes - meh in all sorts of ways

That's a good point, I believe we definitely don't want to open the pandora box of integrating with auth systems, LDAPs... Deployment requiring that will probably want things like backups, https... Hence, will use already quite a lot of tooling around Oxigraph.

> Have you thought about how to use the error mechanism to transport the additional header?

Because it is based on the HTTP status code I feel your implementation is fine, having to build a `Response` for each error is a bit cumbersome imho because we can't just write `impl From<(StatusCode, String) for Response` because of the impl restrictions on foreign types.

BenediktSeidl:
another bold idea: is it possible to use one rocksdb file from a different process in read only mode? if yes the following could work:
* first instance of oxigraph server has read/write access to rocksdb file, tcp access is restricted by a bind argument, or blocked by a firewall. could even be shut down when data will not be changed for longer time.
* second read only instance of oxipraph server binds to 0.0.0.0, but has read only access.

as the readonly access is on rocksdb level, it should be impossible to circumvent it?

yesterday night i found another problem with the basic auth idea: basic auth should only be used with https (password is only base64 encoded). so in order to safely use basic auth, you would need to add a reverse proxy providing https in front of oxigraph server. if you already have this, then adding auth in the reverse proxy would be easy. so it does not make sense to implement it directly in oxigraph server?!

Tpt:
> another bold idea: is it possible to use one rocksdb file from a different process in read only mode? if yes the following could work:

Sadly not. RocksDB has a global lock and won't open if an other process already aquired it.

> yesterday night i found another problem with the basic auth idea: basic auth should only be used with https (password is only base64 encoded). so in order to safely use basic auth, you would need to add a reverse proxy providing https in front of oxigraph server. if you already have this, then adding auth in the reverse proxy would be easy. so it does not make sense to implement it directly in oxigraph server?!

That's a good point. Indeed, secure basic auth is not possible without https. So, it leaves us with adding also https or staying with the "without-auth" approaches (read-only endpoint, IP-whitelist...).

Side note: if you want a simpler task without big design problems, there is e.g. #269

BenediktSeidl:
there seems to be `OpenForReadOnly` and `OpenAsSecondary` [[1]](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ#basic-readwrite) but i've never worked with RocksDB so i don't really know if it works out.

okay. let's see what we've got: basic auth is definitely out, because adding https is out of scope?
still left: ip whitelisting and read-only endpoint.
what do you mean by "read-only endpoint"?

does it help to go back to square one and think about the usecases of auth in general and built in auth especially and come up with solutions to that problem?
my current train of thoughts is:
* auth is only relevant for public instances of oxigraph
* if the instance of oxigraph is public you should have some reverseproxy in front of oxigraph
* if you have a reverseproxy in front of oxigraph, you can add auth there

if you have a small test-instance you may be okay without auth.

i would love to see something coming out of this discussion, but on the same time i don't want to waste your time :-) feel free to close if we are done here, i will then have a look at #269 :-)

Tpt:
> there seems to be OpenForReadOnly and OpenAsSecondary [[1]](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ#basic-readwrite) but i've never worked with RocksDB so i don't really know if it works out.

Thank you so much! I completely forgot about these functions. If I understand correct, `OpenForReadOnly` allows multiple reading threads (but no read/write one created using `Open`) while `OpenAsSecondary` allows multiple reading threads when a read/write one exists.

>  what do you mean by "read-only endpoint"?

A server where all edit routes ("/update" and "/store" with POST, PUT and DELETE) are disabled. Sorry, I used "endpoint" to mean "server".

>  my current train of thoughts is:
>    auth is only relevant for public instances of oxigraph
 >   if the instance of oxigraph is public you should have some reverseproxy in front of oxigraph
  >  if you have a reverseproxy in front of oxigraph, you can add auth there

Yes! I 100% agree.

> i would love to see something coming out of this discussion.

Yes, me too! I believe we are starting to get some good ideas :). Thank you for that!

Proposal (not sure if it's a great idea). Add a `--read-only` option that opens the RocksDB database in read-only mode using `OpenReadOnly` and maybe a  `--secondary` option that opens the RocksDB database as "secondary" using `OpenAsSecondary`. In these cases the update operations would fail from the errors returned by RocksDB (so no need to tweak to much the server code). What do you think about it?

BenediktSeidl:
is it possible that `OpenForReadOnly` and `OpenAsSecondary` are not implemented for `TransactionDB`?
oxigraph is using `rocksdb_transactiondb_open_column_families_with_status` to open the database, but i could not find any way to adapt this to read-only.

Tpt:
> is it possible that OpenForReadOnly and OpenAsSecondary are not implemented for TransactionDB?
oxigraph is using rocksdb_transactiondb_open_column_families_with_status to open the database, but i could not find any way to adapt this to read-only.

Thank you for investigating! It's a great question. `OpenForReadOnly` and `OpenAsSecondary` both return a read-only database so the transactional API is not useful in these cases (snapshots are still available on `DB` objects). It's maybe why these functions are not available for `TransactionDB`. Imho, it's worth testing if having a `TransactionalDB` as the writer process does not create conflict with a `DB` as a secondary process. But making it work with the current codebase might require a bit of work so it's maybe not a Rust/Oxigraph beginner-friendly task. Feel free to pick something else if you feel it's not a good task for you.

server/src/auth.rs
MarcAntoine-Arnaud: or also:
```rust
option.then_some(something)
```

nikolaswise:
For what it's worth, even a simple "all or nothing" basic auth would be very useful for my use cases. I want to have very transparent deployments that can be accessed from many services (via something like fly.io). Just shutting the whole server behind basic auth would be great, and I can figure out the more intensive access control and agent modeling in my applications.

nikolaswise:
@BenediktSeidl totally - if I deploy the Oxigraph Server Docker image to a platform like fly.io, I get all the SSL layers and access points as their service. But my sparql endpoints remain unsecured and available to the public internet.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -647,8 +753,25 @@ impl Drop for Reader {
...
+    fn contains_key(&self, column_family: &ColumnFamily, key: &[u8]) -> Result<bool, StorageError>;
+    fn iter(&self, column_family: &ColumnFamily) -> Result<Iter, StorageError>;
+    fn scan_prefix(
+        &self,
+        column_family: &ColumnFamily,
+        prefix: &[u8],
+    ) -> Result<Iter, StorageError>;
+    fn len(&self, column_family: &ColumnFamily) -> Result<usize, StorageError>;
+    fn is_empty(&self, column_family: &ColumnFamily) -> Result<bool, StorageError>;
+}
```
Tpt, line 771: Instead of a trait we might embedd an enum instead of the `Reader` struct just like what we do for `DbHandler`. It might allow to avoid some code duplication.

Tpt (commented):
@BenediktSeidl Thank you for moving forward with this! It looks nice.

I have not reviewed the PR carefully yet. Some preliminary design question:

1. About naming. Do we want to talk about "read-only" vs "read-write" database modes or "primary" vs "secondary". We should settle on something. I would tend to prefer "primary" vs "secondary" because we have the "catch-up" feature enabled. And it would not prevent us to provide a static "read-only" mode afterward. What do you think about it?

2. RocksDB secondary mode does not supports snapshots. It means that it's likely that if a transaction commits during a SPARQL query evaluation, the query ends up being parially evaluate on the before state and partially after, opposite to the "primary" mode. It's quite sad and should definitely be stressed out in the documentation.

3. I am not sure that your test evaluates properly that there is proper isolation: if we write a single key in the transaction, it is more than likely it's kept in the in-memory overload and written to disk only after "commit", and so, there is no reason for the other instances to read it before commit. This does not prevent the "big transaction" cases where RocksDB writes to disk some data before the transaction commit. But I am not very familiar with RocksDB so I might be wrong.

Thank you again!

oxrocksdb-sys/api/c.cc
```diff
@@ -19,6 +19,38 @@ static bool SaveStatus(rocksdb_status_t* target, const Status source) {
...
+rocksdb_t* rocksdb_open_as_secondary_column_families_with_status(
+        const rocksdb_options_t* db_options, const char* name,
+        const char* secondary_path, int num_column_families,
+        const char* const* column_family_names,
+        const rocksdb_options_t* const* column_family_options,
+        rocksdb_column_family_handle_t** column_family_handles,
+        rocksdb_status_t* statusptr) {
+    std::vector<ColumnFamilyDescriptor> column_families;
+    for (int i = 0; i != num_column_families; ++i) {
+        column_families.emplace_back(
```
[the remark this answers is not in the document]
BenediktSeidl: oh. i assumed all code in this file was a copied from `oxrocksdb-sys/rocksdb/db/c.cc` with `SaveError` being replaced with `SaveStatus`? That's at least what I did.

lib/src/store.rs
```diff
@@ -89,16 +106,17 @@ pub struct Store {
...
+        // TODO: maybe remove?
         Ok(Self {
             storage: Storage::new()?,
         })
     }
 
     /// Opens a [`Store`] and creates it if it does not exist yet.
     #[cfg(not(target_family = "wasm"))]
-    pub fn open(path: impl AsRef<Path>) -> Result<Self, StorageError> {
+    pub fn open(options: StoreOpenOptions) -> Result<Self, StorageError> {
```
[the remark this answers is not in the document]
BenediktSeidl: I reverted the breaking change, and introduced the new method `open_with_options`.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -138,27 +183,28 @@ impl Db {
...
-        )?)))
+        Ok(Self(Arc::new(Self::do_open(options, column_families)?)))
     }
 
     fn do_open(
-        path: PathBuf,
+        store_open_options: StoreOpenOptions,
         mut column_families: Vec<ColumnFamilyDefinition>,
-        in_memory: bool,
     ) -> Result<DbHandler, StorageError> {
```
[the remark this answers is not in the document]
BenediktSeidl: I introduced an enum `OpeningMode`.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -25,6 +25,26 @@ use std::sync::Arc;
...
 
+// TODO: how to get rid of this?
+macro_rules! expect_readonly_db {
+    ($e:expr) => {
+        match $e {
+            DbWrapper::Readonly(value) => value,
+            _ => panic!("dafuck!"),
+        }
+    };
+}
```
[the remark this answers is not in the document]
BenediktSeidl: thanks! now it's a simple function.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -171,6 +217,7 @@ impl Db {
             );
             if let Some(available_fd) = available_file_descriptors()? {
                 if available_fd < 96 {
+                    // TODO: this is probably missleading for readonly? as secondary need -1
```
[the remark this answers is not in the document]
BenediktSeidl: i restructured the code so the check is only executed on primary databases.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -585,7 +684,9 @@ impl Db {
     }
 
     pub fn backup(&self, target_directory: &Path) -> Result<(), StorageError> {
-        if self.0.in_memory {
+        // TODO: it's problematic that both the StoreOpenOptions and the db hold information what
+        // kind of database we have, otherwise this could be simplified with a match statement.
```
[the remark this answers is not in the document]
BenediktSeidl: The enum `InnerDbHandler` holds now the information in which mode the database was opened and the field specific to this opening mode.

BenediktSeidl:
Thanks for the review! 

> I would tend to prefer "primary" vs "secondary" because we have the "catch-up" feature enabled. And it would not prevent us to provide a static "read-only" mode afterward. What do you think about it?

I renamed it and tried to stick to primary vs. secondary.

> RocksDB secondary mode does not supports snapshots. It means that it's likely that if a transaction commits during a SPARQL query evaluation, the query ends up being partially evaluated on the before state and partially after, opposite to the "primary" mode. It's quite sad and should definitely be stressed out in the documentation.

Oh! To be honest, if I knew this beforehand I would not have started implemented it. For me this would be a deal-breaker. I thought working transactions would be enough for data consistency. Should we still move forward here?

lib/src/storage/backend/rocksdb.rs
```diff
@@ -464,13 +602,30 @@ impl Db {
...
-                self.0.db,
-                self.0.read_options,
-                column_family.0,
-                key.as_ptr() as *const c_char,
-                key.len()
-            ))?;
+            let slice = match &self.0.db {
+                InnerDbHandler::Secondary(inner_db_handler) => {
+                    let mut errptr: Vec<i8> = vec![];
+                    let result = rocksdb_get_pinned_cf(
```
Tpt, line 608: Unused intermediate variable.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -464,13 +602,30 @@ impl Db {
...
-            let slice = ffi_result!(rocksdb_transactiondb_get_pinned_cf_with_status(
-                self.0.db,
-                self.0.read_options,
-                column_family.0,
-                key.as_ptr() as *const c_char,
-                key.len()
-            ))?;
+            let slice = match &self.0.db {
+                InnerDbHandler::Secondary(inner_db_handler) => {
+                    let mut errptr: Vec<i8> = vec![];
```
Tpt, line 607: We never read it to check if an error happend or not.

oxrocksdb-sys/api/c.cc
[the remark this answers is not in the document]
Tpt: Yes, exactly. It's what I did.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -749,6 +928,11 @@ impl Reader {
                         ));
                     }
                 }
+                InnerReader::Secondary(inner) => {
+                    let db = secondary_db_handler_or_error(&inner.db)?.db;
+                    ffi_result!(rocksdb_try_catch_up_with_primary_with_status(db))?;
```
Tpt, line 933: Shouldn't we do it when we open a reader in the `snapshot` method?

Tpt (commented):
> I renamed it and tried to stick to primary vs. secondary.

Thank you for all your changes. This PR looks great!

> Oh! To be honest, if I knew this beforehand I would not have started implemented it. For me this would be a deal-breaker. I thought working transactions would be enough for data consistency. Should we still move forward here?

That's a great question. I think it might still be valuable for some users and most of the work done here might be easily reused if we want to add the "read-only" mode. So, this is as you prefer, if you want to move this PR forward feel free to do so. If not I would be happy to pick it from where it is if it's fine for you and finish it (maybe with also the read-only mode). I also allowed myself some nit pickings in the comments.

Also, it seems that rust auto-formatter is not happy, you can easily run it with `cargo fmt` (some IDEs also provide options to "format on save").

lib/src/storage/backend/rocksdb.rs
BenediktSeidl: missed that one. created another `with_status` function.

lib/src/storage/backend/rocksdb.rs
BenediktSeidl: this works, but `storage.snapshot()` could then return an error. i stopped at `quads_for_pattern` which seems to not to be able to return an error and is part of the public api? so i added an `unwrap()` for now and it seems to work.

BenediktSeidl:
I've added a second commit with the read-only mode and tried to follow the comments you left. There are still some TODOs in the code that may improve the current code. If you see room for other improvements, feel free to tell me so :-)

lib/src/storage/backend/rocksdb.rs
```diff
@@ -84,11 +113,45 @@ unsafe impl Send for Db {}
...
+
+enum InnerDbHandler {
+    Primary(InnerDbHandlerPrimary),
+    InMemory(InnerDbHandlerPrimary),
+    Secondary(InnerDbHandlerSecondary),
+    ReadOnly(InnerDbHandlerSecondary),
+}
+
+impl InnerDbHandler {
+    pub fn is_null(&self) -> bool {
```
Tpt, line 138: Nit: no reason for the method to be `pub` if the enum is not `pub` itself.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -261,19 +359,84 @@ impl Db {
...
+            let num_column_families = c_column_families.len().try_into().unwrap();
+            let column_family_names = c_column_families
+                .iter()
+                .map(|cf| cf.as_ptr())
+                .collect::<Vec<_>>()
+                .as_ptr();
+            let column_family_options = cf_options.as_ptr() as *const *const rocksdb_options_t;
+            let column_family_handles = cf_handles.as_mut_ptr();
+
+            let db = match &open_mode {
```
Tpt, line 373: Nit: make `OpeningMode` copy with `#[derive(Clone, Clopy)]` to avoid having to make reference. 

We can also derive `Eq` to be able to make equality comparisons instead of using `if let`.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -261,19 +359,84 @@ impl Db {
...
+                OpeningMode::Primary | OpeningMode::InMemory => {
+                    let transaction_options = rocksdb_transaction_options_create();
+                    assert!(
+                        !transaction_options.is_null(),
+                        "rocksdb_transaction_options_create returned null"
+                    );
+                    rocksdb_transaction_options_set_set_snapshot(transaction_options, 1);
```
Tpt, line 380: I would maybe move `transaction_options` creation before the database opening because if the database opening fails it might do a memory leak.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -261,19 +359,84 @@ impl Db {
...
+                        let primary_db_handler = InnerDbHandlerPrimary {
+                            db,
+                            transaction_options,
+                            transactiondb_options,
+                            path,
+                        };
+                        match open_mode {
+                            OpeningMode::InMemory => InnerDbHandler::InMemory(primary_db_handler),
+                            OpeningMode::Primary => InnerDbHandler::Primary(primary_db_handler),
+                            _ => panic!("this should not happen"),
```
Tpt, line 401: Nit: There is the `unreachable!` macro for this use case.

lib/src/storage/backend/rocksdb.rs
```diff
@@ -370,19 +522,39 @@ impl Db {
...
+                }
+                InnerDbHandler::ReadOnly(_) => Reader {
+                    inner: InnerReader::Secondary(self.0.clone()),
+                    options,
+                },
+                InnerDbHandler::Secondary(_) => {
+                    ffi_result!(rocksdb_try_catch_up_with_primary_with_status(
+                        secondary_db_handler_or_error(&self.0.db).unwrap().db,
+                    ))
+                    .unwrap();
```
Tpt, line 551: It might be nicer to return the error here instead of a panic.
Two options:
- change the method signature return a `Result` (may be very invasive in the upper API levels)
- add a `Option<StorageError>` to the returned object storing the error and then, the reader methods return the error (using e.g. the `Option::take` function) if it is set.

An other option might be to just ignore the error completely (or print it to a log) and read from the outdated version. Not sure what would be the best, it depends if we prefer availability or consistency...).

Tpt (commented):
Thank you for the MR it looks much better!

I allowed myself some nitpicking. Feel free to ignore them if you are not confortable with them.

If you have some time, it might be also great to add some integration tests to `lib/tests/store.rs` in order to check that the read-only and secondary modes are working properly. This tests are also instrumented using LLVM address sanitizer in the CI so having them might allow us to catch some usual C mistakes (memory leaks, use-after-free, double-free...).


Thank you again!

lib/src/storage/backend/rocksdb.rs
```diff
@@ -619,16 +817,24 @@ pub struct Reader {
 enum InnerReader {
     Snapshot(Rc<InnerSnapshot>),
     Transaction(Weak<*mut rocksdb_transaction_t>),
+    Secondary(Arc<DbHandler>), // TODO: we should not hold the whole enum, because we have to use
+                               // secondary_db_handler_or_error to access rocksdb_t. we should
+                               // safe it directly?!
```
Tpt, line 822: Maybe a bad idea (and probably not for this PR), change the types to something like:
```rust
pub enum DB {
    ReadWrite(Arc<RwHandler>),
    ReadOnly(Arc<RoHandler>),
}

struct RwDbHandler { // Same as the previous struct
    db: *mut rocksdb_transactiondb_t,
    options: *mut rocksdb_options_t,
    transaction_options: *mut rocksdb_transaction_options_t,
    transactiondb_options: *mut rocksdb_transactiondb_options_t,
    read_options: *mut rocksdb_readoptions_t,
    write_options: *mut rocksdb_writeoptions_t,
    flush_options: *mut rocksdb_flushoptions_t,
    env_options: *mut rocksdb_envoptions_t,
    ingest_external_file_options: *mut rocksdb_ingestexternalfileoptions_t,
    compaction_options: *mut rocksdb_compactoptions_t,
    block_based_table_options: *mut rocksdb_block_based_table_options_t,
    column_family_names: Vec<&'static str>,
    cf_handles: Vec<*mut rocksdb_column_family_handle_t>,
    cf_options: Vec<*mut rocksdb_options_t>,
    path: PathBuf,
    in_memory: bool
}


struct RoDbHandler {
    db: *mut rocksdb_t,
    options: *mut rocksdb_options_t,
    read_options: *mut rocksdb_readoptions_t,
    env_options: *mut rocksdb_envoptions_t,
    column_family_names: Vec<&'static str>,
    cf_handles: Vec<*mut rocksdb_column_family_handle_t>,
    cf_options: Vec<*mut rocksdb_options_t>,
    path: PathBuf,
    secondary_path: Option<Path> // Maker of secondary vs read-only. TODO: this is not nice
}
```

lib/src/storage/backend/rocksdb.rs
BenediktSeidl: i tried, but `PathBuf` does not implement `Copy`. Not sure how to continue

lib/src/storage/backend/rocksdb.rs
BenediktSeidl: i like the second options, but that would add quite some noise to the Reader-implementation? each function would have to be prepended with an error checking function, right?
maybe we can do this in an additional pr?

lib/src/storage/backend/rocksdb.rs
BenediktSeidl: I don't fully get it. Maybe we can discuss this in an issue or new pull request?

lib/src/storage/backend/rocksdb.rs
Tpt: That's a good point. I missed that. Sorry for the bad comment.

lib/src/storage/backend/rocksdb.rs
Tpt: Yes, definitely!

lib/src/storage/backend/rocksdb.rs
Tpt: Indeed, it would be quite cumbersome. After thinking about it again, a maybe not too bad approach is maybe to just ignore the errors here by just removing the `.unwrap()`. If things are very bad the following  operation will error anyway. There is already the "dirty read" problem so reading also old data is maybe ok too.

Tpt (approved):
Thank you for the updates. I'm planning to publish a new patch release with the changes that are currently in the main branch and then merge this MR.

BenediktSeidl:
i tried to fix the clippy warnings and the typo

BenediktSeidl:
sorry about the formatting (again). `address_sanitizer` will still fail, will need some time to investigate if i should fix this myself.

Tpt:
I am not sure where the memory leak comes from. I believe a good approach is to try running the tests with Valgrind.

BenediktSeidl:
it was easy to spot after I had `llvm-symbolizer` available in my `$PATH`.

Tpt:
> it was easy to spot after I had llvm-symbolizer available in my $PATH.

Great! I have just pushed #385 to add it to the CI

PR #330, opened 2022-12-22
String formatting: uses inline syntax

Tpt:


Tpt:
> did you use the clippy's --fix for this?

@nyurik Yes, exactly. It's a very convenient tool.

nyurik:
Glad it has helped :)

https://github.com/rust-lang/rust-clippy/pull/9233

Tpt:
@nyurik Thank you so much for that. It helped me save quite a lot of tedious dumb changes.

PR #347, opened 2023-01-04
CI: Attempt to cache Cargo cache

Tpt:


adamreichold:
I think you might see much better results using https://github.com/Swatinem/rust-cache

Tpt:
> I think you might see much better results using https://github.com/Swatinem/rust-cache

Thank you! Sadly a huge chuck of the compile time is caused by RocksDB that is included in Oxigraph Cargo workspace so I am not sure it's going to help because Swatinem/rust-cache won't cache it.

adamreichold:
> > I think you might see much better results using https://github.com/Swatinem/rust-cache
> 
> Thank you! Sadly a huge chuck of the compile time is caused by RocksDB that is included in Oxigraph Cargo workspace so I am not sure it's going to help because Swatinem/rust-cache won't cache it.

Couldn't you just put `oxrocksdb-sys` into a separate workspace by adding an empty `[workspace]` table to its `Cargo.toml`? You should still be able to use it as a path dependency. The only downside I see is that you will get a separate target directory when you work on the bindings themselves.

Tpt:
> Couldn't you just put oxrocksdb-sys into a separate workspace by adding an empty [workspace] table to its Cargo.toml? You should still be able to use it as a path dependency. The only downside I see is that you will get a separate target directory when you work on the bindings themselves.

That's a great idea! Thank you!

PR #393, opened 2023-02-23
Improves read-only and secondary store APIs

Tpt:


Tpt:
@BenediktSeidl  I allowed myself to tweak a bit the public API for read-only and secondary stores. I also implemented the "get" method and imporved test coverage. Feedbacks welcomed!

BenediktSeidl:
thanks for the ping :-) looks good for me. the only thing i'm not sure about is changing the meaning of the `--location` argument.

before your change, those three commands would start three oxigraph server with the different options:

```
oxigraph_server serve --location /var/lib/oxigraph/data
oxigraph_server serve --location /var/lib/oxigraph/data --readonly -b "localhost:7778"
oxigraph_server serve --location /var/lib/oxigraph/data --secondary-location /tmp/secondary_location -b "localhost:7777"
```

the same after your change (not 100% sure if i got this right?):

```
oxigraph_server serve --location /var/lib/oxigraph/data
oxigraph_server serve --location /var/lib/oxigraph/data --readonly -b "localhost:7778"
oxigraph_server serve --primary-location /var/lib/oxigraph/data --location /tmp/secondary_location -b "localhost:7777"
```

* in order to activate secondary-mode, you have to specify `--primary-location` (this seem counter-intuitive for me)
* `--location` is used as "primary-writes-data-to-this-folder" for primary and read-only, but not for "secondary" (this makes it hard to c&p command lines: after you figured how to run oxigraph, you may want to try the secondary mode, but then you have to go back and rewrite your whole command)

maybe it would be better to split the `serve` command into three commands: `serve`, `serve-readonly`, and `serve-secondary`? this could make things more clear.

Tpt:
> maybe it would be better to split the serve command into three commands: serve, serve-readonly, and serve-secondary? this could make things more clear.

@BenediktSeidl Thank you! It definitely makes a lot of sense. This way we have a clean way to enforce things like "load action must open the store read-write" and "dump and backup actions only needs to open in a read-only way". I have updated the PR

PR #457, opened 2023-04-12
Adds a simple explanation with profiling for SPARQL queries evaluation

Tpt:
Only exposed in the CLI `query` command for now.

Example to get a flamegraph:
```bash
oxigraph_server query --location MY_DATA --query-file MY_QUERY --results-format csv --explain-file EXPLANATION.json --stats
python3 bench/explanation_to_flamegraph.py EXPLANATION.json FLAMEGRAPH.svg
```
example flamegraph

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -932,6 +932,8 @@ dependencies = [
  "digest",
  "getrandom",
  "hex",
+ "js-sys",
+ "json-event-parser",
  "lazy_static",
  "libc",
  "md-5",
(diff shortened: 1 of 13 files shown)


hobofan:
This is amazing!

FYI: The dependency on `flamegraph.pl` could be replaced with the [inferno](https://crates.io/crates/inferno) crate (the `inferno-flamegraph` binary serves as a drop-in replacement and has a few more styling options).

Tpt:
> This is amazing!

Thank you

> FYI: The dependency on `flamegraph.pl` could be replaced with the [inferno](https://crates.io/crates/inferno) crate (the `inferno-flamegraph` binary serves as a drop-in replacement and has a few more styling options).

Great idea. I just pushed a commit to this MR that makes the flamegraph script use inferno when available.

PR #469, opened 2023-04-20
Start 0.3.16 dev

Tpt:


hobofan:
May I ask if there is a specific reason for doing this? (Maybe we can find another way to achieve what you are looking for?)

Bumping the version in the working branch makes it harder to test out new changes, as the Cargo patch mechanism doesn't consider e.g. version `0.1.2-rc.0` to be a valid patch to the published `0.1.2` (Instead it will usually lead to having the crate present multiple times in the dependencies with distinct version numbers).

PR #470, opened 2023-04-20
Fix missing GSPO index compaction

hobofan:
This fixes a missing compaction of the GSPO index.

With dataset sizes and number of graphs growing, we've recently been seeing degrading query performance for some of our query workloads (made apparent via the new explain feature!). As we have a lot of point-like lookups (via GraphPatterns with all positions known) in our query plans, and this is the index chosen for those, we are probably disproportionally impacted by this.

With this fix a lot of our query times have decreased by ~50%.


Tpt (approved):
Thank you for having spotted this mistake!

PR #476, opened 2023-04-26
OxRDF: Add extra literal escaping to generate **canonical** N-Triples and N-Quads

yamdan:
Hi @Tpt, thank you for your excellent work!

I recently utilized your `oxigraph.io` and `oxrdf` as an N-Quads parser and RDF data structure for [my Rust implementation](https://github.com/yamdan/rdf-canon-rust) of the [W3C RDF Dataset Canonicalization](https://github.com/w3c/rdf-canon/) specification.

In this PR, I've introduced additional literal escaping to ensure compliance with the [canonical N-Triples](https://w3c.github.io/rdf-n-triples/spec/#canonical-ntriples) and [canonical N-Quads](https://w3c.github.io/rdf-n-quads/spec/#canonical-quads) specifications. Although these canonical specifications are still in Editor's Draft status, I believe that appropriately escaping these C0 control characters will contribute to a more secure and stable serialization across various use cases.

All tests have passed. If my understanding or usage of Oxigraph is incorrect in any way, please feel free to leave comments on this PR.


lib/oxrdf/src/literal.rs
```diff
@@ -620,10 +620,41 @@ pub fn print_quoted_str(string: &str, f: &mut impl Write) -> fmt::Result {
...
+            '\u{16}' => f.write_str("\\u0016"),
+            '\u{17}' => f.write_str("\\u0017"),
+            '\u{18}' => f.write_str("\\u0018"),
+            '\u{19}' => f.write_str("\\u0019"),
+            '\u{1a}' => f.write_str("\\u001A"),
+            '\u{1b}' => f.write_str("\\u001B"),
+            '\u{1c}' => f.write_str("\\u001C"),
+            '\u{1d}' => f.write_str("\\u001D"),
+            '\u{1e}' => f.write_str("\\u001E"),
+            '\u{1f}' => f.write_str("\\u001F"),
```
Tpt, line 654: Nit: to avoid the big `match`, use a single option like `'\0'..='\u{1f}' | '\u{7f}' => write!(f, "\\u{:04X}", u32::from(c))`

Tpt (approved):
Thank you! It's very welcomed. I have been planning to do it but never took the time to do so. I just have one nitpick suggestion.

Great to hear that you are working on a canonicalization implementation in Rust. I hope to switch [`Dataset::canonicalize`](https://docs.rs/oxrdf/0.1.5/oxrdf/dataset/struct.Dataset.html#method.canonicalize) to the standardized algorithm in the future.

Tpt:
I am also asking myself a question: should we ship this change in a minor release (0.3.15) or consider it breaking and wait for 0.4? This change might break some usages already relying on a stable serialization, especially because of the commonly used `\t` character.

lib/oxrdf/src/literal.rs
yamdan: Thank you for your suggestion! It seems completely reasonable. I applied it.

yamdan:
> This change might break some usages already relying on a stable serialization, especially because of the commonly used \t character.

Yes, if there exists some quads including horizontal tab (U+0009) for their visual representation in N-Quads layer, such as:
```nquads
<urn:ex:s> <urn:ex:p> "a    literal    with    tabs" .
```
(here four whitespaces should be intepreted as a single tab(U+0009))

Then, after this PR it will be breaking-changed to:
```nquads
<urn:ex:s> <urn:ex:p> "a\tliteral\twith\ttabs" .
```

However, the application can obtain the former bytes (`"a      literal      with       tabs"`) by parsing the latter N-Quads string (`"a\tliteral\twith\ttabs"`) just using [RDF 1.1 N-Quads - 5. Parsing](https://www.w3.org/TR/n-quads/#sec-parsing) rule. Hence, the impact could be not so huge if I understand correctly...

Tpt:
@yamdan Thank you for your reply! I believe the breaking use cases would be people serializing to NTriples and send using shell scripts to parse the RDF output or people using the serialization for hashing (oxrdf already provides a canonicalization implementation). I plan to do a 0.4 release soon so I would feel better to ship this change as part of it. Is it fine with you?

Tpt:
I am merging into the `next` branch that will become the 0.4 release. Thank you so much for this PR!

PR #478, opened 2023-04-27
Codespell: config + workflow

yarikoptic:
Congrats -- no new typos found!  But log has a number of commits to fix typos so I thought it might still be useful to bolt codespell on to ensure that it stays typos free.  No pressure -- feel free to close if not interested and love surprises.

PR #479, opened 2023-04-28
add shellcheck checks and robustify found shell scripts

yarikoptic:
I guess I took https://github.com/oxigraph/oxigraph/pull/478#issuecomment-1526249541 too much as encouragement for more action! ;)  In all my scripts I use `set -eu` to make them behave like any other program in a "proper" language - exit upon error or undefined variable if no explicit handling was coded in.  That addressed some shellcheck warnings and then I addressed the rest.  Individual commits have more explanation etc.


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y shellcheck
+      - name: shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' | xargs shellcheck
```
Tpt, line 23: I have put all the tests in the [tests.yml](https://github.com/oxigraph/oxigraph/blob/main/.github/workflows/tests.yml) file. Could you add it there too?

Tpt (approved):
Thank you! I just have a nitpick.

.github/workflows/shellcheck.yml
yarikoptic: done!

PR #480, opened 2023-04-28
Make all Thomases into one

yarikoptic:
Before

    ❯ git shortlog -sn | head
      1211	Tpt
        46	Thomas Tanon
     ...

After

    ❯ git shortlog -sn | head
      1259	Thomas Tanon
      ...


Tpt (approved):
Thank you!

PR #506, opened 2023-05-21
XSD: Drops nom dependency and adds checked_neg operation

Tpt:


lib/oxsdatatypes/src/parser.rs
```diff
@@ -125,412 +83,525 @@ impl From<DateTimeError> for XsdParseError {
...
+    let (negative, input) = if let Some(left) = input.strip_prefix('-') {
+        (true, left)
     } else {
-        Err(XsdParseError {
-            kind: XsdParseErrorKind::TooMuchData { count: left.len() },
-        })
+        (false, input)
+    };
+    let mut input = expect_char(input, 'P', "Durations must start with 'P'")?;
+    let mut state = 0; // after_year = 1, after_month = 2, after_days = 3, after_T = 4, after_hours = 5, after_minutes = 6, after_seconds = 7
```
adamreichold, line 108: I understand that an `enum`-based approach would make the less-than comparison below unergonomic, but maybe named constants for the values would be useful still?

lib/oxsdatatypes/src/parser.rs
```diff
@@ -125,412 +83,525 @@ impl From<DateTimeError> for XsdParseError {
...
-// [62]   endOfDayFrag ::= '24:00:00' ('.' '0'+)?
-fn end_of_day_frag(input: &str) -> XsdResult<'_, (u8, u8, Decimal)> {
-    map(
-        recognize(tuple((
-            tag("24:00:00"),
-            opt(preceded(char('.'), many1(char('0')))),
-        ))),
-        |_| (24, 0, 0.into()),
-    )(input)
+fn safe_parse<T>(
```
adamreichold, line 547: The name seems more general than the functionality. Maybe something liek `ensure_complete`?

lib/oxsdatatypes/src/parser.rs
```diff
@@ -125,412 +83,525 @@ impl From<DateTimeError> for XsdParseError {
...
-            char('-'),
-            char('-'),
-            char('-'),
-            day_frag,
-            opt(timezone_frag),
-        )),
-        |(_, _, _, day, timezone)| GDay::new(day, timezone),
-    )(input)
+fn g_day_lexical_rep(input: &str) -> Result<(GDay, &str), XsdParseError> {
+    let input = expect_char(input, '-', "gDay values must start with '--'")?;
```
adamreichold, line 389: Should the error message be `"gDay values must start with '---'"` with three dashes?

lib/oxsdatatypes/src/parser.rs
```diff
@@ -125,412 +83,525 @@ impl From<DateTimeError> for XsdParseError {
...
+        if c.is_ascii_digit() {
+            // Ok
+        } else if c == '.' && !dot_seen {
+            dot_seen = true;
+        } else {
+            end = i;
+            break;
+        }
     }
+    (&input[..end], &input[end..])
```
adamreichold, line 594:
```suggestion
    input.split_at(end)
```

lib/oxsdatatypes/src/parser.rs
Tpt: Indeed. It would be way more readable. Done. Thank you!

Tpt:
@adamreichold Thank you for your very nice suggestions!

PR #509, opened 2023-05-26
Speeds up sparql_eval fuzzer

Tpt:
For some reasons explicit Arbitrary call seems way faster (95 iter/s after the change, less than 0.2 iter/s before the change)

### DIFF
--- a/lib/sparql-smith/src/lib.rs
+++ b/lib/sparql-smith/src/lib.rs
@@ -1,6 +1,5 @@
 use arbitrary::{Arbitrary, Result, Unstructured};
 use std::fmt;
-use std::fmt::Debug;
 use std::iter::once;
 use std::ops::ControlFlow;
 
@@ -30,8 +29,12 @@ const LITERALS: [&str; 11] = [
     "1e0",
 ];
 
-#[derive(Arbitrary)]
 pub struct Query {
+    inner: QueryContent,
+}
+
+#[derive(Arbitrary)]
+struct QueryContent {
     // [1]  	QueryUnit	  ::=  	Query
     // [2]  	Query	  ::=  	Prologue ( SelectQuery | ConstructQuery | DescribeQuery | AskQuery ) ValuesClause
     variant: QueryVariant,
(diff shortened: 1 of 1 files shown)


adamreichold:
Are you sure the improvement is not "pyrrhic" as it avoids steps the `fuzz_target!` macro does to improve the overall exploration, e.g. https://github.com/rust-fuzz/libfuzzer/blob/1221c356e993b9f82d1ccd152f1c7636468758d2/src/lib.rs#L262

Tpt:
> Are you sure the improvement is not "pyrrhic" as it avoids steps the fuzz_target! macro does to improve the overall exploration, e.g. https://github.com/rust-fuzz/libfuzzer/blob/1221c356e993b9f82d1ccd152f1c7636468758d2/src/lib.rs#L262

Thank you so much for the explanation! However, with this change I manage to get relevant crashes within minutes whereas I was not able to get any within hours before it. Not sure why the auto `Arbitrary` derive gives such a large min size. Maybe playing with libfuzzer `len_control` might help circumvent the problem.

Tpt:
I played with `-len_control` and `-max_len`. Sadly, it did not seemed to help. The `size_hint` returned by `Query` is 20 because the structure contains a lot of `Option` fields. I believe it's fine to use input smaller than 20 because it only means that some options will stay `None` with these inputs and it's probably ok because it still generates interesting queries. Inputs are still relevant even without e.g. `ORDER BY`, `GROUP BY`, `LIMIT`, `OFFTSET`, and `VALUES`.

adamreichold:
Maybe you could wrap query in side a newtype to augment its `Arbitrary` impl to overide the `size_hint` with something more useful?

Tpt:
New surprise: when derived by `Arbitrary`, Query::size_hint` takes a huge time. I got once 469s of execution time and an other time 20s. Hence the very slow iteration time. After harcoding the expected value (20) I get around 1250 iter/s.

adamreichold:
> New surprise: when derived by `Arbitrary`, Query::size_hint` takes a huge time. I got once 469s of execution time and an other time 20s. Hence the very slow iteration time. After harcoding the expected value (20) I get around 1250 iter/s.

If this happens in optimized builds, this sounds like a bug in `Arbitrary`'s derive macro?

Tpt:
> If this happens in optimized builds, this sounds like a bug in `Arbitrary`'s derive macro?

I believe it's because the SPARQL structure is very recursive and `size_hint` walks through the tree. `Abitrary` sets [a gard]( https://github.com/rust-fuzz/arbitrary/blob/main/src/size_hint.rs#LL12C8-L12C23) to avoid infinite recursion but the bound is at 20 nested call and it's enough to build a pretty big tree.

PR #512, opened 2023-05-31
Python: fixes parsing from text I/O with not-ASCII char

Tpt:
Python text IO is counting by number of character
and not by number of bytes.
However, we were asking to read a number of bytes
and not a number of character,
leading to strange I/O exceptions being raised.

### DIFF
--- a/python/src/io.rs
+++ b/python/src/io.rs
@@ -8,7 +8,8 @@ use oxigraph::io::{
 use pyo3::exceptions::{PyIOError, PySyntaxError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::types::PyBytes;
-use pyo3::wrap_pyfunction;
+use pyo3::{intern, wrap_pyfunction};
+use std::cmp::max;
 use std::error::Error;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader, BufWriter, Cursor, Read, Write};
@@ -282,17 +283,22 @@ impl Write for PyWritable {
 pub struct PyIo(PyObject);
 
 impl Read for PyIo {
-    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         Python::with_gil(|py| {
+            if buf.is_empty() {
+                return Ok(0);
+            }
+            let to_read = max(1, buf.len() / 4); // We divide by 4 because TextIO works with number of characters and not with number of bytes
             let read = self
                 .0
-                .call_method(py, "read", (buf.len(),), None)
+                .as_ref(py)
+                .call_method1(intern!(py, "read"), (to_read,))
                 .map_err(to_io_err)?;
             let bytes = read
-                .extract::<&[u8]>(py)
-                .or_else(|e| read.extract::<&str>(py).map(str::as_bytes).map_err(|_| e))
+                .extract::<&[u8]>()
+                .or_else(|e| read.extract::<&str>().map(str::as_bytes).map_err(|_| e))
                 .map_err(to_io_err)?;
-            buf.write_all(bytes)?;
+            buf[..bytes.len()].copy_from_slice(bytes);
             Ok(bytes.len())
         })
     }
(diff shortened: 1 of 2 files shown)


python/src/io.rs
```diff
@@ -284,13 +285,18 @@ pub struct PyIo(PyObject);
...
         Python::with_gil(|py| {
+            if buf.is_empty() {
+                return Ok(0);
+            }
+            let to_read = max(1, buf.len() / 4); // We divide by 4 because TextIO works with number of characters and not with number of bytes
             let read = self
                 .0
-                .call_method(py, "read", (buf.len(),), None)
+                .as_ref(py)
+                .call_method1("read", (to_read,))
```
adamreichold, line 295: I understand that this is somewhat unrelated to the issue at hand, but as performance trick, I would suggest either do the method look-up once using `getattr`, storing the result in `PyIo` and using `call1` on it. Possibly using a new type distinct from `PyIo` if you want to support the `read` Python method changing during I/O processing.

Depending on how this is used, you could even have `PyIo<'py>(&'py PyAny)` to avoid the overhead of the `with_gil`, but of course, this would imply you cannot release the GIL during the whole operation.

What can also do with the current structure is to use the [`intern!`](https://docs.rs/pyo3/latest/pyo3/macro.intern.html) macro to avoid allocating a `PyString` for `"read"` for each call.

adamreichold: (Since someone else was asking about this, I wonder whether we should provide such `io::Read` and `io::Write` wrappers in PyO3 directly. But the additional copying required seems to not make this universally useful, e.g. if the Python-side is backed by a file descriptor and one does not want to support sharing partially consumed buffers between Python and Rust, it is much more efficient to just pass the file descriptor to Rust and use `File::from_raw_fd`.)

python/src/io.rs
```diff
@@ -284,13 +285,18 @@ pub struct PyIo(PyObject);
...
                 .map_err(to_io_err)?;
             let bytes = read
-                .extract::<&[u8]>(py)
-                .or_else(|e| read.extract::<&str>(py).map(str::as_bytes).map_err(|_| e))
+                .extract::<&[u8]>()
+                .or_else(|e| read.extract::<&str>().map(str::as_bytes).map_err(|_| e))
                 .map_err(to_io_err)?;
             buf.write_all(bytes)?;
```
adamreichold, line 301: Not sure if this would really have helped with the issue, but this could actually be `.unwrap()` instead of `?` since it should never fail. (In this particular case, this would mean that users would have reported panic exceptions, i.e. `BaseException`s, instead of "random I/O errors". Not sure if this would have been nicer for your users, but it would certainly have been hard to ignore.)

python/src/io.rs
Tpt: Thank you so much for the performance tricks!

- `getattr` seems a great idea, I am just a bit worried it's error prone if someone hacks around the `read` method. This method is already quite heavy so I don't know how much it's worth it if it only allows to earn a few % of performance. People reading big files should not use this code anyway but give onl the file path and let the Rust code do all the work.
- I am not sure about `PyIo<'py>(&'py PyAny)` because parsing is quite heavy and, so, releasing the GIL seems worth it.
- I have added the `intern!` calls, thank you for the trick!
- A shared `io::Read` and `io::Write` wrapper looks a great idea. It might be usable in a few places. But indeed there are a lot of "details" like text IO vs bytes IO, file descriptor optimizations just like you said...

python/src/io.rs
Tpt: Good point. Thank you! Done.

python/src/io.rs
adamreichold: > getattr seems a great idea, I am just a bit worried it's error prone if someone hacks around the read method. This method is already quite heavy so I don't know how much it's worth it if it only allows to earn a few % of performance. People reading big files should not use this code anyway but give onl the file path and let the Rust code do all the work.

Indeed it is a trade-off of flexibility/dynamism versus efficiency and hence should not be implemented without measuring the gain.

IIRC the Python interpreter does more and more of this internally since attribute look-ups especially when involving method resolution order can be computationally expensive, but then again the interpreter is in a position to find out when the `read` attribute has changed so it isn't a trade-off for it. (It would be nice if there was a C API exposing this as generic notification so that we could provide such caching method slots in PyO3, but alas it is all deeply internal to CPython.)

And yes, if the underlying I/O call to going to spinning rust or even something like an S3 bucket, then the overhead of the attribute look-up will pale in comparison. With fast I/O devices like local NVMe disk, I suspect that it might be distributed fat though.

python/src/io.rs
```diff
@@ -302,16 +308,17 @@ impl Write for PyIo {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         Python::with_gil(|py| {
             self.0
-                .call_method(py, "write", (PyBytes::new(py, buf),), None)
+                .as_ref(py)
+                .call_method1(intern!(py, "write"), (PyBytes::new(py, buf),))
                 .map_err(to_io_err)?
-                .extract::<usize>(py)
+                .extract::<usize>()
```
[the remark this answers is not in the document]
Tpt: This implementation of write does not support text I/O yet. Most text IO implementations will raise an exception when given a string. Doing something like (bad Python+Rust speudo-code):
```python
if isinstance(self.0, io.TextIOBase):
    self.0.write(buf.decode())
else:
    self.0.write(buf)
```
might work (with the `isinstance` call cached)

PR #539, opened 2023-06-12
Upgrades MSRV to 1.65

Tpt:


lib/oxrdf/src/dataset.rs
```diff
@@ -705,7 +705,9 @@ impl Dataset {
         InternedTerm,
         InternedGraphName,
     )> {
-        let b_prime = partition.iter().find_map(|(_, b)| (b.len() > 1).then(|| b));
+        let b_prime = partition
+            .iter()
+            .find_map(|(_, b)| (b.len() > 1).then_some(b));
```
adamreichold, line 710: Maybe this would be simpler by splitting peeling off the pair and filtering, e.g.

```rust
let b_prime = partition.iter().map(|(_, b)| b).find(|b| b.len() > 1);
```

I also think the call to `fold` a few lines below could be simplified by calling [`reduce`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.reduce) instead.

Tpt: Thank you! Done. It's indeed much better.

PR #562, opened 2023-07-06
Always print xsd:float and xsd:double with exponential notation

Tpt:
Follows a bit more the XML schema canonical mapping


Tpt:
This break a few SPARQL test. It seems that the regular notation is more used anyway. Closing this PR.

Issue #471, opened 2023-04-21
Target WASM runtimes where `wasm-bindgen` is not available


Luca8991:
**Is your feature request related to a problem? Please describe.**
I'm trying to run an RDF database and expose a SPARQL endpoint on an [Internet Computer canister](https://internetcomputer.org/how-it-works/canister-lifecycle/). For this reason, the features I'd need the most from **oxigraph** are the **database** and the **SPARQL parser**.
However, but I keep getting this error when deploying it:
```
Module imports function '__wbg_getRandomValues_3774744e221a22ad' from '__wbindgen_placeholder__' that is not exported by the runtime.
```
From what I understand, this means that the canister runtime doesn't expose `wasm-bindgen` library.

**Describe the solution you'd like**
I'm not that expert in Rust and compiling it to WASM, hence I don't know exactly if it would be possible to avoid the usage of `wasm-bindgen` and what effort it could it take.

**Describe alternatives you've considered**
I've tried [sophia](https://github.com/pchampin/sophia_rs), which compiles and runs smoothly, but is missing the SPARQL parser. Is there a way to use oxigraph's parser with sophia?

**Additional context**
I'd suggest you to dive into the [Internet Computer documentation](https://internetcomputer.org/docs), because I see a great potential here to enable decentralized RDF databases that can have various applications like decentralized social media platforms, etc.

hobofan:
The issue is not with `wasm-bindgen` (you will likely need that if your compile target is WASM), the problem is with the usage of `getrandom` for generating random values for Blank Nodes.

See https://docs.rs/getrandom/latest/getrandom/#webassembly-support for more context. It looks like your WASM runtime environment (the IC canister) doesn't expose a `Crypto.getRandomValues` function to the WASM module.

AFAIK if the `js` feature flag would be removed here (https://github.com/oxigraph/oxigraph/blob/main/lib/Cargo.toml#L51-L53), it would be possible to register a custom random number generator for non-JS WASM runtimes. (I guess oxigraph would need to introduce a feature flag that's turned on by default to allow this in a backwards compatible manner).

From what I know from working with Web3 systems in the past, random numbers are notoriously hard/slow to generate in the runtimes of those systems (I would assume that IC is no different there). So if you want to embark further on this topic, there is a good chance that this could be quite complicated and/or may not yield a result that has a good performance profile.

Luca8991:
Thanks @hobofan!
[...]

PR #575, opened 2023-07-13
Stop assuming JS platform when compiling to wasm32-unknown-unknown

Tpt:
- Adds the "js" feature to enable JS support
- Adds the "custom-now" feature to oxsdatatypes to inject a custom "now" implementation It is already possible for random with the getrandom "custom" feature

Issue #471

### DIFF
--- a/lib/oxsdatatypes/README.md
+++ b/lib/oxsdatatypes/README.md
@@ -32,6 +32,22 @@ Each datatype provides:
 * `from_be_bytes` and `to_be_bytes` methods for serialization.
 
 
+### `DateTime::now` behavior
+
+The `DateTime::now()` function needs special OS support.
+Currently:
+- If the `custom-now` feature is enabled, a function computing `now` must be set:
+  ```rust
+  use oxsdatatypes::{DateTimeError, Duration};
+  
+  #[no_mangle]
+  fn custom_ox_now() -> Result<Duration, DateTimeError> {
+    unimplemented!("now implementation")
+  }
+  ```
+- For `wasm32-unknown-unknown` if the `js` feature is enabled the `Date.now()` ECMAScript API is used.
+- For all other targets `SystemTime::now()` is used.
+
 ## License
 
 This project is licensed under either of
--- a/lib/oxsdatatypes/src/date_time.rs
+++ b/lib/oxsdatatypes/src/date_time.rs
@@ -191,7 +191,7 @@ impl DateTime {
     pub fn checked_sub_day_time_duration(self, rhs: impl Into<DayTimeDuration>) -> Option<Self> {
         let rhs = rhs.into();
         Some(Self {
-            timestamp: self.timestamp.checked_sub_seconds(rhs.all_seconds())?,
+            timestamp: self.timestamp.checked_sub_seconds(rhs.as_seconds())?,
         })
     }
 
@@ -1757,7 +1757,22 @@ impl Timestamp {
     }
 }
 
-#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+#[cfg(feature = "custom-now")]
+#[allow(unsafe_code)]
+pub fn since_unix_epoch() -> Result<Duration, DateTimeError> {
+    extern "Rust" {
+        fn custom_ox_now() -> Result<Duration, DateTimeError>;
+    }
+
+    unsafe { custom_ox_now() }
+}
+
+#[cfg(all(
+    feature = "js",
+    not(feature = "custom-now"),
+    target_family = "wasm",
+    target_os = "unknown"
+))]
 fn since_unix_epoch() -> Result<Duration, DateTimeError> {
     Ok(Duration::new(
         0,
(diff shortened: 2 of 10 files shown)


lib/oxsdatatypes/src/date_time.rs
```diff
@@ -1766,7 +1769,35 @@ fn since_unix_epoch() -> Result<Duration, DateTimeError> {
...
+#[cfg(all(feature = "custom-now", any(target_os = "unknown", target_os = "none")))]
+static GLOBAL_NEW: atomic::AtomicPtr<fn() -> Result<Duration, DateTimeError>> =
+    atomic::AtomicPtr::new(ptr::null_mut());
+
+/// Defines an implementation for fetching the current time.
+///
+/// Only enabled if there is no reachable operating system (target_os is unknown or none).
+#[cfg(all(feature = "custom-now", any(target_os = "unknown", target_os = "none")))]
+pub fn register_custom_now(now: fn() -> Result<Duration, DateTimeError>) {
+    GLOBAL_NEW.store(Box::leak(Box::new(now)), atomic::Ordering::Release);
```
adamreichold, line 1781: Do you really need acquire-release ordering guarantees between calls to `register_custom_now` and `since_unix_epoch`? Wouldn't `Ordering::Relaxed` suffice?

adamreichold: (It also looks strange to box up the function pointer. But I suspect you are considering platforms where function and data pointers cannot be interchanged? Otherwise `AtomicPtr<()>` and transmuting the function pointer to `*mut ()` would probably suffice?)

lib/oxsdatatypes/src/date_time.rs
```diff
@@ -1766,7 +1769,35 @@ fn since_unix_epoch() -> Result<Duration, DateTimeError> {
...
+}
+
+#[cfg(all(feature = "custom-now", any(target_os = "unknown", target_os = "none")))]
+#[allow(unsafe_code)]
+pub fn since_unix_epoch() -> Result<Duration, DateTimeError> {
+    let function = GLOBAL_NEW.load(atomic::Ordering::Acquire);
```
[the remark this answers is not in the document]
adamreichold: (This would also replace the runtime error `NoNow` with a linker error asking for `custom_now`.)

lib/oxsdatatypes/README.md
```diff
@@ -32,6 +32,13 @@ Each datatype provides:
 * `from_be_bytes` and `to_be_bytes` methods for serialization.
 
 
+## Calendar value now operation
+
+The `DateTime::now()` function needs special OS support.
+Currently:
+- For `wasm32-unknown-unknown` if the `js` feature is enabled the `Date.now()` ECMAScript API is used.
+- For all other targets `SystemTime::now()` is used.
```
adamreichold, line 40: Shouldn't the option to use a custom now function via the `custom-now` target be the third bullet point here?

lib/oxsdatatypes/src/date_time.rs
Tpt: > Wouldn't Ordering::Relaxed suffice?

I believe there might be cases where the caller of `register_custom_now` sets some global variables used by the `now` function just before calling `register_custom_now` so it might be good to make sure they are properly accessible from the thread calling `now`. But I might be wrong, I am not an expert on memory ordering.

> But I suspect you are considering platforms where function and data pointers cannot be interchanged?

Not consciously. I just wanted a correct and easy to implement approach. In Oxigraph practical usages outside profiling, the `new` method is called only at most once per SPARQL query so I don't care much about performances. But being able to work on such platforms is nice. My main goal is to allow the author of [this fork](https://github.com/omnia-network/ic-oxigraph) to use upstream Oxigraph.

lib/oxsdatatypes/src/date_time.rs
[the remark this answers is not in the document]
Tpt: Thank you! It looks much nicer for my use case! I am going to investigate this.

lib/oxsdatatypes/README.md
Tpt: Indeed. I forgot about it.

lib/oxsdatatypes/src/date_time.rs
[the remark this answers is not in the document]
Tpt: Indeed, it's much better. I have implemented it.

lib/oxsdatatypes/README.md
```diff
@@ -32,6 +32,20 @@ Each datatype provides:
 * `from_be_bytes` and `to_be_bytes` methods for serialization.
 
 
+### `DateTime::now` behavior
+
+The `DateTime::now()` function needs special OS support.
+Currently:
+- For `wasm32-unknown-unknown` if the `js` feature is enabled the `Date.now()` ECMAScript API is used.
+- For all `unknown` or `none` target OSs, if the `custom-now` feature is enabled, a function computing `now` must be set in the root binary crate following:
```
adamreichold, line 40: A nit, but I think `custom_ox_now` can be provided by anything (crate, external library) linked into the final DSO (whether that is a binary or library).

Personally, I also wouldn't limit the `custom-now` feature to `unknown` and `none` operating systems. If the integrator wants a custom now function, let them have it.

adamreichold: (I think the most notable requirement for linkage is that the function uses the Rust ABI and types from the `oxsdatatypes` crate which in practice means it must be compiled by the same Rust compiler as that crate.)

Tpt: Indeed. I have rewritten the documentation and made `custom-now` possible on all target. Thank you!

Issue #239, opened 2022-08-17
Don't build RockDB?


hobofan:
`oxrocksdb-sys` currently builds `rocksdb` from scratch (as far as I can tell), which dominates the build times for the projects in which we are using `oxigraph`. Instead it would be preferable if it can detect and use an already built version of `rocksdb` provided by the system.

As I've seen that `oxrocksdb-sys` is forked from `librocksdb-sys`, which already had some code in place to use a system-provided version, I'm wondering what the reasoning for doing so was. I'd also be interesting to know whether there are any plans to move to "upstream" `librocksdb-sys` and whether that is something that I could possibly help out with?

Tpt:
> Instead it would be preferable if it can detect and use an already built version of rocksdb provided by the system.

Yes, that would be much better indeed.

> As I've seen that oxrocksdb-sys is forked from librocksdb-sys, which already had some code in place to use a system-provided version, I'm wondering what the reasoning for doing so was.

I wanted to customized the C RocksDB bindings to be able to expose to C/Rust some transaction related functions and better error values than just a plain strings. See [this directory](https://github.com/oxigraph/oxigraph/tree/main/oxrocksdb-sys/api). This make stuff like detection of transaction conflicts much easier on the Rust side (no need to use hugly string comparisons).

I would love to reuse directly `librocksdb-sys` but for building these customized bindings I need to link to the C++ code so to simplify the build step I just choose to fork the sys library and tweak it. But if there is a clean way to do otherwise, it would be much better.

hobofan:
Thanks for answering so swiftly and providing some context!

> I wanted to customized the C RocksDB bindings to be able to expose to C/Rust some transaction related functions and better error values than just a plain strings. See [this directory](https://github.com/oxigraph/oxigraph/tree/main/oxrocksdb-sys/api?rgh-link-date=2022-08-17T07%3A03%3A35Z).

I see - that makes sense. I'm not the biggest C/C++ buff, but from what I can tell that is something that actually needs to be compiled into the library, so with that change it actually makes sense not to use a precompiled version of the library.

Tpt:
0.4.0-alpha.1 adds the `rocksdb-pkg-config` feature that uses pkg-config to link with an existing rocksdb binary

PR #626, opened 2023-09-10
Allows dynamic RocksDB linking

Tpt:


hobofan:
Closes https://github.com/oxigraph/oxigraph/issues/239?

Tpt:
@hobofan Yes!

PR #644, opened 2023-10-16
SPARQL: Fix ASK + FROM combination

vtermanis:
Previously `ASK FROM` only worked without `WHERE`.

This did work:
```sparql
ASK FROM <scheme://a.graph> { FILTER(true) }
```

... but this did not:
```sparql
ASK FROM <scheme://a.graph> WHERE { FILTER(true) }
```


Tpt:
Thank you so much for this great PR!

PR #737, opened 2024-01-18
Simplify complex strings with raw literals

nyurik:
Raw strings are much easier to read most of the time


Tpt (approved):
Thank you! It's much better!

PR #738, opened 2024-01-18
Remove unused returns in doc tests

nyurik:
Many doc tests have unused (implied) return statements that are hidden from the docs, and appear to not be needed. So in the spirit of "less is more" (i.e. KISS, and other fun short words), removing to keep things a bit simpler (and so that IDEs don't complain)


nyurik:
spoke too soon, this is actually needed, i wasn't aware of this usecase - i usually just use `.unwrap()` in the doctests, but the `?` is cleaner

Tpt:
Yes, it's quite unfortunate these are needed.

PR #739, opened 2024-01-18
Add WKT_LITERAL vacab support

nyurik:


lib/oxrdf/src/vocab.rs
```diff
@@ -230,4 +230,7 @@ pub mod xsd {
     /// Duration of time (months and years only).
     pub const YEAR_MONTH_DURATION: NamedNodeRef<'_> =
         NamedNodeRef::new_unchecked("http://www.w3.org/2001/XMLSchema#yearMonthDuration");
+    /// Geospatial datatype like `"Point({longitude} {latitude})"^^geo:wktLiteral`
+    pub const WKT_LITERAL: NamedNodeRef<'_> =
+        NamedNodeRef::new_unchecked("http://www.opengis.net/ont/geosparql#wktLiteral");
```
Tpt, line 235: This is not from the `xsd` vocabulary. May you create a new `mod geosparql` to set it? The datatypes `rdf:XMLLiteral` and `rdf:Html` are in the `rdf` module.

nyurik: thx ,fixed

Tpt (approved):

PR #740, opened 2024-01-18
Add debug print to test_debian_compatibility

nyurik:
Apparently CI began to fail all of a sudden - this prints the reason


Tpt (approved):
Thank you! This script is a quick hack to check if Oxigraph is making Debian packagers life worse.

PR #741, opened 2024-01-18
Consolidate dependency management using workspace

nyurik:
Declaring all dependencies in one place ensures they are consistent across the project.

Even though some dependencies might be used only once, in my experience it is far easier to keep them consistently in one place, than to have a separation between shared vs non-shared ones, and also to move them once something becomes shared. Easier to just keep everything together and consistent.


Tpt (commented):
Thank you! 100% agreed it's much better (it has been in my TODO list for some time). Sadly, the way [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) parses `Cargo.toml` file is quite bad and it does not work properly with the directory layout we use here: it thinks that the `lib/Cargo.toml` file is the workspace declaration of `lib/oxrdf/Cargo.toml` crate. 

So, except if we fix cargo-semver-checks we have to choose between using workspace dependencies or semver-checks.

See: https://github.com/obi1kenobi/cargo-semver-checks/issues/462

nyurik:
what do you mean by crates.io metadata? Each crate is independent there, so they wouldn't rely on each other when published, or am i missing something?

Tpt:
> what do you mean by crates.io metadata?

The source code links [example](https://github.com/oxigraph/oxigraph/blob/5be6f55155329d208632a7b949e378c6e4a253b5/lib/oxrdf/Cargo.toml#L8)

nyurik:
so you would have to update that link to?  Or are you saying the existing code on crates.io would point to a wrong location until you publish a new version?

I have seen people using structures like `/xxx` for all crates to live on the same root level, or if you want to separate rust and non-rust, perhaps a `/rust/xxx` or `/crates/xxx` ?.  TBH, I am not sure it is a good idea to combine `/lib` and other crates in the same dir.

Tpt:
> so you would have to update that link to? Or are you saying the existing code on crates.io would point to a wrong location until you publish a new version?

Yes, the link would point to a wrong location until we publish a new version

> I have seen people using structures like /xxx for all crates to live on the same root level, or if you want to separate rust and non-rust, perhaps a `/rust/xxx` or `/crates/xxx` ?. TBH, I am not sure it is a good idea to combine /lib and other crates in the same dir.

What I liked with the current directory tree is that it higlights that there is a "top" crate and then modules in sub-crates that are also exposed by the "top" crate just like [a](https://github.com/PyO3/pyo3) [lot](https://github.com/clap-rs/clap) [of](https://github.com/rust-lang/regex/) [projects](https://github.com/quickwit-oss/tantivy).

nyurik:
Sure - i looked at your [example](https://github.com/quickwit-oss/tantivy) -- they use `/xxx` approach - where all sorts of crates are in the root, plus there is a `/src` dir to indicate that the root itself is a crate.  Or are you saying this approach won't work either?

Tpt:
> Sure - i looked at your [example](https://github.com/quickwit-oss/tantivy) -- they use /xxx approach - where all sorts of crates are in the root, plus there is a /src dir to indicate that the root itself is a crate. Or are you saying this approach won't work either?

If we remove the `lib` directory and move all its content to the root it will work but will overload a bit the root dir. I don't think fixing `semver-checks` would be too complicated, I am incline to just take some time to fix it.

nyurik:
Hi, any luck with the semvercheck? I was reading that bug, and realized that if the issue is with the identically named subdir & crate, would it be possible to simply rename the subdir?

nyurik:
the reason i pinged you on this PR is that sadly it will go stale pretty fast, and I would have to regen it from scratch - so if there is a way to have some temp workaround, would be great.  Up to you of course, if you want to close this and wait for the tooling, but its seems such a pain to manage deps at the moment?

Tpt:
> Hi, any luck with the semvercheck? I was reading that bug, and realized that if the issue is with the identically named subdir & crate, would it be possible to simply rename the subdir?

Sadly not, this is because crates like `oxrdf` are defined in a directory (`lib/oxrdf`) that is included in a directory (`lib`) that contains a `Cargo.toml` file that deos not contain the workspace definition. An approach to fix it might be to move the `oxigraph` crate from the `lib` directory to `lib/oxigraph`. I can do an MR for that (except if you want to do it)

nyurik:
no no, you go ahead - its 3:30am, time to sleep a bit :)

Tpt:
WIP in #768

nyurik:
awesome!  seems to work :)   I did a rebase, seems like there was one proj missing, plus i moved all internal deps into one section - seems like it would be more readable.

PR #742, opened 2024-01-18
Use `Self::AssocName` to simplify declarations

nyurik:
To keep with DRY principle, I think it makes it a bit less redundant to reuse the Self::<associated_type_name> structure in the well known trait implementations - keeps it consistent with the trait decl too.


Tpt (approved):
Thank you!

PR #745, opened 2024-01-18
wip: add WKT support

nyurik:
This is a very early draft, I have no idea where i am going with this. Any feedback / guidance is welcome


Tpt:
Thank you! geosparql support would be amazing.

I would maybe not include the geosparql types in the `oxsdatatypes` crate that is focused on XML Schema datatypes but in a new crate (`oxgeodatatype`?).

- If you want native support in the storage layer the main thing to change is adding an entry in the [`EncodedTerm`](https://github.com/oxigraph/oxigraph/blob/522bda29066c7c340938979b6f1d8af352230616/lib/src/storage/numeric_encoder.rs#L43) enum and update all the places referencing it to make sure the datatype is properly supported when (en|de)coding terms and when (de)serialising them.
- If you want to add support of GeoSPARQL SPARQL functions you need to update the SPARQL parser spargebra crate. Then you will end up adding elements to a bunch of enums and the compiler should guide you to the places to update. I would maybe add geosparql support behind a `geosparql` feature in `spargebra` and `sparopt` to allow using them without GeoSPARQL.

Tpt:
Done in an other branch. There is now the `spargeo` crate.

nyurik:
thanks!!!

Tpt:
@nyurik don't set yor expectations too high, it's just a naive implementation of some geosparql functions. No index, nothing fast.

PR #747, opened 2024-01-18
update wasm-pack build

etiennept:
remove file transfers on wasm-pack build


js/build_package.js
```diff
@@ -1,15 +1,7 @@
 #! /usr/bin/env node
 
-const fs = require("fs");
+const fs= require("fs");
```
Tpt, line 3: CI is not happy because of the missing space here between `fs` and `=`

etiennept:
bug has been fixed

Tpt:
Thank you!

PR #750, opened 2024-01-19
Linting: Impl ordering, Self refs

nyurik:
For consistency, ordered implementation of a traits the same way as they are declared.

Used `Self::<assoc>` in a few last spots

### DIFF
--- a/lib/oxttl/src/line_formats.rs
+++ b/lib/oxttl/src/line_formats.rs
@@ -54,8 +54,8 @@ impl RuleRecognizer for NQuadsRecognizer {
     fn recognize_next(
         mut self,
         token: N3Token<'_>,
-        context: &mut Self::Context,
-        results: &mut Vec<Self::Output>,
+        context: &mut NQuadsRecognizerContext,
+        results: &mut Vec<Quad>,
         errors: &mut Vec<RuleRecognizerError>,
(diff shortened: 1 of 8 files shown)


lib/oxttl/src/line_formats.rs
```diff
@@ -54,8 +54,8 @@ impl RuleRecognizer for NQuadsRecognizer {
     fn recognize_next(
         mut self,
         token: N3Token<'_>,
-        context: &mut NQuadsRecognizerContext,
-        results: &mut Vec<Quad>,
+        context: &mut Self::Context,
+        results: &mut Vec<Self::Output>,
```
Tpt, line 58: This looks less readable to me than the concrete types because one would need to look at the associated type to know what the argument actually are. I am a bit not at ease with it. May you remove this part from your MR?

Tpt:
Thank you!

lib/oxttl/src/line_formats.rs
nyurik: yeah, i wasn't too certain about this one either - in some cases concrete types are better,  will revert

Tpt: Thank you!

Tpt (approved):

PR #765, opened 2024-02-07
Optimize `write!` performance with `write_str`

nyurik:
As seen in the https://rust.godbolt.org/z/Y8djWsq1P - `write!` macro produces significantly more code than a `write_str` call, so this change should have somewhat better performance.  To my knowledge, a lot of ppl tried to solve this optimization in the compiler, but no luck yet, so may help compiler ourselves for now.


Tpt (approved):
Thank you! I had in my TODO to manually expend these `write!` calls.

PR #766, opened 2024-02-07
Optimize some code, lints

nyurik:
* A few match simplifications
* Order trait impl to match trait itself
* A few unneeded parenthesis
* One dup code de-duplication with a new func

### DIFF
--- a/lib/oxrdf/src/dataset.rs
+++ b/lib/oxrdf/src/dataset.rs
@@ -925,8 +925,8 @@ impl PartialEq for Dataset {
 impl Eq for Dataset {}
 
 impl<'a> IntoIterator for &'a Dataset {
-    type IntoIter = Iter<'a>;
     type Item = QuadRef<'a>;
+    type IntoIter = Iter<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
         self.iter()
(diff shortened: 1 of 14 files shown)


lib/oxrdf/src/dataset.rs
```diff
@@ -925,8 +925,8 @@ impl PartialEq for Dataset {
 impl Eq for Dataset {}
 
 impl<'a> IntoIterator for &'a Dataset {
-    type IntoIter = Iter<'a>;
     type Item = QuadRef<'a>;
+    type IntoIter = Iter<'a>;
```
Tpt, line 929: Not sure about this, RustRover prefers to order `type` based on the decaration order whereas [rustfmt `reorder_impl_items` option](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#reorder_impl_items) prefers to use alphabetical order. I would tend to follow rustfmt because it is also usable by people not using RustRover but happy to be convinced otherwise. What do you think?

Tpt (approved):
Thank you! There is only a small code style topic where I would love your input.

lib/oxrdf/src/dataset.rs
nyurik: I agree that we shouldn't rely on IntelliJ, but it seems that `reorder_impl_items` is still not default, and most users have been asking for "same as in trait" if you look at the first comment in https://github.com/rust-lang/rustfmt/issues/3363#issuecomment-622580136

PR #773, opened 2024-02-08
use github action to install cargo tools

nyurik:
try to install tooling with an action, which has a fallback to binstall.  Should work faster than doing a `cargo install` with compilation


.github/workflows/tests.yml
```diff
@@ -126,7 +126,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-rust
```
Tpt, line 128: I think we can remove this line, no need for specific Rust setup

.github/workflows/tests.yml
```diff
@@ -398,7 +401,8 @@ jobs:
...
```
Tpt, line 403: same here

Tpt (approved):

Issue #774, opened 2024-02-08
Migrate errors to `thiserror`


nyurik:
In the years since https://github.com/oxigraph/oxigraph/issues/25 discussion, it seems [thiserror](https://docs.rs/thiserror/latest/thiserror/) has established itself as a defacto way to handle errors for libraries (but not always the binaries).  Moreover, it is additive - error enums can be migrated one at a time rather than all at once, and make code shorter, more readable, and avoid some accidental mistypes (like missing `impl Error`, or avoiding `From` for error casting, etc). Plus it is far better than any kind of boxed errors anyhow provides.  Note that thiserror is already a dependency in the lock file.

For example, converting `SerializerError` will remove the need to implement `fmt::Display`, `Error`, `From<io::Error>`, `From<StorageError>`.  <details><summary>Click this to see autogenerated code</summary>
<p>

```rust
    /// An error raised while writing a file from a [`Store`](crate::store::Store).
    #[automatically_derived]
    impl ::core::fmt::Debug for SerializerError {
        #[inline]
        fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
            match self {
                SerializerError::Io(__self_0) => {
                    ::core::fmt::Formatter::debug_tuple_field1_finish(f, "Io", &__self_0)
                }
                SerializerError::Storage(__self_0) => {
                    ::core::fmt::Formatter::debug_tuple_field1_finish(
                        f,
                        "Storage",
                        &__self_0,
                    )
                }
                SerializerError::DatasetFormatExpected(__self_0) => {
                    ::core::fmt::Formatter::debug_tuple_field1_finish(
                        f,
                        "DatasetFormatExpected",
                        &__self_0,
                    )
                }
            }
        }
    }

    #[allow(unused_qualifications)]
    impl std::error::Error for SerializerError {
        fn source(&self) -> ::core::option::Option<&(dyn std::error::Error + 'static)> {
            use thiserror::__private::AsDynError as _;
            #[allow(deprecated)]
            match self {
                SerializerError::Io { 0: transparent } => {
                    std::error::Error::source(transparent.as_dyn_error())
                }
                SerializerError::Storage { 0: transparent } => {
                    std::error::Error::source(transparent.as_dyn_error())
                }
                SerializerError::DatasetFormatExpected { .. } => {
                    ::core::option::Option::None
                }
            }
        }
    }

    #[allow(unused_qualifications)]
[...]

PR #775, opened 2024-02-08
Simplify all errors by using thiserror

nyurik:
This converts several errors to use `thiserror` crate, removing some code.   See #774 for an in-depth about `SerializerError` -- how the generated code differs.  Lots of boilerplate removed - over 550 lines!

Note that some errors use a nested `kind` approach.  I am not certain why this is done that way, and if it should be simplified?


lib/oxrdfxml/src/error.rs
```diff
@@ -64,78 +31,45 @@ impl From<quick_xml::Error> for ParseError {
...
     InvalidLanguageTag {
         tag: String,
+        #[source]
         error: LanguageTagParseError,
     },
-    Msg {
-        msg: String,
-    },
+    #[error("{msg}")]
+    Msg { msg: String },
```
Tpt, line 57: The variant list is imho an implementation detail, I would prefer keeping it private (and I am not sure anyone would want to `match` on it)

lib/oxsdatatypes/src/date_time.rs
```diff
@@ -2039,42 +2038,24 @@ fn time_on_timeline(props: &DateTimeSevenPropertyModel) -> Option<Decimal> {
...
 /// A parsing error
-#[derive(Debug, Clone)]
-pub struct ParseDateTimeError {
-    kind: ParseDateTimeErrorKind,
-}
-
-#[derive(Debug, Clone)]
-enum ParseDateTimeErrorKind {
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum ParseDateTimeError {
```
Tpt, line 2042: I would think that
- either we make the variant private
either we build more meaningful variants (the `Message` catch-all is quite hugly)

WDYT?

lib/oxsdatatypes/src/decimal.rs
```diff
@@ -609,67 +608,31 @@ impl fmt::Display for Decimal {
...
 /// An error when parsing a [`Decimal`].
-#[derive(Debug, Clone)]
-pub struct ParseDecimalError {
-    kind: DecimalParseErrorKind,
-}
-
-#[derive(Debug, Clone)]
-enum DecimalParseErrorKind {
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum ParseDecimalError {
```
Tpt, line 612: Same here, private variants is imho better

lib/sparesults/src/error.rs
```diff
@@ -73,28 +40,26 @@ impl From<quick_xml::Error> for ParseError {
...
-pub struct SyntaxError {
-    pub(crate) inner: SyntaxErrorKind,
-}
-
-#[derive(Debug)]
-pub(crate) enum SyntaxErrorKind {
-    Json(json_event_parser::SyntaxError),
-    Xml(quick_xml::Error),
+#[derive(Debug, thiserror::Error)]
+pub enum SyntaxError {
```
Tpt, line 50: I would prefer to make variants private here for encapsulation. They are mostly implementation details and not very meaningful for the user

lib/spargebra/src/parser.rs
```diff
@@ -9,64 +9,34 @@ use oxrdf::vocab::{rdf, xsd};
...
-            ParseErrorKind::InvalidBaseIri(e) => Some(e),
-            ParseErrorKind::Parser(e) => Some(e),
-        }
-    }
+#[derive(Debug, thiserror::Error)]
+pub enum ParseError {
+    #[error("Invalid SPARQL base IRI provided: {0}")]
+    InvalidBaseIri(#[from] IriParseError),
+    #[error(transparent)]
+    Parser(#[from] peg::error::ParseError<LineCol>),
```
Tpt, line 39: I would prefer to expose we use peg here for parsing. Maybe make variants private

Side note: I should maybe rename this type `SyntaxError` for consistency with the other crates

Tpt (commented):
Thank you very much! This very welcome

However, I am a bit reluctant of exposing the variants and internal details of all our error types, I bothered to make them private to give me some change leeway and would love to keep it.

nyurik:
Thx for the feedback!  The two main reasons for this PR (which I probably should have discussed with you first) is 1) to make the code faster, and 2) to make code smaller and more readable, so it can be easier to modify and to spot issues.

The smaller code part - removing ~650 lines of code, is, I think, self evident :)  Obviously it shouldn't compromise on usability (e.g. encapsulation).  The performance part might be a bigger aspect - I was a bit surprised by numerous boxed types you used in the errors. Granted, errors shouldn't happen too often, but in general I think it is better to have specific type everywhere - it keeps code simpler and faster.

WRT encapsulation - is there much value in doing it?  If a user of the library needs to dig deep into the error to match on it, they will also match on the string representation - as they are clearly interested in that specific error and want to handle it in some special way. Except that if they do it on a string, compilation will not break when the error string has changed. And its not like you guarantee the error messages to be stable. So as a result, a string representation and an internal error structure (in my mind) are OK to break somewhat(?).

Secondly, I do not believe encapsulation should be done on the level of the module itself.  E.g. a parsing crate API should return original enum errors.  This allows re-usability - any crate can be used by someone else externally.  On the other hand, if you want parsing crate to be used inside the storage crate, and want to encapsulate that error, you may want to encapsulate parsing error in the storage error. Not sure yet of the exact code, but if you really want to encapsulate, perhaps we can make a macro to standardize that process?

Tpt:
> Thx for the feedback! The two main reasons for this PR (which I probably should have discussed with you first) is 1) to make the code faster, and 2) to make code smaller and more readable, so it can be easier to modify and to spot issues.

Yes! And this is very welcome. Thank you!

>  The performance part might be a bigger aspect - I was a bit surprised by numerous boxed types you used in the errors. Granted, errors shouldn't happen too often, but in general I think it is better to have specific type everywhere - it keeps code simpler and faster.

Something to take into account is that error types `E` that takes a lot of bytes on the stack might lead to bigger `Result<_, E>` types and so, slow the program. It's why some libraries always box errors to make sure to reduce the size of the function results. But all of this sounds like premature optimizations if not driven by benchmarks.

>  If a user of the library needs to dig deep into the error to match on it, they will also match on the string representation

Indeed. My implicit assumption is that the types are stable but error messages are not. If messages are broken, too bad for the user who assumes them to be stable.

My main motivation to encapsulate some errors is to allow things like "add support for a new RDF serialization" or "improve the internal representation of the error" without breaking backward compatibility. I try to have public enums where I tend to think the user might want to match on error causes and private when it looks not much useful and I would be happy to get ability to edit/extend the internal structure.

> Secondly, I do not believe encapsulation should be done on the level of the module itself. E.g. a parsing crate API should return original enum errors. This allows re-usability - any crate can be used by someone else externally. On the other hand, if you want parsing crate to be used inside the storage crate, and want to encapsulate that error, you may want to encapsulate parsing error in the storage error. 

Sorry, I am not 100%  sure to follow. For example, you mean that the general parsing crate erroy type should publicly expose the errors of the parsers used for parsing?

> perhaps we can make a macro to standardize that process?

I believe thiserror supports this use case. See the <q>Another use case is hiding implementation details of an error representation behind an opaque error type</q> part of the [documentation](https://docs.rs/thiserror/latest/thiserror/).

Thank you again!

adamreichold:
> Something to take into account is that error types E that takes a lot of bytes on the stack might lead to bigger Result<_, E> types and so, slow the program. It's why some libraries always box errors to make sure to reduce the size of the function results. But all of this sounds like premature optimizations if not driven by benchmarks.

I would like to second this. The reason `anyhow` works so hard to make `Result<()>` a single word is that that the costs of large `Result` instantiation become an issue quite quickly but are hard to pinpoint due to being "distributed fat" that will often not be very visible in a profiler.

lib/spargebra/src/parser.rs
nyurik: I think it is actually very confusing to have so many identically named errors. I will add a prefix to keep them different

nyurik:
The `oxrdfxml::error::SyntaxErrorKind` is actually a public enum -- I think at least this one should be made into public enum without a wrapper, right?   I made some renaming in #778 to make this migration easier.

nyurik:
In general, I think one of the other complexity with errors relates to how many `From` traits they implement - feels like every error type converts to every other error type - making it less ... fluent i guess.  Hence the desire to reduce the number of various error types.

Thanks a lot on explaining the error boxing aspect - I was not aware of that, will keep it in mind

lib/oxigraph/src/storage/error.rs
```diff
@@ -56,59 +32,42 @@ impl From<StorageError> for io::Error {
...
 
-    /// Builds an error from a printable error message.
     #[inline]
-    pub(crate) fn msg(msg: impl Into<String>) -> Self {
-        Self {
-            inner: CorruptionErrorKind::Msg(msg.into()),
-        }
+    pub(crate) fn from_encoded_term(encoded: &EncodedTerm, term: &TermRef<'_>) -> Self {
+        // TODO: eventually use a dedicated error enum value
+        Self::new(format!("Invalid term encoding {encoded:?} for {term}"))
```
Tpt, line 58: Nit: Is is building a `CorruptionErrorKind::Other(Box<dyn String>)`. What about using `Self::msg` to get a `CorruptionErrorKind::Msg(String)` instead?

lib/oxrdfxml/src/error.rs
```diff
@@ -72,61 +39,39 @@ impl From<quick_xml::Error> for ParseError {
 }
 
 /// An error in the syntax of the parsed file.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
 pub struct SyntaxError {
+    #[from]
     pub(crate) inner: SyntaxErrorKind,
```
Tpt, line 46: Genuine question: Why not `SyntaxError(#[from] SyntaxErrorKind)` like in the other places?

Tpt (commented):
Thank you! I allowed myself two last nits.

> The oxrdfxml::error::SyntaxErrorKind is actually a public enum -- I think at least this one should be made into public enum without a wrapper, right? I made some renaming in #778 to make this migration easier.

The `error` module in `oxrdfxml` is private so it is actually `pub(crate)`.

> In general, I think one of the other complexity with errors relates to how many `From` traits they implement - feels like every error type converts to every other error type - making it less ... fluent i guess. Hence the desire to reduce the number of various error types.

Yes! I arises from the Oxigraph architecture with specialized libraries that are then composed together. Each one has its own error types. I am not sure how to make it significantly better without leaking information from one library to the others. Having shared error types would add yet an other crate and I am a bit afraid of getting strait into the "god class" anti-pattern.

lib/oxrdfxml/src/error.rs
nyurik: thx, missed it, fixed

lib/oxigraph/src/storage/error.rs
nyurik: Sure - I was trying to keep original error codes - you used this before:

```
CorruptionError::new(format!("Invalid term encoding {encoded:?} for {term}")).into())
```

fixed

nyurik:
yeah, "everything and a kitchen sink" is not that great either...  although it does happen that way in effect because parent result inherits all other results - except that some are encapsulated with private content... not sure what the good solution to streamline it is tbh... will see

Tpt (approved):
Thank you!

lib/oxigraph/src/storage/error.rs
Tpt: It seems I got it wrong. sorry for that

Tpt:
I forgot to squash commits before merging, I allowed myself to rewrite the main branch history. Sorry for that

nyurik:
np. I usually configure in repo settings to only allow squash merges, and use pull request title and descriptions.  Seem to work ok

PR #779, opened 2024-02-10
Use io::Error::other

nyurik:
Requires 1.74


Tpt:
Very welcome. I would maybe wait a bit before upgrading MSRV to 1.74. For example Debian [is still in 1.70](https://tracker.debian.org/pkg/rustc).

nyurik:
heh, debian is slow... besides, no rush to merge this one, hopefully won't go too stale, and it is easy to update, so lets table this one for now

Tpt:
Done in an other MR

PR #781, opened 2024-02-10
Normalize unicode refs

nyurik:
Just to keep them a bit more consistent.

Note that there are a lot of code duplications here - but I do not know if they are worth consolidating, and/or any perf implications.


Tpt:
thank you!

Tpt (approved):

PR #783, opened 2024-02-10
Adds an async SPARQL JSON results reader

Tpt:
Same design as the other parsers: a state machine fed by events (here JSON events)

### DIFF
--- a/lib/sparesults/src/json.rs
+++ b/lib/sparesults/src/json.rs
@@ -1,20 +1,16 @@
 //! Implementation of [SPARQL Query Results JSON Format](https://www.w3.org/TR/sparql11-results-json/)
 
 use crate::error::{QueryResultsParseError, QueryResultsSyntaxError};
-#[cfg(feature = "async-tokio")]
-use json_event_parser::ToTokioAsyncWriteJsonWriter;
 use json_event_parser::{FromReadJsonReader, JsonEvent, ToWriteJsonWriter};
+#[cfg(feature = "async-tokio")]
+use json_event_parser::{FromTokioAsyncReadJsonReader, ToTokioAsyncWriteJsonWriter};
 use oxrdf::vocab::rdf;
 use oxrdf::*;
 use std::collections::BTreeMap;
 use std::io::{self, Read, Write};
 use std::mem::take;
 #[cfg(feature = "async-tokio")]
-use tokio::io::AsyncWrite;
-
-/// This limit is set in order to avoid stack overflow error when parsing nested triples due to too many recursive calls.
-/// The actual limit value is a wet finger compromise between not failing to parse valid files and avoiding to trigger stack overflow errors.
-const MAX_NUMBER_OF_NESTED_TRIPLES: usize = 128;
+use tokio::io::{AsyncRead, AsyncWrite};
 
 pub fn write_boolean_json_result<W: Write>(write: W, value: bool) -> io::Result<W> {
     let mut writer = ToWriteJsonWriter::new(write);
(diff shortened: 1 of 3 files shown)


Tpt:
@nyurik If you have some time review on this MR would be very welcome. The code is quite hugly but I am not sure there is a better way to make the code work in both sync and async cases.

lib/sparesults/src/json.rs
```diff
@@ -224,520 +220,882 @@ fn write_json_term<'a>(output: &mut Vec<JsonEvent<'a>>, term: TermRef<'a>) {
...
+                    Ok(var) => {
+                        if self.variables.contains(&var) {
+                            return Err(QueryResultsSyntaxError::msg(format!(
+                                "The variable {var} is declared twice"
+                            )));
+                        }
+                        self.variables.push(var);
+                        Ok(None)
                     }
+                    Err(e) => Err(QueryResultsSyntaxError::msg(format!(
```
nyurik, line 504: Usually it is advised not to use formatting to create an error message because later that message might actually be dropped without being used. Instead, there should be an error enum variant with all the needed values - so that an error message can be constructed at the last moment if needed.

lib/sparesults/src/json.rs
```diff
@@ -224,520 +220,882 @@ fn write_json_term<'a>(output: &mut Vec<JsonEvent<'a>>, term: TermRef<'a>) {
...
                 }
+                _ => Err(QueryResultsSyntaxError::msg(
+                    "Expecting a new solution object",
+                )),
+            },
+            JsonInnerSolutionsReaderState::BetweenSolutionTerms => match event {
+                JsonEvent::ObjectKey(key) => {
+                    let key = *self.mapping.get(key.as_ref()).ok_or_else(|| {
+                        QueryResultsSyntaxError::msg(format!(
+                            "The variable {key} has not been defined in the header"
```
nyurik, line 744: same

lib/sparesults/src/json.rs
```diff
@@ -224,520 +220,882 @@ fn write_json_term<'a>(output: &mut Vec<JsonEvent<'a>>, term: TermRef<'a>) {
...
+                        }
+                    };
+                    Ok(None)
+                }
+                JsonEvent::EndObject => {
+                    self.state = JsonInnerTermReaderState::Middle;
+                    Ok(None)
                 }
-                _ => ignore_value(reader)?,
+                _ => unreachable!(),
```
nyurik, line 1050: here and elsewhere i noticed that you don't have an exhaustive list of enums in a match -- there is even a [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard_enum_match_arm) for that. The issue is that if you add a new enum variant, you should be explicit that you know about that new value and handled it, instead of silently ignored until it crashes in the runtime.

lib/sparesults/src/json.rs
```diff
@@ -224,520 +220,882 @@ fn write_json_term<'a>(output: &mut Vec<JsonEvent<'a>>, term: TermRef<'a>) {
...
+                    return Ok(Some(result));
+                }
+            },
+            JsonInnerSolutions::Iterator(iter) => iter.next(),
+        }
+    }
+}
 
+#[cfg(feature = "async-tokio")]
+pub enum FromTokioAsyncReadJsonQueryResultsReader<R: AsyncRead + Unpin> {
```
nyurik, line 278: ouch, the size of this struct name is nearly unreadable :(  Any way to make it shorter?

nyurik (commented):
looks really good!  I was not aware SPARQL had json support, will need to read up on that.

The code looks good, but there are a few nits I think you may want to address - such as extensive `format!()` when creating errors - even though errors shouldn't occur too often, when they do, you might be wasting cycles if their output is not shown.  Also, the non-exhaustive enums match statements may hide some bugs (see inline comment).

lib/sparesults/src/json.rs
Tpt: Indeed. Thank you! But I am not sure it's very significant, we already allocate quite a lot when building the parser and returning the results (each `Term` wraps at least an allocation) so an extra allocation on error is probably not a big deal. I would tend to think there are a lot of things to optimize in this parser and elsewhere is Oxigraph.

lib/sparesults/src/json.rs
Tpt: Good point! I only did that when inside of objects where the only events are `ObjectKey` and `EndObject`. If not, there is a big bug I would love to get a bug report about it. I also fuzz this parser and the JSON parser to ensure there is no crash.

lib/sparesults/src/json.rs
Tpt: Yes, that's awful. `TokioAsyncRead` is more or less mandatory because we want to state this is the wrapper for an `AsyncRead` from Tokio (and not the possible future ones for `futures::io::AsyncRead` and the possible `std::io::AsyncRead`...), `Json` to distinguish with `Xml` and `Reader` to state what the struct does.

I can maybe drop `QueryResults` because this is an internal type but I won't skip the `QueryResults` part in the public API to avoid name clash with the RDF parsers so it's not a great idea.

What I can easily do is drop the `From` (and `To` in the serializers). What do you think?

Tpt:
Thank you for the review!

lib/sparesults/src/json.rs
nyurik: Its probably not in this case...  To me, this is more of a "good practice to always do" -- this way you never even have to think if this is a perf bottleneck or not, and keeps the code more compact, and allows to track all possible error conditions:

```rust
Err(QueryResultsSyntaxError::msg(
                    "SPARQL JSON results must contain a 'boolean' or a 'results' key",
                )),

// vs

Err(QueryResultsSyntaxError::JsonMissingBoolOrResultsKey),
```

lib/sparesults/src/json.rs
nyurik: "if we rely on fuzzing to catch crashes in Rust, we are doing something wrong"  :rofl:  (not really, but if something can be done by the compiler, might as well? well... usually :)

lib/sparesults/src/json.rs
nyurik: Naming is hard!  (tm)  No idea tbh

lib/sparesults/src/json.rs
Tpt: Yes! Agreed! Just got lazy with this MR. It's a pervasise thing in Oxigraph codebase. I have opened #790 about it.

lib/sparesults/src/json.rs
Tpt: > not really, but if something can be done by the compiler, might as well? well... usually

Yes, I should have found a better way to type the events generated by the JSON parser. This is a workaround of bad typing like often.

nyurik: can't blame you, me same way :)

nyurik: Oops, i meant this about the error enum variants

PR #791, opened 2024-02-15
RocksDB: uses multi-columns flush

Tpt:


oxrocksdb-sys/api/c.cc
```diff
@@ -272,11 +272,15 @@ void rocksdb_transactiondb_put_cf_with_status(
...
 }
 
-void rocksdb_transactiondb_flush_cf_with_status(
+void rocksdb_transactiondb_flush_cfs_with_status(
     rocksdb_transactiondb_t* db, const rocksdb_flushoptions_t* options,
-    rocksdb_column_family_handle_t* column_family,
+    rocksdb_column_family_handle_t** column_families, int num_column_families,
     rocksdb_status_t* statusptr) {
-  SaveStatus(statusptr, db->rep->Flush(options->rep, column_family->rep));
+  std::vector<ColumnFamilyHandle*> column_family_handles(num_column_families);
```
adamreichold, line 279: IIRC, this will create a vector with `num_column_families` default-inserted elements, i.e. the calls to `push_back` will mean that the final vector contains `2 * num_column_families` elements.

I think you want to do either

```cpp
column_family_handles[i] = column_families[i]->rep;
```

or just default-construct the vector and call `reserve`.

Tpt: Indeed. Thank you so much! Fixed.

PR #795, opened 2024-02-20
feat: Support casting of `Term`

jeswr:
This is a proposal to cast `Term`s back to particular types of terms.

I was mainly after the feature of being able to cast `Term`s as `Optional<Subject>`s; and being able to construct an `Optional<Triple>` from arbitrary `Term`s.

If you're on board with this feature I can add documentation and then mark this as ready for review.

### DIFF
--- a/lib/oxrdf/src/blank_node.rs
+++ b/lib/oxrdf/src/blank_node.rs
@@ -114,7 +114,7 @@ impl TryFrom<Term> for BlankNode {
             Ok(node)
         } else {
             Err(TermCastErrorKind::Msg(format!(
-                "Cannot convert a term to a blank node: {}",
+                "Cannot convert term to a blank node: {}",
                 term
             ))
             .into())
--- a/lib/oxrdf/src/blank_node.rs
+++ b/lib/oxrdf/src/blank_node.rs
@@ -103,22 +103,6 @@ impl fmt::Display for BlankNode {
     }
 }
 
-impl TryFrom<Term> for BlankNode {
-    type Error = TermCastError;
-
-    #[inline]
-    fn try_from(term: Term) -> Result<Self, Self::Error> {
-        if let Term::BlankNode(node) = term {
-            Ok(node)
-        } else {
-            Err(
-                TermCastErrorKind::Msg(format!("Cannot convert term to a blank node: {}", term))
-                    .into(),
-            )
-        }
-    }
-}
-
 impl Default for BlankNode {
     /// Builds a new RDF [blank node](https://www.w3.org/TR/rdf11-concepts/#dfn-blank-node) with a unique id.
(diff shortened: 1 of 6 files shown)


lib/oxrdf/src/blank_node.rs
```diff
@@ -102,6 +104,16 @@ impl fmt::Display for BlankNode {
     }
 }
 
+impl From<Term> for Option<BlankNode> {
+    #[inline]
+    fn from(term: Term) -> Self {
+        match term {
+            Term::BlankNode(node) => Some(node),
+            _ => None,
+        }
```
Tpt, line 113: If you like slightly shorter syntax:
```rust
if let Term::BlankNode(node) = term {
   Some(node)
} else {
   None
}
```

lib/oxrdf/src/triple.rs
```diff
@@ -738,6 +751,26 @@ impl Triple {
...
+            Into::<Option<NamedNode>>::into(Into::<Term>::into(predicate)), 
+            Into::<Option<Term>>::into(Into::<Term>::into(object))) {
+            (Some(subject), Some(predicate), Some(object)) => Some(Self {
+                subject,
+                predicate,
+                object,
+            }),
+            _ => None,
+        }
+    }
```
Tpt, line 772: I see the point of this function but I would make it return a `Result<Self, TermCastError>` to highlight that it returns an error with `TermCastError` probably the same error as the one in the new `TryFrom` implementations.

Tpt (commented):
Thank you! 

I allowed myself some design suggestions. The conversion functions might indeed be handy but I am reluctant to merge the current `Option` based design (returning `Result` might be better because we are dealing with errors and not with maybe present data)

jeswr:
Awesome - I'll address the suggestions in a few hrs

lib/oxrdf/src/cast_error.rs
jeswr, line 1: I'm not sure what the most idiomatic way of defining new errors is in rust - so I've just re-used the way `CorruptionError` was defined.

jeswr:
The failing test should be fixed now @Tpt

jeswr:
@Tpt I'm on board with all of these ideas - if you have time today feel free to make the changes so that I don't slow things down in getting this merged; otherwise I can take a look at things tomorrow.

Tpt:
> I've applied those changes @Tpt - feel free to make changes directly to this PR if you have further thoughts/suggestions so I don't slow things down further.

Thank you! Will do so (I am quite buzy at the moment so it might not be today sadly)

Tpt:
Hey @jeswr I am not allowed to edit your branch so I open #807 that is based on yours (the first commit is the squash of yours and the second is some simplifications)

PR #796, opened 2024-02-20
chore: make the LiteralContent enum public

jeswr:
Makes the `LiteralContent` enum public to enable pattern matching.

### DIFF
--- a/lib/oxrdf/src/literal.rs
+++ b/lib/oxrdf/src/literal.rs
@@ -1,6 +1,6 @@
 use crate::named_node::NamedNode;
 use crate::vocab::{rdf, xsd};
-use crate::NamedNodeRef;
+use crate::{NamedNodeRef, Term};
 use oxilangtag::{LanguageTag, LanguageTagParseError};
 #[cfg(feature = "oxsdatatypes")]
 use oxsdatatypes::*;
@@ -34,10 +34,10 @@ use std::option::Option;
 /// # Result::<(), LanguageTagParseError>::Ok(())
 /// ```
 #[derive(Eq, PartialEq, Debug, Clone, Hash)]
-pub struct Literal(LiteralContent);
+pub struct Literal(pub LiteralContent);
 
 #[derive(PartialEq, Eq, Debug, Clone, Hash)]
-pub enum LiteralContent {
+enum LiteralContent {
     String(String),
     LanguageTaggedString { value: String, language: String },
     TypedLiteral { value: String, datatype: NamedNode },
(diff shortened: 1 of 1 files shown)


lib/oxrdf/src/literal.rs
```diff
@@ -37,7 +37,7 @@ use std::option::Option;
 pub struct Literal(LiteralContent);
```
Tpt, line 37: You might want to make `LiteralContent` `pub` here to make it accessible.

Tpt (commented):
Thank you for this pull request!

I am not sure it's a good idea. I would prefer to keep `LiteralContent` private to allow changing the way literals are encoded internally without breaking the public API. There is already [the `destruct` method](https://docs.rs/oxrdf/latest/oxrdf/struct.Literal.html#method.destruct) to get back the literal components and various getters. Would it be enough to cover your use case?

jeswr:
> Would it be enough to cover your use case?

It does indeed - though a minor annoyance with that method is that the type is a little bit looser i.e. one needs to handle cases like `(value, Some(...), Some(...))` which never occur in practice.

lib/oxrdf/src/literal.rs
jeswr: I'm not quite sure what the effect of doing this is, I was assuming it would allow

```rust
pub fn from_term(object: Term) -> Option<String> {
    match object {
        Term::NamedNode(iri) => None,
        Term::BlankNode(id) => None,
        Term::Literal(Literal(LiteralContent::LanguageTaggedString { value, language })) => None,
        Term::Literal(Literal(LiteralContent::String(value))) => Some(value),
        Term::Literal(Literal(LiteralContent::TypedLiteral { value, datatype })) => Some(value),
        #[cfg(feature = "rdf-star")]
        Term::Triple(_) => None,
    }
}
```

but this code still results in the error

```
note: enum `crate::literal::LiteralContent` exists but is inaccessible
label: use of undeclared type `LiteralContent`
```

Tpt: You need to add a `pub use crate::literal::Literal` in the `lib.rs` file.

Tpt:
> It does indeed - though a minor annoyance with that method is that the type is a little bit looser i.e. one needs to handle cases like `(value, Some(...), Some(...))` which never occur in practice.

Indeed, that's true. However I have to say I never used the `destruct` method myself and always manage to just use the `value`, `language` and `datatype` accessors.

Tpt:
To make it work you need to do 3 things:
- make `LiteralContent` `pub` (what your patch was doing)
- make the inner field of `Literal` `pub` (what your patch is doing now)
- because the `literal` module is not public, reexport `Literal` content in the root of the library by adding `pub use crate::literal::LiteralContent;` in the `lib.rs` file.

Anyway, I am not 100% sure it's a good change to make because it leaks the `Literal` implementation internals.

jeswr:
> To make it work you need to do 3 things:

Makes sense now - thankyou for explaining!

> Anyway, I am not 100% sure it's a good change to make because it leaks the Literal implementation internals.

Gotcha - closing the PR now

PR #801, opened 2024-02-25
CI: Read MSRV from Cargo.toml

Tpt:
Extracted from #776

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -173,9 +173,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: true
+      - name: Read MSRV from Cargo.toml
+        id: metadata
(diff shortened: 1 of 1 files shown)


Tpt:
@nyurik Is it fine with you if I reuse a part of #776 here?

nyurik:
Sure, but I layer found a much better way with cargo exporting Metadata as json , and using jq to get the right value. I'm afk, but can look it up later.  But up to you, sed also works

Tpt:
> Sure, but I layer found a much better way with cargo exporting Metadata as json , and using jq to get the right value. I'm afk, but can look it up later. But up to you, sed also works

Nice! Thank you! However, if I remind correctly `cargo metadata` already does quite a lot of processing so `sed` might be slightly faster and simpler. And if we break it for some reason CI will scream

PR #837, opened 2024-03-26
Upgrades to PyO3 0.21

Tpt:


python/src/model.rs
```diff
@@ -1115,18 +1116,16 @@ impl<'a> From<&'a PyNamedOrBlankNodeRef<'a>> for NamedOrBlankNodeRef<'a> {
...
 }
 
-impl<'a> TryFrom<&'a PyAny> for PyNamedOrBlankNodeRef<'a> {
-    type Error = PyErr;
-
-    fn try_from(value: &'a PyAny) -> PyResult<Self> {
-        if let Ok(node) = value.extract::<PyRef<'_, PyNamedNode>>() {
+impl<'py> FromPyObject<'py> for PyNamedOrBlankNodeRef<'py> {
+    fn extract_bound(ob: &Bound<'py, PyAny>) -> PyResult<Self> {
+        if let Ok(node) = ob.extract::<PyRef<'_, PyNamedNode>>() {
```
adamreichold, line 1121: Are the type ascriptions necessary here?

python/src/dataset.rs
```diff
@@ -50,15 +50,16 @@ impl PyDataset {
     /// >>> store = Dataset([Quad(NamedNode('http://example.com'), NamedNode('http://example.com/p'), Literal('1'), NamedNode('http://example.com/g'))])
     /// >>> list(store.quads_for_subject(NamedNode('http://example.com')))
     /// [<Quad subject=<NamedNode value=http://example.com> predicate=<NamedNode value=http://example.com/p> object=<Literal value=1 datatype=<NamedNode value=http://www.w3.org/2001/XMLSchema#string>> graph_name=<NamedNode value=http://example.com/g>>]
-    pub fn quads_for_subject(&self, subject: &PyAny) -> PyResult<QuadIter> {
-        Ok(QuadIter {
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn quads_for_subject(&self, subject: PySubjectRef<'_>) -> QuadIter {
```
adamreichold, line 54: If you do change the signature, why not take a `&PySubjectRef<'_>` to make Clippy happy? Do you want to avoid the double indirection?

python/src/model.rs
Tpt: Not really but it allows to avoid [a false positive of `clippy::same-functions-in-if-condition`](https://github.com/rust-lang/rust-clippy/issues/10928)

python/src/dataset.rs
Tpt: It's because I don't think there is a proper way to implement `FromPyObject` on `&PySubjectRef<'_>` because a place to own the ownership of the Python object is required.

I see two ways:
* Either make `PySubjectRef` a `PySubject` that owns the Python object instead of borrowing it.
* Implement `PyFunctionArgument` instead of `FromPyObject` but it's part of PyO3 internal API.

I am not very happy with either of the approaches.

adamreichold: Indeed, this will have to wait until `FromPyObject` gains another lifetime. Well, maybe you could take us up on

> Please contact the PyO3 maintainers if you believe you have a use case for implementing this trait before PyO3 is ready to change the main FromPyObject trait to take an additional lifetime.

from the docs of [`FromPyObjectBound`](https://docs.rs/pyo3/latest/pyo3/conversion/trait.FromPyObjectBound.html) but I am not sure the Clippy lint alone is sufficient motivation. :)

Tpt: > I am not sure the Clippy lint alone is sufficient motivation. :)

Yes, I think there are way higher priorities for PyO3. For example I think moving forward on PyO3/pyo3#3977 (automated stub generation) would be way more useful

PR #849, opened 2024-04-16
DESCRIBE: returns the Concise Bounded Description instead of just the triples describing the subject

Tpt:
https://www.w3.org/submissions/CBD/


Tpt:

PR #851, opened 2024-04-16
JS: Allows to return query results in JSON/XML/CSV/TSV

Tpt:


Tpt:

PR #853, opened 2024-04-18
JS: Properly serializes CONSTRUCT results using RDF formats rather than SPARQL results formats

Tpt:
Adds more tests and documentation


Tpt:

PR #865, opened 2024-04-25
Rewrites the in-memory storage

Tpt:
- more correct: uses MVCC with single writer to avoid dirty reads and bad transaction rollbacks
- faster: allows any number of readers during write
- no copy on iteration

All quads are stored in a hash set and pointers are used to "jump" between the quads with the same subject/predicate/object/graph name. Indexes are also used to get the start of the by subject/predicate/... linked lists


lib/oxigraph/src/storage/memory.rs
```diff
@@ -5,75 +5,239 @@ use crate::storage::binary_encoder::{
...
+
+/// In-memory storage working with MVCC
+///
+/// Each quad and graph name is annotated by a version range, allowing to read old versions while updates are applied.
+/// To simplify the implementation a single write transaction is currently allowed. This restriction should be lifted in the future.
 #[derive(Clone)]
 pub struct MemoryStorage {
     content: Arc<RwLock<Content>>,
-    id2str: Arc<RwLock<HashMap<StrHash, String>>>,
+    id2str: Arc<DashMap<StrHash, String, BuildHasherDefault<StrHashHasher>>>,
```
adamreichold, line 34: Not necessarily better or worse, but the `concread` crate might a good fit for the desired MVCC semanitcs here. Not sure if you really can rely on their versioning though as need to keep multiple maps in sync.

Tpt: Thank you! I had a look at it. Sadly, it is under MPLv2 and Oxigraph is Apache 2/MIT. I did not wanted to introduce MPLv2 dependencies.

lib/oxigraph/src/storage/memory.rs
```diff
@@ -5,75 +5,239 @@ use crate::storage::binary_encoder::{
...
 
-/// Low level storage primitives
+pub type MemoryChainedQuadIterator =
+    Chain<std::vec::IntoIter<EncodedQuad>, std::vec::IntoIter<EncodedQuad>>;
+pub type MemoryDecodingGraphIterator = std::vec::IntoIter<EncodedTerm>;
+
+/// In-memory storage working with MVCC
+///
+/// Each quad and graph name is annotated by a version range, allowing to read old versions while updates are applied.
+/// To simplify the implementation a single write transaction is currently allowed. This restriction should be lifted in the future.
```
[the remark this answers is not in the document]
Tpt: Yes! Definitely. It's why I made the choice of not bothering with concurrent writes for the moment. I took the MVCC approach instead of "copy on write" to make it easier to reuse existing datastructures. The next step seems to me is to fix the `iter` method (and some other ones) that currently buffers all the quads into a `Vec` before returning them.

Tpt:
CC @ludovicm67 @BenjaminHofstetter Big change of the in-memory storage to make it more correct (avoid dirtly reads and bad transaction rollbacks). Should be a good foundation to improve performances (and already speeds up bulk loading by 50%). The "JS: " benchmarks seems pretty unreliable but the Rust benchmarks are way more stable.

Tpt:
Sadly memory consumption is still wost than the RocksDB memory storage. It takes around 500MB per milion triples. However, it is already more than 2x better than the previous memory storage used before in WASM builds.

Issue #868, opened 2024-04-29
Build on OpenBSD 7.5 amd64


donpellegrino:
It would be useful if Oxigraph could build and run on OpenBSD.

PR #869, opened 2024-04-29
10 exploratory build on openbsd 75

donpellegrino:
Fix for #868.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,8 @@ available by default limits in OpenBSD.
 
 ## Help
 
-- Feel free to use [GitHub discussions](https://github.com/oxigraph/oxigraph/discussions) or [the Gitter chat](https://gitter.im/oxigraph/community) to ask questions or talk about Oxigraph.
-- [Bug reports](https://github.com/oxigraph/oxigraph/issues) are also very
-welcome.
+Feel free to use [GitHub discussions](https://github.com/oxigraph/oxigraph/discussions) or [the Gitter chat](https://gitter.im/oxigraph/community) to ask questions or talk about Oxigraph.
+[Bug reports](https://github.com/oxigraph/oxigraph/issues) are also very welcome.
 
 If you need advanced support or are willing to pay to get some extra features, feel free to reach out to [Tpt](https://github.com/Tpt/).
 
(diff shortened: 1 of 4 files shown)


lib/oxigraph/Cargo.toml
```diff
@@ -31,6 +31,7 @@ digest.workspace = true
 hex.workspace = true
 json-event-parser.workspace = true
 md-5.workspace = true
+openssl-sys = "0.9.102"
```
Tpt, line 34: This will try compile openssl-sys for all plateforms, includings ones where we use the host native SSL implementation if the `http-client-native-tls` feature is enabled (macOS, Windows) and WASM where openssl can't be compiled. I would tend to think it would be better to dynamically link to libressl on OpenBSD, it [seems to work](https://github.com/sfackler/rust-openssl/issues/2050) or, if not, only add this dependency if the target is openbsd and the `http-client-native-tls` feature is enabled.

README.md
```diff
@@ -53,10 +53,23 @@ The library layers in Oxigraph. The elements above depend on the elements below:
...
 
+## Building
+
 When cloning this codebase, don't forget to clone the submodules using
 `git clone --recursive https://github.com/oxigraph/oxigraph.git` to clone the repository including submodules or
 `git submodule update --init` to add the submodules to the already cloned repository.
 
+When building on OpenBSD, libclang must be available for
+RocksDB. Setting the LIBCLANG_PATH environment variable will ensure it
+is found. For example:
```
Tpt, line 64: We can maybe just link from there to [bindgen installation requirement](https://rust-lang.github.io/rust-bindgen/requirements.html) that covers this topic in more details. What do you think?

Tpt (commented):
Thank you so much for working on it. This is very welcome!

lib/oxigraph/Cargo.toml
donpellegrino: That makes perfect sense. I will investigate the dynamic linking approach on OpenBSD and try that instead.

I have to research this a bit further, but it seemed that an older version of openssl-sys was being used in the build. I have to track down how that happened without it being listed in Cargo.toml or Cargo.lock files. I could have misinterpreted the build errors at the time.

Tpt: Thank you! This is weird, I see `openssl-sys` in [the lock file](https://github.com/oxigraph/oxigraph/blob/196526f01e3aa12999a9160255b94200cdae22b0/Cargo.lock#L1010). It is a dependency of `native-tls` that is itself a dependency of `oxhttp` if the `native-tls` feature is enabled (this feature is enabled by the `http-client-native-tls` feature of `oxigraph`)

PR #873, opened 2024-05-17
Moves back some structs to 8 bytes aligment to save memory

Tpt:


Tpt:
Small memory improvement. Decreases memory consumption from 400MB to 330MB per million triple on x86-64

PR #877, opened 2024-05-18
JS: Refactors Store.load and Store.dump API to enable more options

Tpt:


Tpt:
CC @ludovicm67 @BenjaminHofstetter The `no_transaction` and `unchecked` both enabled seems to provide a ~20% speedup when loading data (actual speedup is quite dependant on the actual RDF data)

Issue #847, opened 2024-04-12
Serve yasgui by oxigraph


MarcAntoine-Arnaud:
**Is your feature request related to a problem? Please describe.**
Several times I cannot access to the oxigraph ui because yasgui.min.css and yasgui.min.js are not available.

**Describe the solution you'd like**
Do not use unpkg server, but store these files in the server and serve itself these ressources.

Tpt:
Yes! This would be very handy definitely

ktk:
Reminds me that I recommend to switch to our fork, in case this didn't happen yet. The original repo is basically unmaintained (I did talk to Triply about this). We fixed several security issues in our version. https://github.com/zazuko/Yasgui

Tpt:
@ktk Thanks, done in #852

@MarcAntoine-Arnaud I am investigating a way to embed the yasgui.min.css and yasgui.min.js files in the built Oxigraph binary without making npm a required dependency to build Oxigraph. Sadly the `build` directory is not commited in zazuko/Yasgui (it would have allowed me to include it as a git submodule). Some options:
- Create a small git repository with only the Yasgui `build` directory and a github action that updates the repository each time a new Yasgui release is done. Then I can use this repository as a submodule in Oxigraph.
- Download at build time from unpkg or some similar repository. It's fairly simple but not very nice and packagers like Debian will hate that.
- Make the HTML UI optional in Oxigraph CLI and, if enabled use npm/webpack to build the user interface. A bit heavy but might be convenient if we want to improve the embedded HTML UI in the future.

Do you have better ideas?

MarcAntoine-Arnaud:
Good points!
I think the first point is the best to have a pre-build submodule.

In case for the 3rd point, it may be an alternative to rebuild locally based on a feature.
You can find here an example:
- Cargo feature declaration https://github.com/zmwangx/rust-ffmpeg-sys/blob/master/Cargo.toml#L34
- in the build.rs, based on the env feature variable: https://github.com/zmwangx/rust-ffmpeg-sys/blob/master/build.rs#L649 the build may be alternative build (inc. yasgui build). It maybe relevant to use a feature dependency like https://crates.io/crates/npm_rs.

Tpt:
Thank you! Let's move forward on the first point, then

ludovicm67:
@Tpt is it OK for you if I add it as an artifact on GitHub releases?

ludovicm67:
[...]

PR #879, opened 2024-05-21
Server: Vendors Yasgui instead of downloading it

Tpt:
Issue #847


Tpt:
@ludovicm67 @MarcAntoine-Arnaud This MR vendors Yasgui by using as submodule oxigraph/yasgui-release. oxigraph/yasgui-release has a [CRON action](https://github.com/oxigraph/yasgui-release/blob/main/.github/workflows/update.sh) that updates it automatically with the latest release.

PR #882, opened 2024-05-28
Implement Sophia traits

pchampin:
the long awaited bridge between Oxigraph and Sophia!

This indtroduce a feature `sophia` on `oxrdf` and `oxigraph`.
On `oxrdf`, this makes terms, triples and quad implement the corresponding traits in Sophia.
On `oxigraph`, it makes `Store` implement a number of traits describing datasets.

It is based on the stable version (0.8) of Sophia.


Tpt:
Thank you!

Quick question: the new parsers (`oxttl` and `oxrdfxml`) depends on `oxrdf`. If we add the `oxrdf` -> `sophia_api` dependency it means that if you make `sophia` depends on `oxttl` we end up with the `sophia` -> `oxttl` -> `sophia_api` dependency. I thought it is something we wanted to avoid.

pchampin:
> Quick question: the new parsers (`oxttl` and `oxrdfxml`) depends on `oxrdf`. If we add the `oxrdf` -> `sophia_api` dependency it means that if you make `sophia` depends on `oxttl` we end up with the `sophia` -> `oxttl` -> `sophia_api` dependency loop. I thought it is something we wanted to avoid.

Indeed, we want to avoid that.

My plan is for `sophia_turtle` `and `sophia_xml` to depend on `oxttl` and `oxrdfxml`, respectively, *without requiring* the `sophia` feature. Instead they will provide their own adapters for Sophia traits on `oxrdf` types, using wrappers. This will lead to a little code redundancy between Oxigraph and Sophia (basically, the content of [this file](https://github.com/oxigraph/oxigraph/blob/e4b37a9837e8bc9ba763725af188068dd0c91f79/lib/oxrdf/src/sophia.rs)), but this is an acceptable cost for allowing both projects to evolve independently. All the more than this redundant code is pretty straightforwd, and should therefore be very low-maintenance.

Tpt:
> My plan is for sophia_turtle and sophia_xmlto depend onoxttlandoxrdfxml, respectively, *without requiring* the sophiafeature. Instead they will provide their own adapters for Sophia traits onoxrdf` types, using wrappers. This will lead to a little code redundancy between Oxigraph and Sophia (basically, the content of [this file](https://github.com/oxigraph/oxigraph/blob/e4b37a9837e8bc9ba763725af188068dd0c91f79/lib/oxrdf/src/sophia.rs)), but this is an acceptable cost for allowing both projects to evolve independently. All the more than this redundant code is pretty straightforwd, and should therefore be very low-maintenance.

In this case, what is the advantage compared to just make `oxttl` and `oxrdfxml` implement the Sophia traits? Having some Oxigraph libraries implement the Sophia traits and other being wrapped by Sophia seems a bit weird and inconsistent to me.

pchampin:
It would indeed make a lot of sense to implement Sophia's `Parser` and `Serializer` traits in `oxttl` and `oxrdfxml`. I'll add this to my TODO list.

> In this case, what is the advantage [...]

If I follow your reasoning, I could just abandon `sophia_ttl` and `sophia_xml`, and rely directly on `oxttl` and `oxrdfxml` in `sophia` to provide parsers and serializers.

This would indeed be the cleaner solution, but it would make the release cycle more complex:

* I release a new version of `sophia_api`
* You release a new version `oxttl` and `oxrdfxml`, based on the new `sophia_api`
* I release a new version of `sophia` based on the new `oxttl` and `oxrdfxml`

That's the kind of choreography that I want to avoid. The duplication of code is not ideal, but I don't need to synchronize with you on each new release of Sophia (even if, to be fair, you do releases much more frequently than me :laughing:).

To alleviate this problem, I would add a feature to `sophia` on which `sophia_ttl` and `sophia_xml` would depend.
When people want to use `Oxigraph` together with `Sophia`, they would be encouraged to disable this feature and use `oxttl` and `oxrdfxml` directly instead.

> Having some Oxigraph libraries implement the Sophia traits and other being wrapped by Sophia seems a bit weird and inconsistent to me.

With the proposed solution above, you would be in one or the other situation :

* you use Sophia without Oxigraph ; in fact, you are using wrapped versions of `oxttl` and  `oxrdfxml` inside `sophia_ttl` and `sophia_xml`, but you don't even need to realize it ; or
* you use Oxigraph with Sophia, in which case you refrain from importing `sophia_ttl` or `sophia_xml`, and use the Oxigraph parsers/serializers directly (but you can also use other Sophia parsers/serializers, e.g. JSON-LD). There are no wrappers, only traits directly implemented by Oxigraph types.

pchampin:
oxttl now provides implementations for the Sophia traits. This was significantly harder than for oxrdf and oxigraph, as the notions of "parser" and "serializer" are slightly different in the two projects. But in the end, it works.

I marked this PR as draft until I make the same of oxrdfxml, then I think it will be ready.

I see that the debian_compatibility test keeps failing, apparently because sophia_api is not available as a debian package. Is this a strong requirement on your side? If so, I would appreciate some advice about how to fix it.

mkatychev:
@Tpt what is needed to move this forward? One of the immediate benefits is that RFC 1.0 canonicalization would be accessible (through [`sophia_c14n`](https://docs.rs/sophia_c14n/latest/sophia_c14n/)) for current `oxrdf::Graph` structures.

pchampin:
See also https://github.com/pchampin/sophia_rs/issues/23#issuecomment-2194711481

jonassmedegaard:
Debian maintainer of most RDF-related packages here.

> I see that the debian_compatibility test keeps failing, apparently because sophia_api is not available as a debian package. Is this a strong requirement on your side? If so, I would appreciate some advice about how to fix it.

The most elegant fix is of course to have sophia crates packaged in Debian :-)

I will prioritize that task at my end, but not sure how fast I will be, so best if you folks can do some workaround until it is in place.

Tpt:
> I will prioritize that task at my end, but not sure how fast I will be, so best if you folks can do some workaround until it is 

@jonassmedegaard This MR hides the Sophia traits implementation behind a feature so if this MR is merged, one would just need to not enable this feature on Debian

jonassmedegaard:
Ah, of course. Not a problem, then.

pchampin:
> The most elegant fix is of course to have sophia crates packaged in Debian :-)

Indeed :)

> I will prioritize that task at my end,

I appreciate, because I have no experience in doing that myself. Let me know if I can still be of any help.

> but not sure how fast I will be, so best if you folks can do some workaround until it is in place.

Of course. An obvious workaround would be to *not* include the feature in the Debian distribution of Oxigraph, so that the absence of Sophia in Debian is not a problem anymore...

Jezza:
I've love to see this to completion.

Anything that I can do to help move this forward?

PR #900, opened 2024-06-18
Parallel parsing of Turtle

magbak:
Added heuristic parallel parsing of the Turtle format based on finding splits of the Turtle file on every period.
Still missing comments and tests, will fix after feedback on API changes.


Tpt:
Thank you! May you resolve the merge conflicts :pray:

lib/oxttl/src/toolkit/parser.rs
```diff
@@ -43,7 +44,7 @@ impl<S: Into<String>> From<S> for RuleRecognizerError {
         }
     }
 }
-
+#[derive(Clone)]
```
Tpt, line 47: I am not sure all these `Clone` derives are useful anymore

lib/oxttl/src/turtle.rs
```diff
@@ -267,6 +268,130 @@ impl TurtleParser {
...
+    ///         count
+    ///     }).sum();
+    /// assert_eq!(2, count);
+    /// # Result::<_,Box<dyn std::error::Error>>::Ok(())
+    /// ```
+    #[allow(clippy::unwrap_in_result)]
+    pub fn parse_slice(
+        mut self,
+        slice: &[u8],
+    ) -> Result<Vec<FromSliceTurtleReader<'_>>, TurtleParseError> {
```
Tpt, line 342: Idea: instead of a custom struct we can maybe have a function in `TurtleParser`:
```rust
fn split_slice_for_parsing(mut self, slice: &[u8], target_parallelism: usize) -> Result<Vec<FromSliceTurtleReader<'_>>, TurtleParseError> {
```

this way we can avoid to duplicate all the options. `target_parallelism` would be the parallelism upper bound. It is nice imho to have it as a parameter to allow the users to customize it. We might suggest `std::thread::available_parallelism` as a sensible default.

lib/oxttl/src/turtle.rs
```diff
@@ -267,6 +268,130 @@ impl TurtleParser {
...
+        const MIN_PAR_SIZE: usize = 10_000;
```
Tpt, line 343: Instead of a `MIN_PAR_ITER` that is global what about splitting into something like `max(target_parallelism, slice.len() / 16384)`. An other option might be to always try to split by `target_parallelism` and let the user deal with minimal chunk size if they want

lib/oxttl/src/turtle.rs
```diff
@@ -267,6 +268,130 @@ impl TurtleParser {
...
+        const INCREMENT: usize = 1_000;
+
+        let slice_len = slice.len();
+        let threads = if slice_len >= MIN_PAR_SIZE {
+            // Prefixes must be determined before chunks, since determining chunks relies on parser with prefixes determined.
+            let mut end = INCREMENT;
+            let mut prefix_parser = self.parser.clone().parse();
```
Tpt, line 350: You can just use `parse_from_slice` here. It will have the same behavior and will avoid a copy.

lib/oxttl/src/turtle.rs
```diff
@@ -267,6 +268,130 @@ impl TurtleParser {
...
+            while !found_first_triple && slice_len >= (end + INCREMENT) {
+                if let Some(r) = prefix_parser.read_next() {
+                    r?;
+                    found_first_triple = true;
+                } else {
+                    prefix_parser.extend_from_slice(&slice[end..end + INCREMENT]);
+                    end += INCREMENT;
+                }
+            }
+            let prefixes: Vec<_> = prefix_parser.prefixes().collect();
```
Tpt, line 362: nit: no need to `.collect()`

lib/oxttl/src/toolkit/chunker.rs
```diff
@@ -0,0 +1,107 @@
...
+        let search_pos = last_pos + chunk_size;
+
+        if search_pos >= bytes.len() {
+            break;
+        }
+
+        let end_pos = match next_terminating_period(parser.clone(), &bytes[search_pos..]) {
+            Some(pos) => search_pos + pos,
+            None => {
+                return vec![(0, total_len)];
```
Tpt, line 52: I am confused by this, what about just using the end of file? I see two options:
- there is no data (only blank line): having an empty chunk is imho fine
- there is an error and having it inside of a chunk is fine

lib/oxttl/src/toolkit/chunker.rs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        };
+        offsets.push((last_pos, end_pos));
+        last_pos = end_pos;
+    }
+    offsets.push((last_pos, total_len));
```
Tpt, line 58: nit: maybe check it's not empty?

lib/oxttl/src/toolkit/chunker.rs
```diff
@@ -0,0 +1,107 @@
...
+    offsets
+}
+
+// Heuristically, we assume that a period is terminating (a triple) if we can start immediately after it and parse three triples.
```
Tpt, line 62: dumb question: why 3? I would guess 1 would work except if we are in a string that contains triples. But it's a nasty case

lib/oxttl/src/toolkit/chunker.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut rejected_periods = 0_u16;
+
+    let mut total_pos = 0;
+    if input.is_empty() {
+        return None;
+    }
+    loop {
+        if rejected_periods >= 1_000 {
+            return None;
+        }
```
Tpt, line 91: nit: a `for _ in 0..1_000` is slightly shorter

Tpt (commented):
Thank you!

lib/oxttl/src/turtle.rs
magbak: Should be fixed now.

lib/oxttl/src/turtle.rs
magbak: Done!

lib/oxttl/src/toolkit/parser.rs
magbak: Removed all clone derives from oxttl and reintroduced the ones needed to compile.

magbak: Still need to clone TurtleParser since parse_slice consumes it.

lib/oxttl/src/turtle.rs
magbak: Done

lib/oxttl/src/turtle.rs
magbak: Fixed!

lib/oxttl/src/toolkit/chunker.rs
magbak: Fixed

lib/oxttl/src/toolkit/chunker.rs
magbak: Saw they used the same logic in Polars CSV parsing to make the heuristic a bit more robust for a case like you mentioned. Perhaps not worth doing though :-) A triple inside a string would be escaped anyway and would not parse?

lib/oxttl/src/toolkit/chunker.rs
magbak: The idea here is that failing to find a terminating period (after 1_000 tries or end of file) reverts back to single threaded parsing. We may have correct file where we need to look at 1001 periods to find the next valid one..

lib/oxttl/src/turtle.rs
```diff
@@ -213,6 +215,78 @@ impl TurtleParser {
...
+    /// assert_eq!(2, count);
+    /// # Result::<_,Box<dyn std::error::Error>>::Ok(())
+    /// ```
+    #[allow(clippy::unwrap_in_result)]
+    pub fn split_slice_for_parsing(
+        mut self,
+        slice: &[u8],
+        target_parallelism: usize,
+    ) -> Result<Vec<FromSliceTurtleReader<'_>>, TurtleParseError> {
+        let min_par = max(target_parallelism, slice.len() / 16384);
```
Tpt, line 261: Isn't that `n_chunks`?

lib/oxttl/src/toolkit/chunker.rs
magbak: Added a check - hope this is what you meant!

lib/oxttl/src/toolkit/chunker.rs
Tpt: Ok! What about keeping the previous valid chunks  we found instead to fall back to fully single threaded parsing?

lib/oxttl/src/toolkit/chunker.rs
Tpt: If the triple is `<http://example.com/s> <http://example.com/p> <http://example.com/o> .` it does not need any parsing :(

lib/oxttl/src/toolkit/chunker.rs
magbak: Fixed now!

lib/oxttl/src/turtle.rs
magbak: Should be fixed now!

Tpt (approved):
Thank you! I believe I found a last possible source of bug and some small improvement ideas.

Ignore the debian_compatibility  failing test. This test is very flaky.

Cargo.toml
```diff
@@ -54,6 +54,7 @@ predicates = ">=2.0, <4.0"
 pyo3 = "0.21.2"
 quick-xml = "0.32"
 rand = "0.8"
+rayon = "1.10.0"
```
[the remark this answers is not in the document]
magbak: There is no 1.11 of Rayon yet, but you probably mean 1.10 :-)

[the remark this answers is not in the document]
magbak: Fixed that now

lib/oxrdfio/src/parser.rs
```diff
@@ -393,6 +393,89 @@ impl RdfParser {
...
+        }
+        let mut slice_quad_readers = Vec::with_capacity(from_quad_reader_kinds.len());
+        for parser in from_quad_reader_kinds {
+            slice_quad_readers.push(FromSliceQuadReader {
+                parser,
+                mapper: QuadMapper {
+                    default_graph: self.default_graph.clone(),
+                    without_named_graphs: self.without_named_graphs,
+                    blank_node_map: self.rename_blank_nodes.then(HashMap::new),
+                },
```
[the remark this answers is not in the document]
magbak: I see this is a serious bug / weakness now, giving the same blank nodes different names across threads.

A good idea to do it low level only - puts more demands on whoever uses (mostly me at this point) it to create a unified API to parse. I guess if you want to do this hack you can also handle that, and can handle renaming blank nodes on the client side.

lib/oxttl/src/turtle.rs
```diff
@@ -608,6 +678,7 @@ impl<'a> Iterator for FromSliceTurtleReader<'a> {
 /// assert_eq!(2, count);
 /// # Result::<_,Box<dyn std::error::Error>>::Ok(())
 /// ```
+
```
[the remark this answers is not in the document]
magbak: Fixed!

lib/oxttl/src/turtle.rs
```diff
@@ -213,6 +214,75 @@ impl TurtleParser {
...
+    ///             }
+    ///         }
+    ///         count
+    ///     })
+    ///     .sum();
+    /// assert_eq!(2, count);
+    /// # Result::<_,Box<dyn std::error::Error>>::Ok(())
+    /// ```
+    #[allow(clippy::unwrap_in_result)]
+    pub fn split_slice_for_parsing(
```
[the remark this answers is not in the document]
magbak: Fixed!

lib/oxttl/src/turtle.rs
```diff
@@ -213,6 +214,75 @@ impl TurtleParser {
...
+        target_parallelism: usize,
+    ) -> Result<Vec<FromSliceTurtleReader<'_>>, TurtleParseError> {
+        #[allow(clippy::decimal_literal_representation)]
+        let n_chunks = (slice.len() / 16384).clamp(1, target_parallelism);
+
+        if n_chunks > 1 {
+            // Prefixes must be determined before chunks, since determining chunks relies on parser with prefixes determined.
+            let mut from_slice_reader = self.clone().parse_slice(slice);
+            if let Some(r) = from_slice_reader.next() {
+                r?;
```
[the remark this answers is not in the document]
magbak: Fixed!

lib/oxttl/src/toolkit/chunker.rs
```diff
@@ -0,0 +1,103 @@
...
+            } else {
+                return false;
+            }
+        }
+        true
+    }
+    let mut total_pos = 0;
+    if input.is_empty() {
+        return None;
+    }
```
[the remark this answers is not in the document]
magbak: Fixed

lib/oxrdfio/src/parser.rs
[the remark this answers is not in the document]
magbak: Fixed this thing :-)

lib/oxrdfio/src/parser.rs
```diff
@@ -419,6 +419,7 @@ impl From<RdfFormat> for RdfParser {
 /// assert_eq!(quads[0].subject.to_string(), "<http://example.com/s>");
 /// # std::io::Result::Ok(())
 /// ```
+
```
Tpt, line 422: nit: weird whitespace

lib/oxttl/src/ntriples.rs
```diff
@@ -374,6 +374,7 @@ impl<'a> Iterator for FromSliceNTriplesReader<'a> {
 /// assert_eq!(2, count);
 /// # Result::<_,Box<dyn std::error::Error>>::Ok(())
 /// ```
+
```
Tpt, line 377: nit: weird whitespace

lib/oxrdfio/src/parser.rs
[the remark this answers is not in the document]
Tpt: Thank you!

Cargo.toml
```diff
@@ -54,6 +54,7 @@ predicates = ">=2.0, <4.0"
 pyo3 = "0.21.2"
 quick-xml = "0.32"
 rand = "0.8"
+rayon = "1.10"
 rayon-core = "1.11"
```
Tpt, line 58: I think you need to update the version to 1.12.1 to make MSV CI happy

lib/oxttl/src/turtle.rs
```diff
@@ -213,6 +214,75 @@ impl TurtleParser {
         }
     }
 
+    /// Creates a vector of iterators that may be used to parse a Turtle document slice in parallel.
+    /// To dynamically specify target_parallelism, use e.g. std::thread::available_parallelism
```
Tpt, line 218: nit:  use
```
[`std::thread::available_parallelism`]
```
 to get a link + add a dot at the end of the sentence

lib/oxttl/src/turtle.rs
```diff
@@ -213,6 +214,75 @@ impl TurtleParser {
...
+    /// Intended to work on large documents.
+    /// Can fail if there are prefixes that are not defined at the top of the document.
```
Tpt, line 220: Suggestion: "Can fail or return wrong results if there are prefixes or base iris that are not defined at the top of the document. or valid turtle syntax inside of literal values"

Tpt (commented):
Thank you! Sadly CI is not green yet and I allowed myself comme comments nits at this occasion, but feel free to ignore them.

Cargo.toml
magbak: Fixed!

lib/oxrdfio/src/parser.rs
magbak: Fixed

lib/oxttl/src/ntriples.rs
magbak: Fixed

lib/oxttl/src/turtle.rs
magbak: Fixed!

lib/oxttl/src/turtle.rs
magbak: Fixed!

Tpt:
Thank you! Sadly it needs a rebase before merging :(

magbak:
> Thank you! Sadly it needs a rebase before merging :(

Made into a single commit as agreed:
git checkout 00c95cb4a2db7a0ef88c2273185e073eeeac41b5 -- lib/oxttl/ 
also added rayon and fixed rayon core version for test.

Issue #925, opened 2024-07-15
UTF-16 Surrogate Pair Support


lubomirkurcak:
Oxigraph does not accept Unicode characters encoded as [UTF-16 surrogate pairs.](https://www.unicode.org/glossary/#surrogate_pair) Instead it treats each part as a separate unicode value.

| Unicode | UTF-16 surrogate pair | UTF-32 |
| --- | --- | --- |
| 👨 | `\uD83D\uDC68` | `\U0001F468` |

#### Unicode (works)

```
curl -v -X POST "localhost:7878/store" -H "Content-Type: application/n-triples" -d "<http://a> <http://b> \"👨\" ."
```

```
HTTP/1.1 201 Created
server: Oxigraph/0.4.0-alpha.7
content-length: 0
```

#### UTF-16 surrogate pair (fails)

```
curl -v -X POST "localhost:7878/store" -H "Content-Type: application/n-triples" -d "<http://a> <http://b> \"\uD83D\uDC68\" ."
```

```
HTTP/1.1 400 Bad Request
content-type: text/plain; charset=utf-8
server: Oxigraph/0.4.0-alpha.7
content-length: 148

Parser error between at line 1 between columns 25 and column 31: The escape sequence '\uD83D' is encoding D83D that is not a valid unicode character
```

#### UTF-32 (works)

```
curl -v -X POST "localhost:7878/store" -H "Content-Type: application/n-triples" -d "<http://a> <http://b> \"\U0001F468\" ."
```

```
HTTP/1.1 201 Created
server: Oxigraph/0.4.0-alpha.7
content-length: 0
```

#### Describe the solution you'd like
Oxigraph checks whether the character decoded from `\uXXXX` format is a surrogate character.[^1] If so, decode the following part of the surrogate pair and convert them to `char` using [`char::decode_utf16`](https://doc.rust-lang.org/std/primitive.char.html#method.decode_utf16).

In theory the handling could be a part of the `Err` branch here:

https://github.com/oxigraph/oxigraph/blob/5a240901ae45f3a4c3c23bc38fbc9e7f9f95330a/lib/oxttl/src/lexer.rs#L770-L773

#### Describe alternatives you've considered
Declare UTF-16 surrogate pair escape encoding unsupported and force users to handle transcoding on their side. Potentially create guides for how to format strings for various languages.

#### Additional context
[...]

PR #927, opened 2024-07-16
UTF-16 Surrogate Pair Support

lubomirkurcak:
Decode UTF-16 surrogate pairs and convert them to `char`s.
Linked issue: #925

#### Example
Store `\uD83D\uDC68`
```
curl -v -X POST "localhost:7878/store?graph=http://g" -H "Content-Type: application/n-triples" -d "<http://a> <http://b> \"\\uD83D\\uDC68\" ."
```

```
HTTP/1.1 204 No Content
server: Oxigraph/0.4.0-alpha.7
```

Retrieve correct unicode
```
curl -v -X POST "localhost:7878/query" -H "Content-Type: application/sparql-query" -d "select ?o where { graph <http://g> { ?s ?p ?o } }"
```

```
HTTP/1.1 200 OK
content-type: application/sparql-results+json
server: Oxigraph/0.4.0-alpha.7

{"head":{"vars":["o"]},"results":{"bindings":[{"o":{"type":"literal","value":"👨"}}]}}
```

### DIFF
--- a/lib/oxttl/src/lexer.rs
+++ b/lib/oxttl/src/lexer.rs
@@ -769,7 +769,18 @@ impl N3Lexer {
         match *data.get(1)? {
             b'u' => match Self::recognize_hex_char(&data[2..], 4, 'u', position) {
                 Ok(c) => Some((5, Ok(c?))),
-                Err(e) => Some((5, Err(e))),
+                Err(e) => {
+                    let lenient = true;
+                    if lenient {
+                        match Self::recognize_utf16_surrogate_pair(&data[2..], position) {
+                            Ok(None) => Some((5, Err(e))),
+                            Ok(Some(c)) => Some((11, Ok(c))),
+                            Err(e) => Some((5, Err(e))),
+                        }
+                    } else {
+                        Some((5, Err(e)))
+                    }
+                }
             },
             b'U' => match Self::recognize_hex_char(&data[2..], 8, 'u', position) {
                 Ok(c) => Some((9, Ok(c?))),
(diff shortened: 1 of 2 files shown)


lib/oxttl/src/lexer.rs
```diff
@@ -823,6 +834,93 @@ impl N3Lexer {
...
+        if matches!(surrogate_high, 0xDC00..=0xDFFF) {
+            return Err((
+                position..position + 6,
+                format!("UTF-16 low-surrogate escape sequence '\\u{val_high}' must be preceded by a high-surrogate"),
+            )
+                .into());
+        }
+
+        // https://www.unicode.org/glossary/#high_surrogate_code_point
+        if matches!(surrogate_high, 0xD800..=0xDBFF) {
```
Tpt, line 865: If false we are returning `Ok(None)` and not an error even if we know it's an error. I believe it will lead to the parser doing an infinite loop on invalid code points.

lib/oxttl/src/lexer.rs
```diff
@@ -823,6 +834,93 @@ impl N3Lexer {
...
+            if data.len() < 10 || data[4] != b'\\' || data[5] != b'u' {
```
Tpt, line 866: Note that if `data.len() < 10` might be because we have not buffered enough data. In this case we should return `Ok(None)`

lib/oxttl/src/lexer.rs
```diff
@@ -823,6 +834,93 @@ impl N3Lexer {
...
+        })?;
+
+        // https://www.unicode.org/glossary/#low_surrogate_code_point
+        if matches!(surrogate_high, 0xDC00..=0xDFFF) {
+            return Err((
+                position..position + 6,
+                format!("UTF-16 low-surrogate escape sequence '\\u{val_high}' must be preceded by a high-surrogate"),
+            )
+                .into());
+        }
```
Tpt, line 862: Open question: do we care about validating this here? I believe`char::decode_utf16` should already fail in this case.

lib/oxttl/src/lexer.rs
```diff
@@ -823,6 +834,93 @@ impl N3Lexer {
...
+                    "Surrogate pair should combine to exactly one character"
+                );
+
+                let mut b = [0; 2];
+                c.encode_utf16(&mut b);
+                debug_assert_eq!(
+                    b,
+                    [surrogate_high, surrogate_low],
+                    "The decoded character should encode to the same surrogate pair"
+                );
```
Tpt, line 910: Nit: Is it really useful? It's mostly asserting `char::decode_utf16` is valid

Tpt (commented):
Thank you so much for pushing this so quickly!

lib/oxttl/src/lexer.rs
lubomirkurcak: Originally I wanted to fall back to the original error message from `recognize_hex_char` by mapping `Ok(None)` to it. (We cannot do that because `Ok(None)` is actually a signal to the parser)

`recognize_utf16_surrogate_pair` only triggers when `recognize_hex_char` fails. I wonder what error message should `recognize_utf16_surrogate_pair` print when the first escape sequence is not a surrogate character. Currently it is set to print:
```
The escape sequence '\\u{val_high}' is not a UTF-16 surrogate
```
However, I don't think there is another reason why parsing `\uXXXX` can fail though other than it being a surrogate, so maybe this is ok? (It will never trigger anyway)

lib/oxttl/src/lexer.rs
lubomirkurcak: Removed.

lubomirkurcak:
I am not sure why clippy complains about range checks because they should be covered by early returns.

To sidestep clippy we can use `let ... else` instead:
```diff
-        if data.len() < 4 {
-            return Ok(None);
-        }
-
-        let val_high = str_from_utf8(&data[..4], position..position + 6)?;
+        let Some(val_high_slice) = data.get(..4) else {
+            return Ok(None);
+        };
+        let val_high = str_from_utf8(val_high_slice, position..position + 6)?;
```

which in this case is zero-cost.

The next case causes two redundant bound checks:

```rust
            if data.len() < 10 { // <-- because of this
                return Ok(None);
            }

            if data[4] != b'\\' || data[5] != b'u' { // <-- compiler/clippy should know they can access these without bounds checks
                return Err((
                    position..position + data.len().min(10) + 2,
                    format!(
                        "UTF-16 surrogate escape sequence '\\u{val_high}' must be followed by another surrogate escape sequence"),
                )
                    .into());
            }

            let val_low = str_from_utf8(&data[6..10], position + 6..position + 12)?;
```

I don't think this function warrants optimizing two extra bound checks, so I would suggest replacing `[]`s with `get`s to appease clippy.

lib/oxttl/src/lexer.rs
Tpt: Yes, I also think it's ok. Thank you!

Tpt:
Aside: Don't bother with code coverage, I can easily add a few tests in a follow up MR.

Tpt (commented):
Happy to merge the MR when Clippy is happy

lib/oxttl/src/lexer.rs
```diff
@@ -823,6 +835,84 @@ impl N3Lexer {
...
+
+                debug_assert_eq!(
+                    chars.next(),
+                    None,
+                    "Surrogate pair should combine to exactly one character"
+                );
+
+                Ok(Some(c))
+            } else {
+                unreachable!();
```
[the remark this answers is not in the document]
lubomirkurcak: clippy complained about unwraps, but was fine with unreachable

Tpt:
Thank you for the MR!

PR #929, opened 2024-07-19
Adds test coverage for unchecked parsing

Tpt:


Tpt:
CC @lubomirkurcak

cli/src/main.rs
```diff
@@ -3008,18 +3008,15 @@ mod tests {
         )
         .with_header(HeaderName::ACCEPT, "application/n-triples")?
         .build();
-        server.test_body(
-            request,
-            "<http://example.com/s> <http://example.com/p> \"\u{1f468}\" .\n",
-        )?;
+        server.test_body(request, "<s> <p> \"\u{1f468}\" .\n")?;
```
lubomirkurcak, line 3011: I'm not sure application/n-triples supports BASE IRIs

Tpt: Indeed. But the lenient mode should not fail on relative IRIs and just allow them. Thank you!

PR #968, opened 2024-08-20
CLI: Generate on build autocomplete and man

Tpt:
Autocompletion for various shells are generated on build in the `target/{debug,release}/build/oxigraph-cli-<hash>/out/complete` directory.
Similarly, man pages are generated in the `target/{debug,release}/build/oxigraph-cli-<hash>/out/man` directory.


Tpt:
CC @jonassmedegaard Would it be easy to integrate these generated autocompletions and man in the debian package?

jonassmedegaard:
Wonderful!

Yes, that should be easy to integrate with the Debian package (haven't tested yet, only skimmed the source code). Thanks for asking!

By the way, please consider shaping at least the initial paragraph as imperative (I think, see below) not passive, and omit final full stop.
Example: "Creates database backup into a target directory." -> "Create database backup into a target directory"

I am not really sure if the correct term is "imperative" - what I mean is to do similar to `man find` and `man git-commit`.
Please do tell if better to create a separate issue for that...

Tpt:
@jonassmedegaard Thank you! Done in 0d0cc23a1932fbe41a64b1446cfe9182b95d5529

PR #969, opened 2024-08-21
CLI: Increases max SPARQL body size to 128MB

Tpt:


ch-sander:
Great. This is going to be published with the latest docker image soon?

Tpt:
@ch-sander It will be in the `ghcr.io/oxigraph/oxigraph:main` docker image as soon as this MR is merged

PR #970, opened 2024-08-21
Drop pyoxigraph build for Android target.

chanijjani:


Tpt:
If you need to fix some stuff to make Oxigraph work on Android, it would be very welcome to upstream them :)

Issue #973, opened 2024-08-22
Set `union-default-graph` to `true` by default for endpoint


ch-sander:
https://github.com/oxigraph/oxigraph/blob/591a13115c359095ae2b91244df905a6f530b6e7/cli/src/main.rs#L1060

I might be useful to set `union-default-graph` to `true` by default for an endpoint for all queries. can this be set in some global config?

Issue #948, opened 2024-07-31
Enpoint access to pyoxigraph RDF Store


ch-sander:
A beginner's question: I have successfully loaded my RDF in a Store with pyoxigraph (local directory).

In my docker compose I have put 
```yaml
    volumes:
      - ./oxigraph_store:/data
```
I can access my data by `store.query` with pyoxigraph but any access via API `(http://localhost:7878/query`) fails.

**How can I access my RDF store via the endpoint?**

ch-sander:
a file permission issue. The oxigraph server cannot run in ro-mode

Tpt:
You can use the "serve-read-only" command instead of "serve" to serve a directory without write access

ch-sander:
Thanks, that solved the initial problem!
I would like to use the bulk load from pyoxigraph (as working in a flask environment) and have the API endpoint as a read-only for a ResearchSpace instance.
Yet:

- the oxigraph docker container needs to restart to discover the new data from the pyoxigraph `Store`
- the /store call from the endpoint which might be designed for bulk upload doesn't perform on my data, I get a 201 but no data is returned from queries. Some of the .log files are around 400mb, so I suspect them to decode the errors of the parser. I tried with a decent file both in ttl and trig (50mb in ttl). In smaller files, it works. But without being able to read the logs, I would not like to start graph per graph to catch the problem. blazegraph can load any of then, and they were parsed from json-ld with rdflib.

If you have any idea how I could have such a dynamic interchange or interface between the pyoxigraph Store and a public read-only sparql endpoint, while the pyoxigraph app is run in the same docker network as the oxigraph service...

Otherwise a great app that is apt to replace blazegraph (my own test actually show a huge performance difference).

Tpt:
Amazing!

> the /store call from the endpoint which might be designed for bulk upload doesn't perform on my data, I get a 201 but no data is returned from queries.

Beware, if you POST a RDF file like Turtle or NTriples, the SPARQL graph store HTTP protocol specifies that they get inserted into a new named graph and not the default graph. The 201 response should provide a `Location` header with the created graph name. To add triples to the default graph, you should POST to `/store?default`. It's a quite confusing behavior but I assumed not following the standard would be even more confusing.

> the oxigraph docker container needs to restart to discover the new data from the pyoxigraph Store

[...]

PR #975, opened 2024-08-22
CLI: Add the union-default-graph option to make SPARQL queries use the union of all graphs as the default graph

Tpt:
Issue #973

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -24,6 +24,11 @@ pub enum Command {
         /// Allows cross-origin requests
         #[arg(long)]
         cors: bool,
+        /// If the SPARQL queries should look for triples in all the dataset graphs by default (ie. without `GRAPH` operations)
+        ///
+        /// This is equivalent as setting the union-default-graph option in all SPARQL queries
+        #[arg(long)]
+        union_default_graph: bool,
     },
     /// Start Oxigraph HTTP server in read-only mode
     ///
@@ -39,6 +44,11 @@ pub enum Command {
         /// Allow cross-origin requests
         #[arg(long)]
         cors: bool,
+        /// If the SPARQL queries should look for triples in all the dataset graphs by default (ie. without `GRAPH` operations)
+        ///
+        /// This is equivalent as setting the union-default-graph option in all SPARQL queries
+        #[arg(long)]
+        union_default_graph: bool,
     },
     /// Create a database backup into a target directory
     ///
@@ -169,6 +179,9 @@ pub enum Command {
         /// Beware, computing the statistics adds some overhead to the evaluation runtime.
         #[arg(long)]
         stats: bool,
+        /// If the SPARQL queries should look for triples in all the dataset graphs by default (ie. without `GRAPH` operations)
+        #[arg(long)]
+        union_default_graph: bool,
     },
     /// Execute a SPARQL update against the store
     Update {
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -51,6 +51,7 @@ pub fn main() -> anyhow::Result<()> {
             location,
             bind,
             cors,
+            union_default_graph,
         } => serve(
             if let Some(location) = location {
                 Store::open(location)
@@ -60,12 +61,20 @@ pub fn main() -> anyhow::Result<()> {
             &bind,
             false,
             cors,
+            union_default_graph,
         ),
         Command::ServeReadOnly {
             location,
             bind,
             cors,
-        } => serve(Store::open_read_only(location)?, &bind, true, cors),
+            union_default_graph,
+        } => serve(
+            Store::open_read_only(location)?,
+            &bind,
+            true,
+            cors,
+            union_default_graph,
+        ),
         Command::Backup {
             location,
             destination,
@@ -250,6 +259,7 @@ pub fn main() -> anyhow::Result<()> {
             explain,
             explain_file,
             stats,
+            union_default_graph,
         } => {
             let query = if let Some(query) = query {
                 query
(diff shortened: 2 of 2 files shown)


ch-sander:
Wonderful. How is the CLI configured then? An environment variable?
Sorry for the basic question.

Tpt:
Just set the `--union-default-graph` option when starting Oxigraph server

Tpt:
@ch-sander Hey! I was not aware you were using `serve-secondary` (or of any user of it). I dropped this command from the 0.4.0-rc.1 because of all the bugs I encountered with it and RocksDB 9. Distributions have migrated to RocksDB 9 and I did not wanted to ship a badly broken feature. Can you use easily `serve` or `serve-read-only` or do you badly need it?

ch-sander:
What is great is that a new store is target for queries immediately, so no restart of service needed. That's the main reason, I use it. But that might also be the thing to cause trouble.

It's also used by the app I run within my service: 
https://github.com/epoz/shmarql/blob/8811fddb8f7f4b104e504437564616cd2f290776/src/app/main.py#L73-L84

Maybe there's a work-around for both cases? I guess the shmarql exception handling could also just fall back to `read_only` @epoz ?

Tpt:
@ch-sander If I understand your code correctly, you are loading some data into Oxigraph when starting the script and then only do read-only access. If yes, I see two approaches:
- if you only have a few triples (less than a million) I would just load the data into memory each time you start the script, maybe with a on-disk cache for the data you load from the web.
- if you have more triples what you can do is that the first process open the store as read/write, load the file and then create a file with a well known name. The other processes would just wait with a sleep loop for this file to exist before opening the store as read only. Idea:
```python
try:
    GRAPH = px.Store(STORE_PATH)
    # Do loading ....
   with open(MY_WELL_KNOWN_FILE, "wt") as fp:
       fp.write(" ") # Create the file
except OSError:
   while not os.path.exists(MY_WELL_KNOWN_FILE):
      sleep(1)
  GRAPH = px.store.read_only(STORE_PATH)
```
I hope it helps and that I have understood your use-case well

ch-sander:
Thanks, but the code is @epoz 's -- my use case is much simpler: I load triples into the store with pyoxigraph (similar to shamrql) but with oxigraph server, running on its own, and having set the location to this store fed by pyoxigraph, I want to have this data ready for a query without restarting oxigraph server. You suggested the serve-secondary solution, I guess, here: https://github.com/oxigraph/oxigraph/issues/948#issuecomment-2262205513

Can the server be triggered to realod the store from location without restart and http only?

Tpt:
@ch-sander Thank you for the context. I missed it. Sadly a reload from location cannot be triggered without a restart (this kind of reload without secondary system would be in fact more or less a restart). If you don't have too many data, the easiest way is maybe to send the extra data to the HTTP server as RDF files or SPARQL updates.

To add graph data to the default graph: `curl -X POST -H "content-type: text/turtle" http://localhost:7878/store?default`
To add dataset data: `curl -X POST -H "content-type: application/trig" http://localhost:7878/store`
To add graph data to a named graph: `curl -X POST -H "content-type: text-turtle" http://localhost:7878/store?graph=http://example.com`

Add `?no_transaction&lenient` if you want faster loading to disable transactional updates and most data validations.

ch-sander:
I will have to find a way. The files are/will be c. 500mb in ttl. I think I remember `/store` can also load from remote. And I understand uploads via `/store` will work also if the server is run `serve-read-only`, right? Or would I need to `serve` and have a nginx (or other proxy) config to virtually disallow access to `store`?

Tpt:
> The files are/will be c. 500mb in ttl.

It should be fine via `/store`. If you set the `no_transaction` option, parsing and loading will be streamed so memory consumption should be ok.

> I think I remember /store can also load from remote.

You can indeed run SPARQL UPDATE `LOAD` command.

> And I understand uploads via /store will work also if the server is run serve-read-only, right?

Sadly not. You need a read/write server via the `serve` option and a proxy.

A quick and dirty approach might be to still do the loading via pyoxigraph or something similar and restart the read-only server when needed. The restart operation should be fast (less than a second).

ch-sander:
Thanks, will do so. The updates will occur quite regularly, the restart time is not the issue, but it safely needs to be triggered from remote via http. I need to find a good solution. 

Previously the pyoxigraph was part of fastapi app behind http authentication on a VM with a store directory mounted to another VM with Oxigraph server served secondary from the mount.

Tpt:
Quick thought: to avoid to have two Oxigraph processes, you might write in fastapi an implementation of the [SPARQL HTTP protocol]([http://www.w3.org/TR/sparql11-protocol/](https://www.w3.org/TR/sparql11-protocol/) to serve SPARQL queries from the same app. Implementing it only for queries should not be very hard.

ch-sander:
It's wonderful to have you reflecting on my use case, especially as I'm a beginner.

The shmarql app (fastapi) that runs beneath my own app to ingest RDF files into the `/store` via http has already an oxigraph-based SPARQL endpoint. I could well integrate this as endpoint for ResearchSpace, but I thought the "actual" oxigraph server is better maintained and more stable for production. E.g., shmarql does not support the flags for union-default or graph-uri (can be done easily, I know).

But with all these options, I will find a solution that fits and this without oxigraph having a buggy feature such as `serve-secondary`. If you ever resurrect it, maybe send me a note (and I will check the CHANGELOG).

Tpt:
Thank you for the nice reply!

The SPARQL protocol is quite stable and fastapi is very well maintained and stable so I tend to think it might be as good as Oxigraph http server on the stability side.

I will ping you if I resurect `serve-secondary`!

PR #1011, opened 2024-09-26
JS: Allows setting query named and default graphs

Tpt:


js/src/store.rs
```diff
@@ -50,6 +50,8 @@ export class Store {
         options?: {
             base_iri?: NamedNode | string;
             results_format?: string;
+            default_graph?: BlankNode | DefaultGraph | NamedNode | Iterable<BlankNode | DefaultGraph | NamedNode>;
+            named_graphs?: Iterable<BlankNode | NamedNode>;
```
tpluscode, line 54: Curious, how do blank nodes make sense as query context?

tpluscode: Also, nitpick, I'd propose to keep a consistent form. In SPARQL protocol both are singular
```suggestion
            default_graph?: BlankNode | DefaultGraph | NamedNode | Iterable<BlankNode | DefaultGraph | NamedNode>;
            named_graph?: Iterable<BlankNode | NamedNode>;
```

Tpt: Here Oxigraph considers blank node as terms in the queried dataset, opposite to the query string where blank nodes are assumed to be variables. This provides a way to query specifically a graph identified by a blank node.

For example if `default_graph: [blankNode("a)]` the query will use the named graph `_:a` as the default graph.

Tpt: >Also, nitpick, I'd propose to keep a consistent form and type. In SPARQL protocol both are singular.

Good point. The Rust and Python APIs are plural (`named_graphs`). Hence, I used plural for consistency. These parameters are not exactly the same as the SPARQL protocol (they allows blank nodes and the default graph).

tpluscode: Right. I don't think I've seen blank node graph identifiers outside of n3 rules 🤔

Tpt: Yes, it's quite rare but supported by RDF 1.1

PR #1039, opened 2024-10-12
Minimal naive geosparql support

Tpt:
`geof:sfEquals`, `geof:sfContains` and `geof:sfWithin` functions working on both WKT and GeoJSON

The implementation is in a plug-in `spargeo` crate. The HTTP server depends on it behind an enabled-by-default option


lib/spargeo/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Parse a WKT literal including reference system http://www.opengis.net/def/crs/OGC/1.3/CRS84
+fn parse_wkt_literal(value: &str) -> Option<Geometry> {
+    let mut value = value.trim_start();
+    if let Some(val) = value.strip_prefix('<') {
+        // We have a reference system
+        let (system, val) = val.split_once('>').unwrap_or((val, ""));
+        if system != "http://www.opengis.net/def/crs/OGC/1.3/CRS84" {
+            // We only support CRS84
```
ch-sander, line 72: Is that a relevant constraint?

testsuite/oxigraph-tests/geosparql/sf_contains.rq
```diff
@@ -0,0 +1,9 @@
+PREFIX geo: <http://www.opengis.net/ont/geosparql#>
+PREFIX geof: <http://www.opengis.net/def/function/geosparql/>
+
+SELECT ?left ?right (geof:sfContains(?left, ?right) AS ?result) WHERE {
+    VALUES (?left ?right) {
+        ("Polygon((0 0, 0 3, 3 3, 3 0))"^^geo:wktLiteral "Point(1 2)"^^geo:wktLiteral)
```
ch-sander, line 6: What about MULTIPOLYGON and POLYGON Z?

testsuite/oxigraph-tests/geosparql/sf_equals.srx
```diff
@@ -0,0 +1,43 @@
...
+            <binding name="result">
+                <literal datatype="http://www.w3.org/2001/XMLSchema#boolean">false</literal>
+            </binding>
+        </result>
+        <result>
+            <binding name="left">
+                <literal datatype="http://www.opengis.net/ont/geosparql#wktLiteral">Point(-83.38 33.95)</literal>
+            </binding>
+            <binding name="right">
+                <literal datatype="http://www.opengis.net/ont/geosparql#geoJSONLiteral">{"type": "Point", "coordinates": [-83.38,33.95]}</literal>
```
ch-sander, line 36: A GeoJSON will be a xsd:string literal? Or a RDF of it following the JSON-LD specs?

Tpt: [According to GeoSPARQL](https://docs.ogc.org/is/22-047r1/22-047r1.html#_rdfs_datatype_geogeojsonliteral) it is a RDF literal with datatype http://www.opengis.net/ont/geosparql#geoJSONLiteral like `"{\"type\": \"Point\", \"coordinates\": [-83.38,33.95]}"^^geo:geoJSONLiteral`

testsuite/oxigraph-tests/geosparql/sf_contains.rq
Tpt: `MULTIPOLYGON` definitely works and I need to check for `POLYGONZ` but [it seems supported by the WKT parser I use](https://github.com/georust/wkt/blob/b2f73990d0f0b3b800b09d7e868291dce4f3d263/src/lib.rs#L225) so I have good hope.

lib/spargeo/src/lib.rs
Tpt: Good point. I guess that if the two arguments have the same system then computation should be valid. It seems a nice topic for a follow up PR.

PR #1082, opened 2024-11-06
Memory store: avoid Arc in linked lists

Tpt:
It causes stack overflows because of too large chain of drop() calls

### DIFF
--- a/lib/oxigraph/src/storage/memory.rs
+++ b/lib/oxigraph/src/storage/memory.rs
@@ -14,7 +14,7 @@ use std::error::Error;
 use std::hash::{BuildHasherDefault, Hash, Hasher};
 use std::mem::transmute;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, Mutex, RwLock, Weak};
 
 /// In-memory storage working with MVCC
 ///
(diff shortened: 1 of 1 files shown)


lib/oxigraph/src/storage/memory.rs
```diff
@@ -30,15 +30,15 @@ pub struct MemoryStorage {
 
 struct Content {
     quad_set: DashSet<Arc<QuadListNode>, BuildHasherDefault<FxHasher>>,
```
adamreichold, line 32: As a general design question beyond the bug fix, I wonder if a concurrent slab would be a good data structure here so replace the pointers by indexes to avoid the reference counter memory and coherence traffic?

lib/oxigraph/src/storage/memory.rs
```diff
@@ -307,9 +307,9 @@ impl MemoryStorageReader {
         // By subject chain
         let mut count_last_by_subject = 0;
         for entry in &self.storage.content.last_quad_by_subject {
-            let mut next = Some(Arc::clone(&entry.value().0));
+            let mut next = Some(entry.value().0.clone());
```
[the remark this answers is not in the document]
Tpt: No reason outside of lazyness and that `Weak<T>` does not deref to `&T` so there is no risk of method name clash. Clippy is complaining to I switched to the associated function syntax.

lib/oxigraph/src/storage/memory.rs
Tpt: Yes! It's definitely a good idea! Will add it to my TOOD. Thank you!

ashleysommer:
@Tpt Thanks for that! This looks like a good solution.

PR #1087, opened 2024-11-08
enable sparql query of hdt storage

GregHanson:
Adding Support for SPARQL querying for HDT storage.

I have taken over the work and am submitting on behalf of @donpellegrino 

There's been some discussion around this feature on gitter in the past ([thread](https://matrix.to/#/!XCeJUkelgobygpWLrl:gitter.im/$j2SXyh7NW5MAVN8-vF4QtnbS8erKv-5-ieMrfzGlebQ?via=gitter.im&via=matrix.org&via=matrix.keiron.xyz))

At the time, the underlying [HDT library](https://github.com/KonradHoeffner/hdt) had a dependency on rust nightly, but the requirement is no more with the latest release!

First time submitting to oxigraph so please let me know if there is anything I can do to help !


Tpt:
Amazing! Thank you! I will have a look at it in the next few days

donpellegrino:
@GregHanson - thanks for putting this all together and getting it ready to go!

GregHanson:
cc @KonradHoeffner since we leverage their HDT library implementation

Tpt:
Thank you again for putting together this PR and sorry for the review delay.

After thinthing about it, I am not sure integrating HDT support into Oxigraph like it is now is a good idea. It's kind of a hack on top of the current Oxigraph internals and it will be likely painful to maintain.

However, I am currently working on making Oxigraph SPARQL evaluator standalone, in a new `spareval` crate. It will allow to execute SPARQL query on any storage system implementing a given trait, more or less like your MR does ([WIP version here](https://github.com/oxigraph/oxigraph/blob/f4ce1b30e1cdf40d9b035b2865d1a665898b5685/lib/spareval/src/dataset.rs#L12)). This should make SPARQL querying of HDT much easier and performant: Instead of decoding every term into an Oxigraph `EncodedTerm`, the HDT implementation can just return the internal integer encoding of the term and fetch the term value only for expression evaluation and returning the results.

What do you think about it?

GregHanson:
@Tpt thanks for the feedback, might take to a little bit to fully understand the impact to our proposed HDT changes, but my main take away is that instead of defining a new `pub trait DatasetView` like the PR currently does, you have something WIP that introduces a similar [pub trait SparqlQueryDataset](https://github.com/oxigraph/oxigraph/compare/main...spareval#diff-c3223d3961eee62033cde56e51da2b7b57acebabe4b10bd5f87be8bb2ab72ed5R12) the underlying HDT implementation could use instead?

If so, what's the timeline for the [spareval](https://github.com/oxigraph/oxigraph/tree/spareval) feature branch? Any issues for tracking that work?

Tpt:
@GregHanson 

> you have something WIP that introduces a similar [pub trait SparqlQueryDataset](https://github.com/oxigraph/oxigraph/compare/main...spareval#diff-c3223d3961eee62033cde56e51da2b7b57acebabe4b10bd5f87be8bb2ab72ed5R12) the underlying HDT implementation could use instead?

Yes!

> If so, what's the timeline for the [spareval](https://github.com/oxigraph/oxigraph/tree/spareval) feature branch? Any issues for tracking that work?

No issue tracking this yet. This is my main goal for Oxigraph right now. I hope for it to be done in the next couple of months. The basics are here, what is missing is expression evaluation support, `SERVICE` clauses, updates and migrating Oxigraph to it.

GregHanson:
Great! Can you cc me on any PR's or when the branch starts to stabilize? I would be interested in testing things out once it's available!

donpellegrino:
> However, I am currently working on making Oxigraph SPARQL evaluator standalone, in a new `spareval` crate. It will allow to execute SPARQL query on any storage system implementing a given trait, more or less like your MR does ([WIP version here](https://github.com/oxigraph/oxigraph/blob/f4ce1b30e1cdf40d9b035b2865d1a665898b5685/lib/spareval/src/dataset.rs#L12)). This should make SPARQL querying of HDT much easier and performant: Instead of decoding every term into an Oxigraph `EncodedTerm`, the HDT implementation can just return the internal integer encoding of the term and fetch the term value only for expression evaluation and returning the results.

That is a nice insight. Returning the internal integer encoding from HDT should be much more efficient for basic retrieval. I would expect other parts of the query plan such as evaluation of SPARQL [SUBSTR()](https://www.w3.org/TR/sparql11-query/#func-substr), or other such functions on RDF terms, might still need to work on some intermediate data structure. The [WIP Optional Methods](https://github.com/oxigraph/oxigraph/blob/f4ce1b30e1cdf40d9b035b2865d1a665898b5685/lib/spareval/src/dataset.rs#L52) seems like a good place to introduce functions that could be optimized for whichever persistence layer implements the Trait. Another use case could be using [MPI](https://github.com/rsmpi/rsmpi), or other parallel and distributed APIs, on the intermediate results so that queries could process intermediate results that are too large for a single compute node to handle.

Let @GregHanson and I know if we can be of any help with the change. The WIP design is very exciting!

Tpt:
> Can you cc me on any PR's or when the branch starts to stabilize?

@GregHanson With pleasure! Feedbacks are very welcome!

> Returning the internal integer encoding from HDT should be much more efficient for basic retrieval

@donpellegrino Yes! An other possible significant performance improvement I see is to read the HDT files using mmap to avoid opening the files multiple times and to skip syscalls as much as possible (opening the HDT files on `.clone()` looks very scary to me and is a performance trap if `clone` is done a lot).

> The [WIP Optional Methods](https://github.com/oxigraph/oxigraph/blob/f4ce1b30e1cdf40d9b035b2865d1a665898b5685/lib/spareval/src/dataset.rs#L52) seems like a good place to introduce functions that could be optimized for whichever persistence layer implements the Trait

Yes! This is my idea. I plan to add a bunch of methods for optimization places and provide for each of them a working default implementation. For most functions I expect them to work on the `ExpressionTerm` enum that is created with the `externalize_expression_term` method.

>Another use case could be using [MPI](https://github.com/rsmpi/rsmpi), or other parallel and distributed APIs, on the intermediate results so that queries could process intermediate results that are too large for a single compute node to handle.

This looks like a huge project. A first step if likely to introduce some parallelism in the query evaluator to use more than a single core per query.

> Let GregHanson and I know if we can be of any help with the change.

Thank you! I think you will already have a bit of work with just the HDT part.

Tpt:
@donpellegrino @GregHanson the stand-alone SPARQL evaluator is now released: https://docs.rs/spareval/

GregHanson:
Thanks for the update @Tpt ! I saw some of the recent PRs and was wondering where it stood. I'll try to take a look at integrating your changes with our HDT work sometime this week!

GregHanson:
closing as duplicate of https://github.com/oxigraph/oxigraph/pull/1211

PR #1143, opened 2025-01-09
feat: `no-duration-coercion` flag

### LINKED ISSUE #1140: `oxigraph` build without IO and FFI
**Is your feature request related to a problem? Please describe.**

I would like to run oxigraph inside the [risczero zkvm](https://risczero.com) to be able to build [verifiable SPARQL results](https://github.com/jeswr/queryable-credentials?tab=readme-ov-file#initial-design-thoughts-for-a-queryable-api).

To run code inside the VM, any packages used must:
 - Use only pure rust code (no FFI / C++ bindings allowed)
 - Not have any IO components; the triples are passed to the VM in memory.

**Describe the solution you'd like**

A build or version of Oxigraph that allows me to run SPARQL queries over in-memory data, and does not include any IO functionality in the package.

It may be that is what I am after, in which case, would it be possible to publish that package to crates.io?

jeswr:
This flag enables the use of `oxsdatatypes` in an environment where `std::time` is unavailable.

Closes #1140. With this change spareval can be run inside a zkvm - as demonstrated in https://github.com/jeswr/risc0-sparql-poc.

Please squash when merging.


jeswr:
@Tpt I'm running on the assumption that the current test failures are not related to changes made in this PR.

Please let me know if I am wrong on this and I can investigate.

jeswr:
> I have to say I am a bit surprised, std::time::Duration does not rely on any syscall.

The issue is that none of `std::time` is implemented in the risc0, see https://github.com/risc0/risc0/issues/2714

> A quick search seems to suggest that some crates are using cfg(target_os = "zkvm").

Good idea - working on it now

jeswr:
Can confirm that this change works downstream!

Tpt (approved):
Thank you so much! This is much nicer this way!

PR #1160, opened 2025-01-25
CI: Fuzz: Properly save fuzz cache

Tpt:


Tpt:
Was already working, got confused

PR #1166, opened 2025-01-29
Update README.md to mention Docker

### LINKED ISSUE #1165: Document Docker image
The README should briefly document there is a Docker image and what's included. The cli? A web server? If also the latter, which port is used and where is data stored to mount a volume at?

nichtich:
People no familiar with Rust may prefer Docker to try out oxygraph, so existence of Docker image should be mentioned.

Closes #1165.


Tpt (approved):
Thank you!

Issue #1156, opened 2025-01-19
Implement `Serialize` and `Deserialize` traits on `Term`s


jeswr:
**Is your feature request related to a problem? Please describe.**
Per the [rust API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#data-structures-implement-serdes-serialize-deserialize-c-serde) it is best practice to implement `serde::serialize` and `serde::deserialise` on each struct in the library.

I would like this feature to be able pass `Quad`s between guest and host in risc0.

**Describe the solution you'd like**
I would like for serialisation and deserialisation to be inexpensive. Thus for terms to be represented in object form; such as below, which mitigates the needs for doing regex on iri's etc.:

```
{
  "termType": "iri",
  "value": "http://example.org"
}
```

**Describe alternatives you've considered**
The serialisation could be as the n-quads representation of the term. Whilst more compact, this will be more expensive to parse.

**Additional context**
I am willing to implement. I would like:
 - Confirmation that `serde` can be added as a dependency to `oxrdf`
 - Confirmation that the object form for serialisation is acceptable

Tpt:
Hey! Thank you for raising this.

>  The serialisation could be as the n-quads representation of the term. Whilst more compact, this will be more expensive to parse.

Indeed, and it's already supported using `to_string` and `from_str`. However, I am not fully sure it will be significantly more expensive if we properly validate the data on deserialization. Checking that the IRI/blank node id/language tag are valid is the most expensive part of NQuads parsing and we might want to keep these checks in serde serialization.

However I am not against having `serde` support with two constraints:

- it should be behind a `serde` feature
- we should adopt a well known format. I see two options:
   1. JSON-LD
   2. SPARQL results JSON format

  JSON-LD is very complex so I am slightly more in favor of option 2.

PR #1172, opened 2025-02-02
feat: implement serde for terms

jeswr:
Closes https://github.com/oxigraph/oxigraph/issues/1156

Draft as I am working out how to mark serde as optional

### DIFF
--- a/lib/oxrdf/Cargo.toml
+++ b/lib/oxrdf/Cargo.toml
@@ -16,16 +16,17 @@ rust-version.workspace = true
 [features]
 default = []
 rdf-star = []
-serde = []
+serde = ["dep:serde", "dep:serde_json"]
 
 [dependencies]
 oxilangtag.workspace = true
 oxiri.workspace = true
 oxsdatatypes = { workspace = true, optional = true }
 rand.workspace = true
-serde = { version = "1.0.217", features = ["derive"] }
-serde_json = "1.0.138"
 thiserror.workspace = true
+serde = { version = "1.0.217", optional = true, features = ["derive"] }
+serde_json = { version = "1.0.138", optional = true }
+
 
 [lints]
 workspace = true
--- a/lib/oxrdf/Cargo.toml
+++ b/lib/oxrdf/Cargo.toml
@@ -16,6 +16,7 @@ rust-version.workspace = true
 [features]
 default = []
 rdf-star = []
+serde = []
 
 [dependencies]
 oxilangtag.workspace = true
--- a/lib/oxrdf/Cargo.toml
+++ b/lib/oxrdf/Cargo.toml
@@ -22,6 +22,8 @@ oxilangtag.workspace = true
 oxiri.workspace = true
 oxsdatatypes = { workspace = true, optional = true }
 rand.workspace = true
+serde = { version = "1.0.217", features = ["derive"] }
+serde_json = "1.0.138"
 thiserror.workspace = true
 
 [lints]
--- a/lib/oxrdf/Cargo.toml
+++ b/lib/oxrdf/Cargo.toml
@@ -27,7 +27,6 @@ thiserror.workspace = true
 serde = { version = "1.0.217", optional = true, features = ["derive"] }
 serde_json = { version = "1.0.138", optional = true }
 
-
 [lints]
 workspace = true
 
--- a/lib/oxrdf/src/blank_node.rs
+++ b/lib/oxrdf/src/blank_node.rs
@@ -377,6 +377,8 @@ pub struct BlankNodeIdParseError;
 #[cfg(test)]
 #[allow(clippy::panic_in_result_fn)]
 mod tests {
+    use crate::Term;
+
     use super::*;
     #[cfg(not(target_family = "wasm"))]
     use std::mem::{align_of, size_of};
@@ -455,12 +457,12 @@ mod tests {
         assert_eq!(b2, b);
     }
 
-    // #[test]
-    // fn test_serde_term() {
-    //     let b: Term = BlankNode::new("foo").unwrap().into();
-    //     let json = serde_json::to_string(&b).unwrap();
-    //     assert_eq!(json, "\"foo\"");
-    //     let b2: BlankNode = serde_json::from_str(&json).unwrap();
-    //     assert_eq!(b2, BlankNode::new("foo").unwrap());
-    // }
+    #[test]
+    fn test_serde_term() {
+        let b: Term = BlankNode::new("foo").unwrap().into();
+        let json = serde_json::to_string(&b).unwrap();
+        assert_eq!(json, "{\"type\":\"bnode\",\"value\":\"foo\"}");
+        let b2: BlankNode = serde_json::from_str(&json).unwrap();
+        assert_eq!(b2, BlankNode::new("foo").unwrap());
+    }
 }
--- a/lib/oxrdf/src/blank_node.rs
+++ b/lib/oxrdf/src/blank_node.rs
@@ -453,9 +453,14 @@ mod tests {
         assert_eq!(json, "{\"value\":\"42\"}");
         let b2: BlankNode = serde_json::from_str(&json).unwrap();
 
-        // WARNING: this a lossy process as blank nodes will always
-        // be reconstructed as named
         assert_eq!(b2, BlankNode::new("42").unwrap());
+
+        match b2 {
+            BlankNode(BlankNodeContent::Anonymous(id)) => {
+                assert_eq!(id, "42");
+            }
+            _ => panic!("Expected anonymous blank node"),
+        }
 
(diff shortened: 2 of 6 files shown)


lib/oxrdf/src/blank_node.rs
```diff
@@ -414,4 +444,33 @@ mod tests {
...
+    #[cfg(feature = "serde")]
+    fn test_serde() {
+        let b = BlankNode::new_from_unique_id(0x42);
+        let json = serde_json::to_string(&b).unwrap();
+        assert_eq!(json, "{\"value\":\"42\"}");
+        let b2: BlankNode = serde_json::from_str(&json).unwrap();
+
+        // WARNING: this a lossy process as blank nodes will always
+        // be reconstructed as named
+        assert_eq!(b2, BlankNode::new("42").unwrap());
```
jeswr, line 458: FYI

jeswr:
@Tpt I think the failing CI is unrelated to changes in this PR. Please ping me if otherwise and I will investigate.

lib/oxrdf/src/blank_node.rs
```diff
@@ -121,6 +125,29 @@ impl Default for BlankNode {
     }
 }
 
+#[cfg(feature = "serde")]
+impl Serialize for BlankNode {
+    #[inline]
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        json!({ "value": self.as_str() }).serialize(serializer)
```
Tpt, line 132: I don't think you need serde_json for this. Either you can create some struct 
```rust
#[derive(Serialize)]
struct NamedNodeForSerialization {
   r#type: &str,
   value: &str
}
```
or use directly the `Serializer` API:
```rust
        let mut state = serializer.serialize_struct("NamedNode", 2)?;
        state.serialize_field("type", "bnode")?;
        state.serialize_field("value", &self.value)?;
        state.end()
```

lib/oxrdf/src/blank_node.rs
Tpt: Is this true? The blank node ID `"42"` should be properly parsed as an anonymous one.

lib/oxrdf/src/literal.rs
```diff
@@ -32,13 +38,78 @@ use std::fmt::Write;
...
+            deserialize_with = "deserialize_string_variant"
+        )
+    )]
     String(String),
-    LanguageTaggedString { value: String, language: String },
-    TypedLiteral { value: String, datatype: NamedNode },
+}
+
+#[cfg(feature = "serde")]
+fn serialize_string_variant<S>(s: &String, serializer: S) -> Result<S::Ok, S::Error>
```
Tpt, line 72: Implementing `Serialize` directly seems simpler to me.

lib/oxrdf/src/literal.rs
```diff
@@ -32,13 +38,78 @@ use std::fmt::Write;
...
+where
+    S: Serializer,
+{
+    // Create an object with a single key "value".
+    let mut map = serializer.serialize_map(Some(1))?;
```
Tpt, line 77: Imho this should be a struct and not a map

lib/oxrdf/src/triple.rs
```diff
@@ -780,17 +790,69 @@ impl<'a> From<TermRef<'a>> for Term {
...
+fn predicate_deserialize<'de, D>(deserializer: D) -> Result<NamedNode, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Term::deserialize(deserializer)
+        .and_then(|term| NamedNode::try_from(term).map_err(serde::de::Error::custom))
+}
+
+#[cfg(feature = "serde")]
+fn subject_serialise<S>(subject: &Subject, serializer: S) -> Result<S::Ok, S::Error>
```
Tpt, line 838: Is it needed? `Subject` properly implement (de)serialize

lib/oxrdf/Cargo.toml
```diff
@@ -16,13 +16,16 @@ rust-version.workspace = true
...
+serde = ["dep:serde", "dep:serde_json"]
 
 [dependencies]
 oxilangtag.workspace = true
 oxiri.workspace = true
 oxsdatatypes = { workspace = true, optional = true }
 rand.workspace = true
 thiserror.workspace = true
+serde = { version = "1.0.217", optional = true, features = ["derive"] }
+serde_json = { version = "1.0.138", optional = true }
```
Tpt, line 28: It would be great to avoid this dependency, at least on the non-test path

Tpt (commented):
Thank you for moving forward with this!

lib/oxrdf/src/literal.rs
jeswr: Not quite, it's a little more verbose as you have to implement `Serialize` and `Deserialize`. Looks something like the following (note following is copilot generated)


```

#[cfg(feature = "serde")]
impl Serialize for Literal {
    #[inline]
    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        let mut map = serializer.serialize_map(Some(2))?;
        match &self.0 {
            LiteralContent::String(value) => {
                map.serialize_entry("value", value)?;
                map.serialize_entry("datatype", &xsd::STRING)?;
            }
            LiteralContent::LanguageTaggedString { value, language } => {
                map.serialize_entry("value", value)?;
                map.serialize_entry("language", language)?;
            }
            LiteralContent::TypedLiteral { value, datatype } => {
                map.serialize_entry("value", value)?;
                map.serialize_entry("datatype", datatype)?;
            }
        }?;
        map.end()
    }
}

#[cfg(feature = "serde")]
impl <'de> Deserialize <'de> for Literal {
    #[inline]
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        #[derive(Deserialize)]
        #[serde(field_identifier, rename_all = "lowercase")]
        enum Field {
            Value,
            Datatype,
            Language,
        }

        struct LiteralVisitor;

        impl<'de> serde::de::Visitor<'de> for LiteralVisitor {
            type Value = Literal;

            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                formatter.write_str("struct Literal")
            }

            fn visit_map<A: serde::de::MapAccess<'de>>(self, mut map: A) -> Result<Self::Value, A::Error> {
                let mut value = None;
                let mut datatype = None;
                let mut language = None;
                while let Some(key) = map.next_key()? {
                    match key {
                        Field::Value => {
                            if value.is_some() {
                                return Err(serde::de::Error::duplicate_field("value"));
                            }
                            value = Some(map.next_value()?);
                        }
                        Field::Datatype => {
                            if datatype.is_some() {
                                return Err(serde::de::Error::duplicate_field("datatype"));
                            }
                            datatype = Some(map.next_value()?);
                        }
                        Field::Language => {
                            if language.is_some() {
                                return Err(serde::de::Error::duplicate_field("language"));
                            }
                            language = Some(map.next_value()?);
                        }
                    }
                }
                match (value, datatype, language) {
                    (Some(value), Some(datatype), None) => Ok(Literal::new_typed_literal(value, datatype)),
                    (Some(value), None, Some(language)) => Ok(Literal::new_language_tagged_literal(value, language)?),
                    _ => Err(serde::de::Error::missing_field("value")),
                }
            }
        }

        const FIELDS: &[&str] = &["value", "datatype", "language"];
        deserializer.deserialize_struct("Literal", FIELDS, LiteralVisitor)
    }
}
```

Tpt: For `Serialize` you don't need to set the datatype explicitely for `LiteralContent::String` case, `xsd:string` is the default.

For `Deserialize` here is a likely simpler idea (not tested):

```rust
#[cfg(feature = "serde")]
impl <'de> Deserialize <'de> for Literal {
    #[inline]
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        #[derive(Deserialize)]
        struct LiteralValue {
            value: String,
            #[serde(default)]
            datatype: Option<String>,
            #[serde(default)]
            language: Option<String>,
        }
        let fields = LiteralValue::deserialize(deserializer)?;
        // Build a value from fields
    }
}
```

lib/oxrdf/Cargo.toml
jeswr: `serde_json` is indeed only used by tests; so it could be a `dev-dependency`. However, cargo does not support optional dev-dependencies (https://github.com/rust-lang/cargo/issues/1596). So the options are:
 - Make it a non-optional dev-dependency
 - Make it an optional dependency
 - Create a new package for testing purposes only which uses `serde_json`
 
 A new package might be the way to go, this is also required [should tests for the zkvm environment need to be added](https://github.com/rust-lang/rust/issues/135376#issuecomment-2657851915).

Issue #1181, opened 2025-02-13
`cargo test --all-features` fails on MacOS Sequia 15.3.0


jeswr:
**Describe the bug**

Test fails with error:

```
error: could not compile `oxigraph-testsuite` (bin "oxigraph-testsuite") due to 1 previous error
error: linking with `cc` failed: exit status: 1

[...]

note: ld: warning: object file (/Users/jesght/Documents/GitHub/jeswr/o3/oxigraph/target/debug/deps/liboxrocksdb_sys-2f0b6a54d3d172d2.rlib[4](ca3326909c0282bf-c.o)) was built for newer 'macOS' version (15.2) than being linked (11.0)
          Undefined symbols for architecture arm64:
            "_custom_ox_now", referenced from:
                oxsdatatypes::date_time::since_unix_epoch::h793e217ce4a54fd5 in liboxsdatatypes-93d7ad516731fe35.rlib[19](oxsdatatypes-93d7ad516731fe35.3775vjnulrmcp6rvb1kzenpy7.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

**To Reproduce**
1. Clone main branch (commit https://github.com/oxigraph/oxigraph/commit/c1909aa1df098f01359f2f99078214665d962bd0)
2. Run the following commands
```
gh repo clone oxigraph/oxigraph
cd ./oxigraph
git submodule update --init --recursive
cargo test --all-features
```

Tpt:
This is expected, Oxigraph allows to customize the clock (used eg for the `NOW` function). This customization is useful for some exotic platforms. `--all-features` is not really supported.

PR #1182, opened 2025-02-13
chore: add windows and macos tests

jeswr:
Originally opened to identify if #1181 is an issue with my local setup.

Now adds missing macos test regardless.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -169,6 +169,7 @@ jobs:
           submodules: true
       - uses: ./.github/actions/setup-rust
       - run: cargo test
+      - run: cargo test --all-features
 
   test_linux_i686:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -258,16 +258,6 @@ jobs:
       - run: bash .github/workflows/install_rocksdb.sh
       - run: cargo test --tests --features oxrocksdb-sys/pkg-config
 
-  test_windows:
-    runs-on: windows-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          submodules: true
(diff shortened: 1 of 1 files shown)


jeswr:
This is ready for review @Tpt

.github/workflows/tests.yml
```diff
@@ -157,8 +157,12 @@ jobs:
...
       - run: cargo semver-checks check-release --exclude oxrocksdb-sys --exclude oxigraph-js --exclude pyoxigraph --exclude oxigraph-testsuite --exclude oxigraph-cli --exclude spareval
 
-  test_linux_x86_64:
-    runs-on: ubuntu-latest
+  test_os:
+    strategy:
+      matrix:
+        os-version: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os-version }}
+    name: Test on ${{ matrix.os-version }}
```
Tpt, line 165: nit: Not sure if the name is useful.

.github/workflows/tests.yml
```diff
@@ -157,8 +157,12 @@ jobs:
       - run: bash .github/workflows/install_rocksdb.sh
       - run: cargo semver-checks check-release --exclude oxrocksdb-sys --exclude oxigraph-js --exclude pyoxigraph --exclude oxigraph-testsuite --exclude oxigraph-cli --exclude spareval
 
-  test_linux_x86_64:
-    runs-on: ubuntu-latest
+  test_os:
```
Tpt, line 160: maybe just `test`?

Tpt (commented):
Thank you for that!

Something that would be great would be to remove the `test_windows` job that is redundant is with the introduced `windows` variant of the new test.

An other approach might just be to not introduce `test_os` but write a fixed `test_macos` only running on mac OS.

PR #1185, opened 2025-02-16
Serde/v2

jeswr:


lib/oxrdf/src/triple.rs
```diff
@@ -780,17 +793,86 @@ impl<'a> From<TermRef<'a>> for Term {
...
     pub subject: Subject,
 
     /// The [predicate](https://www.w3.org/TR/rdf11-concepts/#dfn-predicate) of this triple.
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            serialize_with = "serialize_predicate",
+            deserialize_with = "deserialize_predicate"
+        )
+    )]
```
jeswr, line 808: Custom serialisation is needed as `NamedNode` alone serializes as `{ "value": "..." }` without the type key.

lib/oxrdf/Cargo.toml
```diff
@@ -16,13 +16,21 @@ rust-version.workspace = true
...
 oxsdatatypes = { workspace = true, optional = true }
 rand.workspace = true
 thiserror.workspace = true
+serde = { version = "1.0.217", features = ["derive"] }
+
+[dev-dependencies]
+# Ideally this would be an optional dependency, used only when the `serde` feature is enabled.
+# However, optional dev-dependencies are not currently supported by cargo.
+# See https://github.com/rust-lang/cargo/issues/1596
+serde_json = { version = "1.0.138" }
```
jeswr, line 33: For further info see https://github.com/oxigraph/oxigraph/pull/1172#discussion_r1957361275

.github/workflows/tests.yml
```diff
@@ -166,6 +166,19 @@ jobs:
...
+      matrix:
+        features: [ "serde", "rdf-star", "serde rdf-star" ]
+        os: [ "linux", "windows", "macos" ]
+    runs-on: ${{ matrix.os }}-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: ./.github/actions/setup-rust
+      - run: cargo test --features ${{ matrix.features }}
```
Tpt, line 180: This sounds a bit overkill. I believe the current tests already cover what we care about

lib/oxrdf/src/named_node.rs
```diff
@@ -14,8 +16,10 @@ use std::fmt;
 /// );
 /// # Result::<_,oxrdf::IriParseError>::Ok(())
 /// ```
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Hash)]
 pub struct NamedNode {
+    #[cfg_attr(feature = "serde", serde(rename = "value"))]
```
Tpt, line 22: would be nice to implement `Serialize` by hand to validate the IRI just like we do for blank nodes

lib/oxrdf/src/literal.rs
```diff
@@ -159,6 +164,100 @@ impl fmt::Display for Literal {
...
+        };
+        map.end()
+    }
+}
+
+#[cfg(feature = "serde")]
+struct LiteralVisitor;
+
+#[cfg(feature = "serde")]
+impl<'de> Visitor<'de> for LiteralVisitor {
```
Tpt, line 196: instead of a visitor what about deserializing into
```rust
#[derive(Deserialize)]
struct LiteralValue {
    value: String,
    language: Option<String>,
    datatype: Option<String>
}
```
then doing the conversion. It would make code much simpler I think (same for BlankNode and NamdeNode)

lib/oxrdf/src/triple.rs
```diff
@@ -780,17 +793,86 @@ impl<'a> From<TermRef<'a>> for Term {
...
+    state.serialize_field("value", &node.as_str())?;
+    state.end()
+}
+
+#[cfg(feature = "serde")]
+fn deserialize_predicate<'de, D>(deserializer: D) -> Result<NamedNode, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    struct PredicateVisitor;
```
Tpt, line 832: what about deserializing into a `Term` then failing if the term is not a `NamedNode`? Will be shorter and I think as correct.

Tpt (commented):
Thank you!

lib/oxrdf/src/literal.rs
jeswr: My main concern is that this would likely have a slower performance and higher memory consumption due to the need to build an extra structure.

Both big assumptions since I'm not familiar w. rust internals. Is there a particular toolset you typically use to benchmark performance and memory consumption that I could use to test this?

lib/oxrdf/src/triple.rs
jeswr: Similarly to https://github.com/oxigraph/oxigraph/pull/1185#discussion_r1960496168, only concern would be performance-related.

lib/oxrdf/src/literal.rs
Tpt: This would only build an extra `struct` so no memory allocation (we already pay the price of allocating the three `String`. My guess is that, if there is a slow down, it would be negligible and it can even lead to a speed up if code generated by `derive(Deserialize)` is faster than the hand-written one. A great tool for this kind of benchmark is https://github.com/bheisler/criterion.rs

jeswr: Amazing - thanks for the quick and informative response! Will submit changes shortly.

lib/oxrdf/src/named_node.rs
Tpt: This is a good point. I do not have a stong opinion but if we do not validate we should also skip validation of literal language tags and blank node ids for consistency.

I agree that for your use case validation is not useful, the trick here is that `oxrdf` has quite a bit of users and that I am afraid not doing validation might be a footgun for other users using the serde API to transmit data over the network or save them on disk. For example serde_json validates strings carefully so it feel a bit weird not validating the IRIs/language tags/...

adamreichold: Just as a data point informing the discussion, the `url` has opted to support the avoid-validation-overhead-during-serialization use case via a separate method called [`deserialize_internal`](https://docs.rs/url/latest/url/struct.Url.html#method.deserialize_internal). It is a bit more painful to compose, e.g. via Serde's [`deserialize_with`](https://serde.rs/field-attrs.html#deserialize_with) attributes, but it avoids the non-local reason implied by using Cargo features.

lib/oxrdf/src/literal.rs
jeswr: Hah, you're right!

```
serialize_named_node    time:   [40.539 ns 40.582 ns 40.626 ns]
                        change: [-3.8011% -1.8894% -0.5345%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  9 (9.00%) high mild
  2 (2.00%) high severe

deserialize_named_node  time:   [43.713 ns 43.777 ns 43.856 ns]
                        change: [+0.1525% +0.4638% +0.7617%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe

serialize_blank_node    time:   [37.631 ns 38.198 ns 39.303 ns]
                        change: [+0.5358% +1.3853% +2.6152%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 18 outliers among 100 measurements (18.00%)
  7 (7.00%) low mild
  2 (2.00%) high mild
  9 (9.00%) high severe

deserialize_blank_node  time:   [63.743 ns 63.832 ns 63.950 ns]
                        change: [-0.2541% -0.0384% +0.1985%] (p = 0.75 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe

serialize_literal       time:   [34.753 ns 34.782 ns 34.821 ns]
                        change: [-0.0387% +0.2234% +0.4904%] (p = 0.10 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe

deserialize_literal     time:   [41.395 ns 41.435 ns 41.481 ns]
                        change: [-7.5914% -3.7763% -1.4055%] (p = 0.01 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  7 (7.00%) high severe

serialize_typed_literal time:   [56.120 ns 56.190 ns 56.284 ns]
                        change: [-0.3061% -0.0868% +0.1326%] (p = 0.46 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe

deserialize_typed_literal
                        time:   [104.43 ns 106.71 ns 109.76 ns]
                        change: [-5.3163% -3.8312% -2.3816%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

serialize_triple        time:   [223.03 ns 224.54 ns 226.07 ns]
                        change: [-4.2299% -2.7129% -1.6589%] (p = 0.00 < 0.05)
                        Performance has improved.

deserialize_triple      time:   [450.54 ns 455.21 ns 460.74 ns]
                        change: [-7.6437% -5.1711% -2.2650%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe

File: lib/oxrdf/src/triple.rs (hunk 6 above)
Comment (jeswr) [reply lib/oxrdf/src/triple.rs:832]: There appears to be a non-negligible performance regression when doing this

```
deserialize_triple      time:   [512.28 ns 513.97 ns 515.64 ns]
                        change: [+9.7608% +12.148% +14.066%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

jeswr: However, having run multiple tests on my machine - approximately half of that regression may be noise

.github/workflows/tests.yml
jeswr: I've removed the multiple os' and cut out the `rdf-star` combos. I would like to make sure that both `serde` and `serde-unvalidated` are tested.

lib/oxrdf/Cargo.toml
jeswr: I'm not sure how to resolve the debian compatibilty error that this still seems to be causing

lib/oxrdf/src/named_node.rs
jeswr: I've gone with the `serde-unvalidated` feature flag approach.

A custom `deserialize_internal` method won't work for the use-case I am working on as the virtual machine is responsible for calling the serialization and deserialization methods.

adamreichold: > A custom deserialize_internal method won't work for the use-case I am working on as the virtual machine is responsible for calling the serialization and deserialization methods.

That appears like a shortcut IMHO. That VM just needs to use the right Serde annotations to request serialization without validation without affecting anything else in the dependency tree.

> I've gone with the serde-unvalidated feature flag approach.

The valence of this feature is also problematic. Cargo features are supposed to be additive and add functionality, not remove something like validation. The reason is that a feature can be enabled anywhere in the dependency tree of a crate. So one crate enabling this feature might remove validation for another crate expecting it thereby introducing unsoundness there. This appears especially problematic for something that might be used to process external input.

jeswr: I'm seeing what the options on the risc0 side are in https://discord.com/channels/953703904086994974/1342620330052550656

PR #1211, opened 2025-03-11
Greghanson/hdt spareval

GregHanson:
Follow-up PR based on discussion from https://github.com/oxigraph/oxigraph/pull/1087

Utilizes the stand-alone SPARQL evaluator for the HDT SPARQL implementation


GregHanson:
@Tpt sorry for the delay on this one - I am curious to hear if this implementation matches with what you envisioned.

> the HDT implementation can just return the internal integer encoding of the term and fetch the term value only for expression evaluation and returning the results.

I got a first pass working pretty quick, but I struggled a bit with optimizing the encoding/decoding portion and ultimately just used Strings. While @KonradHoeffner 's HDT library does have a concept of an internal integer ID for terms ([link](https://github.com/KonradHoeffner/hdt/blob/main/src/four_sect_dict.rs#L88)) it is dependent on knowing whether the requested Term is a subject/predicate/object, a prerequisite that I had a hard time determining in the relevant encode/decode functions for the QueryableDataset trait. Also this type of lookup gets messier the more HDT datasets are included

Tpt:
It's great progress, thank you!

On the implementation I think you can avoid using `EncodedTerm` completely: you implement a conversion function between `oxrdf:Term` and `String` in both directions and let the default implementation of methods in `QueryableDataset` handle `ExpressionTerm` (no need to deal with it directly if you don't want to).
It means you only have to implement the [internal_quads_for_pattern](https://docs.rs/spareval/latest/spareval/trait.QueryableDataset.html#tymethod.internal_quads_for_pattern), [internalize_term](https://docs.rs/spareval/latest/spareval/trait.QueryableDataset.html#tymethod.internalize_term) and [externalize_term](https://docs.rs/spareval/latest/spareval/trait.QueryableDataset.html#tymethod.externalize_term) methods.
 This way your implementation does not depend on the `oxigraph` crate at all and you can use a standalone crate for HDT querying that does not even need to be part of Oxigraph git repository.

> it is dependent on knowing whether the requested Term is a subject/predicate/object, a prerequisite that I had a hard time determining in the relevant encode/decode functions for the QueryableDataset trait. Also this type of lookup gets messier the more HDT datasets are included

Good point, I missed this issue. Thanks! Maybe there are workarounds but they do not look simple.

GregHanson:
Thanks for the advice! that does simplify things quite a bit

> This way your implementation does not depend on the oxigraph crate at all and you can use a standalone crate for HDT querying that does not even need to be part of Oxigraph git repository.

If I'm following this right, are you suggesting `evaluate_hdt_query()` and the `HDTDataset+QueryableDataset`  implementation just ends up in its own standalone crate? Interesting, I'll have to defer to @donpellegrino

donpellegrino:
> Thanks for the advice! that does simplify things quite a bit
> 
> > This way your implementation does not depend on the oxigraph crate at all and you can use a standalone crate for HDT querying that does not even need to be part of Oxigraph git repository.
> 
> If I'm following this right, are you suggesting `evaluate_hdt_query()` and the `HDTDataset+QueryableDataset` implementation just ends up in its own standalone crate? Interesting, I'll have to defer to @donpellegrino

Refactoring the ability to query HDT with SPARQL into a standalone crate could create a streamlined option for anyone needing only that capability. That seems like a reasonable way to go.

Ideally, the approach would still allow the use of HDT as read-only backend with equivalent architectural standing to other backends, such as RocksDB. With that approach, other Oxigraph capabilities such as the REST SPARQL Endpoint, would be available to users who wanted those additional features.

GregHanson:
> Ideally, the approach would still allow the use of HDT as read-only backend with equivalent architectural standing to other backends, such as RocksDB. With that approach, other Oxigraph capabilities such as the REST SPARQL Endpoint, would be available to users who wanted those additional features.

As it stands, this PR just adds the ability to perform SPARQL query over HDT files. It currently is not integrated with oxigraph CLI commands such as `serve` or `query`

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
 use std::cell::RefCell;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
+use std::io::{Error, ErrorKind};
 use std::iter::empty;
+use std::str::FromStr;
 use std::sync::Arc;
 
+pub struct HDTDataset {
```
Tpt, line 20: If you add `#[derive(Clone)]` here you can then also add it on `#[derive(Clone)]` and remove the custom implementation

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+    /// Path to the HDT file.
+    path: String,
```
Tpt, line 22: Do you still need the path?

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+/// Create the correct OxRDF term for a given resource string.  Slow,
+/// use the appropriate method if you know which type (Literal, URI,
+/// or blank node) the string has. Based on
+/// https://github.com/KonradHoeffner/hdt/blob/871db777db3220dc4874af022287975b31d72d3a/src/hdt_graph.rs#L64
+fn hdt_bgp_str_to_term(s: &str) -> Result<Term, Error> {
+    match s.chars().next() {
+        None => Err(Error::new(ErrorKind::InvalidData, "empty input")),
+
+        // Double-quote delimters are used around the string.
+        Some('"') => match s.rfind('"') {
```
Tpt, line 74: Is `rfind` useful? I would guess that `Term::from_str` would take care of the validation

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+                Err(e) => Err(Error::new(
+                    ErrorKind::InvalidData,
+                    format!("literal parse error {e} for {s}"),
+                )),
+            },
+        },
+
+        // Underscore prefix indicating an Blank Node.
+        Some('_') => {
+            let term = match oxrdf::BlankNode::new(&s[2..]) {
```
Tpt, line 91: An other option is `BlankNode::from_str(s)` if you prefer

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+                Ok(n) => n,
+                Err(e) => {
+                    return Err(Error::new(
+                        ErrorKind::InvalidData,
+                        format!("blanknode parse error {e} for {s}"),
+                    ))
+                }
+            };
+            match Term::from_str(&term.to_string()) {
```
Tpt, line 100: Using just `term.into()` should do the job, no need for `Term::from_str(&term.to_string())` that does a serialization roundtrip (same for literals)

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+            }
+        }
+
+        // Double-quote delimiters not present. Underscore prefix
+        // not present. Assuming a URI.
+        _ => {
+            // Note that Term::from_str() will not work for URIs
+            // (OxRDF NamedNode) when the string is not within "<"
+            // and ">" delimiters.
+            let named_node = match NamedNodeRef::new(*Arc::from(s)) {
```
Tpt, line 115: `NamedNode::new(s)` should do the job and then allow to use `.into()` to convert into a `Term`

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+
+/// Convert triple string formats from OxRDF to HDT.
+fn term_to_hdt_bgp_str(term: &Term) -> Result<String, StorageError> {
+    let hdt_str = match term {
+        // Remove double quote delimiters from URIs.
+        Term::NamedNode(named_node) => named_node.clone().into_string(),
+
+        // Get the string directly from literals and add
+        // quotes to work-around handling of "\n" being
+        // double-escaped.
```
Tpt, line 143: The `\n` unescaping is scary because oxrdf escape other characters (", \r...) and unescape them in `Literal::from_str`

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+                    vec![Err(StorageError::Corruption(CorruptionError::new(
+                        format!("HDT does not support named graph: {graph_name:?}"),
+                    )))]
+                    .into_iter(),
+                );
+            }
+        }
+        let s = subject.cloned();
+        let p = predicate.cloned();
+        let o = object.cloned();
```
Tpt, line 191: Genuine question: do you need these clones?

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+                    graph_name: None,
+                }));
+            }
+        }
+
+        Box::new(v.into_iter())
+    }
+
+    fn internalize_term(&self, term: Term) -> Result<String, StorageError> {
+        term_to_hdt_bgp_str(&term)
```
Tpt, line 222: I think you can convert `term_to_hdt_bgp_str` to take an owned `Term` and not a reference and avoid the `.clone()` call for NamedNodes.

lib/oxigraph/src/sparql/dataset.rs
```diff
@@ -1,18 +1,235 @@
...
+    }
+
+    fn externalize_term(&self, term: String) -> Result<Term, StorageError> {
+        match hdt_bgp_str_to_term(&term) {
```
Tpt, line 226: Same here, I think you can make `hdt_bgp_str_to_term` take an owned `String`

Tpt (commented):
Thank you for the improvements

> Ideally, the approach would still allow the use of HDT as read-only backend with equivalent architectural standing to other backends, such as RocksDB. With that approach, other Oxigraph capabilities such as the REST SPARQL Endpoint, would be available to users who wanted those additional features.

This is an interesting topic. My fear is that the API for the HDT querying is going to be very different from the API from RocksDB: one opens multiple files and not a single folder, no insertions/transactions/.... Hence, all layers on top (SPARQL query...) will have to be customized for HDT. A SPARQL query endpoint is likely around 100 LOC, duplicating it for HDT is maybe not very bad.

lib/oxigraph/src/sparql/dataset.rs
GregHanson: I'll have to test this out, perhaps it is no longer applicable with the new implementation.  But we hit some problems early on with the conversion and there was an issue when including the leading "_" chars.

lib/oxigraph/src/sparql/dataset.rs
Tpt: You can do `s.map(|s| s.as_str())` to convert from `Option<&String>` to `Option<&str>` without reallocating.

lib/oxigraph/src/sparql/dataset.rs
Tpt: Weird, `BlankNode::from_str` should work with `_:ad233` like identifiers

lib/oxigraph/src/sparql/dataset.rs
Tpt: Yes, indeed.
It seems hdt-cpp is missing some escapes that are required for proper NTriples. For example data imported from Turtle are not escaped at all: https://github.com/rdfhdt/hdt-cpp/blob/d9ae092bb37d9fe85558dfb3edfe0bb6ddddf41a/libhdt/src/rdf/RDFParserSerd.cpp#L46

This might cause weird bugs for string containing eg. `"`.

Tpt (commented):
Thanks

lib/oxigraph/src/sparql/dataset.rs
adamreichold: Or shorter yet, `s.as_deref()`, c.f. [`Option::as_deref`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.as_deref).

GregHanson:
Sorry for the lack of updates here. I haven't been too impressed with responsiveness and release cadence of the [hdt-cpp](https://github.com/rdfhdt/hdt-cpp) library, so I just decided to re-write it in Rust. @KonradHoeffner's HDT crate only supports a subset of that library's HDT implementation so it wasn't as much work as initially thought. Utilizing oxrdfio utils, I think the rust library handles some of the edgecases/bugs we found using the CPP version. The Rust library can be found [here](https://github.com/DeciSym/rdf2hdt) for the time being and I'm working with Konrad to determine which crate that conversion logic will end up residing in, whether in hdt_rs or its own crate.

I'm testing my rust rdf2hdt implementation when I can so I can feel a little more confident about the dataset implementation I have in this PR

PR #1213, opened 2025-03-15
feat: add example of running query exaplanation

jeswr:
This PR:
 - Creates an `examples` directory and adds an example of running a query with explanation
 - Adds a script and CI to make sure that all toml files have newlines since this was not caught by `cargo fmt`.
 - Configures `.cargo/cargo.toml` for better dev performance/caching.

Running jointly as an experiment to test the capabilities of [cursor](https://www.cursor.com) ... which is quite impressive running with sonnet-3.7-thinking ... and so that I can see what is produced in an explanation. Figured this is a useful artefact that was produced in the process so may as well contribute it.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -2,6 +2,7 @@
 name = "oxigraph-examples"
 version = "0.1.0"
 edition = "2021"
+license = "MIT OR Apache-2.0"
 publish = false
 
 [[example]]


Tpt:
Thank you for that. Cargo has some built-in support for examples so if you move the examples to the `lib/oxigraph/examples` directory you can then run an example by doing `cargo run --package oxigraph --example foo` to run the example `foo.rs`.

.cargo/config.toml
Tpt, line 1: I am not sure we want to commit this file, it's something that should be up to each developper imho

rustfmt.toml
```diff
@@ -8,3 +8,4 @@ normalize_doc_attributes = true
 group_imports = "One"
 use_field_init_shorthand = true
 use_try_shorthand = true
+final_newline = true
```
Tpt, line 11: I don't find any occurence of this configuration variable in rustfmt. Is this addition intended?

PR #1272, opened 2025-05-16
feat: serde no unvalidated

jeswr:
Implements #1185 without the option to skip validation.

Once this is merged I suggest implement the approach in https://github.com/oxigraph/oxigraph/pull/1185#discussion_r1960670744


lib/oxrdf/src/blank_node.rs
```diff
@@ -121,6 +126,49 @@ impl Default for BlankNode {
...
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for BlankNode {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_struct("BlankNode", &["value"], BlankNodeVisitor)
```
Tpt, line 168: Can we make it much simpler with something like (not sure about the reference)?
```rust
#[derive(Deserialize)]
struct BNode<'a> {
    value: &'a str
}
let bnode = Bnode::deserialize(deserializer)?;
Self::new(bnode.value).map_err(de::Error::custom)
```
(same for IRIs)

lib/oxrdf/src/blank_node.rs
```diff
@@ -414,4 +466,64 @@ mod tests {
...
+        let b2: BlankNode = serde_json::from_str(&json).unwrap();
+
+        assert_eq!(b2, BlankNode::new("42").unwrap());
+
+        match b2 {
+            BlankNode(BlankNodeContent::Anonymous { str: _, id }) => {
+                assert_eq!(id, 0x42u128.to_ne_bytes());
+            }
+            _ => panic!("Expected anonymous blank node"),
+        }
```
Tpt, line 485: Do we need that? I think this is already covered by the previous assertion `assert_eq!(b2, BlankNode::new("42").unwrap());`

lib/oxrdf/src/blank_node.rs
```diff
@@ -414,4 +466,64 @@ mod tests {
...
+        assert!(b3.is_err());
+
+        let b4: Result<BlankNode, serde_json::Error> =
+            serde_json::from_str::<BlankNode>(&"{\"art\":\"r\"}");
+        assert!(b4.is_err());
+    }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn as_str_partial_reader() {
```
Tpt, line 504: Is this test useful? It looks like a repetition of the previous one

lib/oxrdf/src/literal.rs
```diff
@@ -159,6 +161,63 @@ impl fmt::Display for Literal {
...
 }
 
+#[cfg(feature = "serde")]
+impl Serialize for Literal {
+    #[inline]
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut map: <S as Serializer>::SerializeStruct;
+        match self {
+            Self(LiteralContent::String(value)) => {
+                map = serializer.serialize_struct("Literal", 1)?;
```
Tpt, line 171: Nit: I think you can factorize this line

lib/oxrdf/src/literal.rs
```diff
@@ -159,6 +161,63 @@ impl fmt::Display for Literal {
...
+        } = LiteralValue::deserialize(deserializer)?;
+        match (datatype, language) {
+            (Some(datatype), None) => Ok(Literal::new_typed_literal(
+                value,
+                NamedNode::new(datatype).map_err(de::Error::custom)?,
+            )),
+            (None, Some(language)) => {
+                Literal::new_language_tagged_literal(value, language).map_err(de::Error::custom)
+            }
+            _ => Ok(Literal::new_simple_literal(value)),
```
Tpt, line 216: It would be nice imho to fail on `(Some(_), Some(_))`

lib/oxrdf/src/literal.rs
```diff
@@ -159,6 +161,63 @@ impl fmt::Display for Literal {
...
+    }
+}
+
+#[cfg(feature = "serde")]
+#[derive(Deserialize)]
+struct LiteralValue {
+    value: String,
+    language: Option<String>,
+    datatype: Option<String>,
+}
```
Tpt, line 195: hyper-nit: you can move this struct inline in the `deserialize` function to make clear it's only a temporary helper

lib/oxrdf/src/literal.rs
```diff
@@ -665,4 +726,102 @@ mod tests {
...
+        let node: Literal = Deserialize::deserialize(&mut de).unwrap();
+        assert_eq!(
+            node,
+            Literal::new_language_tagged_literal("foo", "en").unwrap()
+        );
+    }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn test_serde_from_reader() {
```
Tpt, line 767: Isn't it a given from Serde?

bench/criterion/Cargo.toml
Tpt, line 1: Thank you for writing these benchmark, the implementation is fairly small, I am not sure it's really worth it to maintain it in the repository.

Tpt (commented):
Thank you for the improvements!

Some nits around simplifications

lib/oxrdf/src/literal.rs
jeswr: Not easily because this has `1` field and the others have `2`.

bench/criterion/Cargo.toml
jeswr: Check - removed in https://github.com/oxigraph/oxigraph/pull/1272/commits/4bf0b0aff699073f93b7a8ee80d7e68de064a9af

lib/oxrdf/src/blank_node.rs
jeswr: Because it uses a `Reader` to deserialise - and `Reader`s require implementors to not borrow strings. See  https://github.com/oxigraph/oxigraph/pull/1272#discussion_r2106251075 for more detail.

lib/oxrdf/src/blank_node.rs
adamreichold: > the implementation will no longer error if unexpected properties are present in the object being deserialised

You should be able to avoid this by adding the [`deny_unknown_fields`](https://serde.rs/container-attrs.html#deny_unknown_fields) to the intermediate struct.

jeswr: Fine w. me - lets leave as is.

lib/oxrdf/src/literal.rs
jeswr: Marking as resolved since we've had the `Deserialise` vs. `DeserialiseOwned` conversation in the other thread.

lib/oxrdf/src/blank_node.rs
jeswr: Removed in https://github.com/oxigraph/oxigraph/pull/1272/commits/6b7373010502a2b9d3b423e44ea1b391fa4ccdb7

lib/oxrdf/src/blank_node.rs
```diff
@@ -414,4 +447,58 @@ mod tests {
...
+        assert_eq!(b2, b);
+    }
+
+    // This test is required to make sure that we are not borrowing any strings
+    // when deserializing a BlankNode.
+    #[test]
+    #[cfg(feature = "serde")]
+    fn as_str_partial_reader() {
+        let j = serde_json::to_string(&BlankNode::new("abc").unwrap()).unwrap();
+        let reader = std::io::Cursor::new(j.into_bytes());
```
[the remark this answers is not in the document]
jeswr: `into_bytes` returns `Vec<u8>` which does not implement read.

I tried to use `try_into` to cast from `&Vec<u8>` to `&[u8]` but it gives the error:
```bash
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> lib/oxrdf/src/blank_node.rs:481:45
    |
481 |         let reader: &[u8] = &j.into_bytes().try_into().unwrap();
    |                                             ^^^^^^^^ doesn't have a size known at compile-time
```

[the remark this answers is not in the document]
Tpt: Or just string::as_bytes()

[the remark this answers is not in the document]
jeswr: `as_slice` works - thanks!

https://github.com/oxigraph/oxigraph/pull/1272/commits/05a00e03e3084da8a5462463ea44b80713b88e7b

PR #1275, opened 2025-05-17
Feat/serde query

jeswr:
Extends #1272 with support for `Serde` on the query object.


jeswr:
This is ready for review @Tpt .

If you agree with the approach I would request this be merged _in favour of_ #1272

PR #1276, opened 2025-05-17
feat: borsh

jeswr:
Extends #1275 and requires https://github.com/oxigraph/oxiri/pull/52 to be released

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -172,6 +172,19 @@ jobs:
       - uses: ./.github/actions/setup-rust
       - run: cargo test
 
+  test_features:
+    strategy:
+      matrix:
+        features: [ "serde" ]
+        os: [ "linux", "windows", "macos" ]
+    runs-on: ${{ matrix.os }}-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: ./.github/actions/setup-rust
+      - run: cargo test --features ${{ matrix.features }}
+
   test_linux_i686:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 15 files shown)


Tpt (commented):
Thank you for this! Sadly I don't feel very comfortable to add support for a non-Serde serialization system, there are so many of them. External implementations seems safer to me.

jeswr:
Understood - that's very reasonable. The reason for choosing is because that is what [risc0](https://github.com/risc0/risc0) seems to be evolving towards for passing things into the ZKVM.

I'm not sure how I would go about implementing this externally; I tried to do the following in another crate - but was hit with `error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate`.

```rs
impl BorshSerialize for Iri<String> {
    ...
}
```

Tpt:
> I'm not sure how I would go about implementing this externally

Sadly, it's a bit painful. You need to do something like:

```rust
#[derive(BorshSerialise, BorshDeserialize)]
struct BorshIri {
   iri: String
}

impl From<BorshIri> for Iri<String> {
   fn from(iri: BorshIri) -> Self {
       Self::parse_unchecked(iri)
   }
}


impl From<Iri<String>> for BorshIri {
   fn from(iri: Iri<String>) -> Self {
     Self { iri: iri.into_inner() }
   }
}
```
note that this way you can customize how data is serialized, which validation is done.
To optimise a bit you can also define something like:

```
#[derive(BorshSerialise)]
struct BorshIriRef<'a> {
   iri: &'a str
}


impl<'a> From<&'a Iri<String>> for BorshIri {
   fn from(iri: &'a Iri<String>) -> Self {
     Self { iri: iri.as_str() }
   }
}
```

and serialize references

jeswr:
Wonderful - thanks for the pointers!

PR #1284, opened 2025-05-22
feat: Add interactive CLI for SPARQL query and update

AhmedLukman:
The interactive mode allows you to run SPARQL queries and updates directly from a terminal or via programmatic integration (e.g., Elixir ports) using standard input and output.

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -262,4 +262,25 @@ pub enum Command {
         #[arg(long, value_hint = ValueHint::Url)]
         to_graph: Option<String>,
     },
+    /// Start an interactive CLI for SPARQL query and update
+    Interactive {
+        /// Directory in which Oxigraph data are persisted
+        #[arg(short, long, value_hint = ValueHint::DirPath)]
+        location: PathBuf,
+        /// Base IRI of the query (optional, used for queries)
+        #[arg(long, value_hint = ValueHint::Url)]
+        query_base: Option<String>,
+        /// The results format (optional, used for queries)
+        #[arg(long)]
+        results_format: Option<String>,
+        /// If the SPARQL queries should look for triples in all the dataset graphs by default (ie. without `GRAPH` operations)
+        #[arg(long)]
+        union_default_graph: bool,
+        /// Base IRI of the update (optional, used for updates)
+        #[arg(long, value_hint = ValueHint::Url)]
+        update_base: Option<String>,
+        /// Suppress all output except the response generated
+        #[arg(short, long)]
+        quiet: bool,
+    },
 }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -25,7 +25,7 @@ use std::cmp::{max, min};
 use std::env;
 use std::ffi::OsStr;
 use std::fs::File;
-use std::io::{self, stdin, stdout, BufWriter, Read, Write};
+use std::io::{self, stdin, stdout, BufRead, BufWriter, Read, Write};
 use std::net::ToSocketAddrs;
 #[cfg(target_os = "linux")]
 use std::os::unix::net::UnixDatagram;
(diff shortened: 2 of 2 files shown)


Tpt:
Thank you for pushing this! Quick question before moving forward, would starting a new `oxigraph query` or `oxigraph update` process for each SPARQL query/update be fast enough? Spawning a new process might not be a non starter and it would have the advantage of allowing multiple `oxigraph query` in parallel.

amiramix:
> Thank you for pushing this! Quick question before moving forward, would starting a new `oxigraph query` or `oxigraph update` process for each SPARQL query/update be fast enough? Spawning a new process might not be a non starter and it would have the advantage of allowing multiple `oxigraph query` in parallel.

Thank you for your prompt review.

Can you actually run one query and one update in parallel against the same database? I would guess that 'update' creates a lock and wouldn't allow for the query to complete?

If that's so, spawning a new process for each query would require serializing updates and synchronizing multiple queries before issuing a new update.

On top of that comes the process management to make sure that each process exits gracefully before starting a new one.

We need something very simple, a FIFO queue of updates and queries against the same database without a complicated process management. Something akin to opening a Word when editing a document. You open the document, read it, update, read, update, until it's done, then close.

We are planning to interface it with Elixir through OTP ports and are more interested in orchestrating a large number of long-lived processes able to issue updates and queries against one database each rather than having one database for everything. Please let us know if you see any problems with such approach.

Tpt:
Thank you for your detailed reply!

I see that having to figure out open port is not a fun thing. Would adding support of UNIX sockets help?

>  HTTP: 3 processes, one for BEAM, one for CLI management, one for Oxigraph

I am a bit confused by this. You mean you can't spawn the Oxigraph HTTP server directly from Elexir.

> Just to be clear, if you don't think this is a useful addition we can maintain it as a fork. Or we can also create a separate lighter cli command that could be used for this type of communication.

I am indeed not very comfortable to add this feature in this current shape. Writing a lighter CLI is likely the easiest way to go, it's likely you will be able to remove 95% of the current code and won't have to maintain a full fork. The Rust API of Oxigraph is fairly stable so it should not be a huge burden.

GregJunkaTR:
> > HTTP: 3 processes, one for BEAM, one for CLI management, one for Oxigraph
> 
> I am a bit confused by this. You mean you can't spawn the Oxigraph HTTP server directly from Elexir.

I might have gotten it wrong. You are right, It would still be 2 processes, one for Elixir and one for oxigraph. But regardless, we would still need to start the process and monitor it using ports then issue the actual queries using HTTP/S. So, in any case, more complex than just using ports throughout.

Seems that Unix sockets would require a similar approach to HTTP/S, i.e. start the application through ports then communicate with it through Unix sockets.

I hope you see the issue. Ports are able to start and monitor an external command line application, and also communicate with it through file descriptors. Unix sockets or HTTP/S solve only the communication part, not starting, stopping or monitoring for exits.

Tpt:
> we would still need to start the process and monitor it using ports then issue the actual queries using HTTP/S.

Yes, definitely. I have to confess I am not familiar enough with Erlang to get a concrete idea on the pain it might be.

> an alternative is to use different file descriptors than stdin/stdout and stderr. Would that make it more likely to be considered for merging?

Sadly not.

I tend to think that building a small bespoke CLI is the best way to go here, you will be able to control exactly how it will behave and it is likely to not be very large.

Thank you for this MR and this great discussion and sorry for having made you waste some time on it

amiramix:
> > we would still need to start the process and monitor it using ports then issue the actual queries using HTTP/S.
> 
> Yes, definitely. I have to confess I am not familiar enough with Erlang to get a concrete idea on the pain it might be.

That's why I wanted the feature to be useful in general case. The interactive mode allows you to work with the DB by querying and sending updates without exiting the process. Isn't that useful? If it wasn't they wouldn't have added the command line to node, Elixir, Python, Ruby, etc.

> > an alternative is to use different file descriptors than stdin/stdout and stderr. Would that make it more likely to be considered for merging?
> 
> Sadly not.
> 
> I tend to think that building a small bespoke CLI is the best way to go here, you will be able to control exactly how it will behave and it is likely to not be very large.
> 
> Thank you for this MR and this great discussion and sorry for having made you waste some time on it

We need the full functionality of the Query and Update commands. There isn't much more than that in the oxigraph-cli, is there?

Would you at least consider merging the refactoring of the query and update into separate functions? In that case we can only maintain the interactive loop separately.

Thanks for your prompt responses, I can't keep up :)

Tpt:
> Isn't that useful? If it wasn't they wouldn't have added the command line to node, Elixir, Python, Ruby, etc.

I would think that the equivalent is more the REPL of DuckDB/postgreSQL client... I feel we have not nailed the UX for it yet: having something both friendly to humans and friendly to software trying to call Oxigraph does not sound easy. It seems to me REPLs are mostly focused on humans and other software use specific protocols (postgres protocol...)

> We need the full functionality of the Query and Update commands. There isn't much more than that in the oxigraph-cli, is there?

There is a bunch of things around loading RDF data, handling HTTP... You don't need all of that.

> Would you at least consider merging the refactoring of the query and update into separate functions? In that case we can only maintain the interactive loop separately.

I can but I would nudge you to avoid depending on them but only depend on the `oxigraph` crate API. It's a more stable API.

PR #1285, opened 2025-05-23
add Clang and LLVM requirements to oxigraph CLI readme

KonradHoeffner:


cli/README.md
```diff
@@ -37,7 +37,7 @@ Note that Oxigraph CLI was previously named Oxigraph Server before version 0.4.
 
 ## Installation
 
-You need to have [a recent stable version of Rust and Cargo installed](https://www.rust-lang.org/tools/install).
+You need to have [a recent stable version of Rust and Cargo installed](https://www.rust-lang.org/tools/install) as well as Clang and LLVM for the RocksDB Rust bindings.
```
Tpt, line 40:
```suggestion
You need to have [a recent stable version of Rust and Cargo installed](https://www.rust-lang.org/tools/install) as well as Clang for the RocksDB Rust bindings.
```

Tpt (approved):
Thank you! I don't think having the LLVM libraries exposed needed, only clang compiler

cli/README.md
KonradHoeffner: @Tpt you are right! Contrary to what is [written in the RocksDB README](https://github.com/rust-rocksdb/rust-rocksdb?tab=readme-ov-file#requirements), Clang alone is enough.

PR #1314, opened 2025-06-15
feat: add serde support for query object

jeswr:
Supercedes #1275


Tpt:
Thank you for this! I am a bit reluctant to add serde to the query AST. This is mostly an Oxigraph implementation detail I allow myslef to change between versions. So, allowing people to (de)serialize it seems a possible footgun. What would be the use case for this that is not already covered by serializing again the query into SPARQL syntax?

Tpt:
Thank you for the flame graph, parsing does not seems to be the overwhelming bottleneck. Maybe taking some time to optimize the SPARQL parser (I never took time to do it outside of fixing some obvious bottlenecks) would reduce this share even further and have a great impact on every Oxigraph user? What do you think?

Tpt:
Stalled for a while. Making the SPARQL parser better is likely a nicer option here.

PR #1316, opened 2025-06-16
Introduce `LowLevelTurtleParser::seen_comment(&self) -> bool`

hoijui:
I am pretty sure you do not want that,
but I need this (or somethign similar),
and it woudl make my life easier if it were with you (upstream).
If you would accept it if it were behind a feature,
or even making comment content available
in a way similar to prefixes and bases (behind a feautre),
I could have a go at that too.

If none of that, I'll just have to maintain a fork for that;
possible too.

### DIFF
--- a/lib/oxttl/src/toolkit/lexer.rs
+++ b/lib/oxttl/src/toolkit/lexer.rs
@@ -64,6 +64,11 @@ pub struct Lexer<B, R: TokenRecognizer> {
     min_buffer_size: usize,
     max_buffer_size: usize,
     line_comment_start: Option<&'static [u8]>,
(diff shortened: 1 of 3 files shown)


Tpt:
Thank you! It makes a lot of sense.

>  I do not want to use a custom parser just for that, and I did not want to parse the source two times, even though that would probably not be a real issue in many practical scenarios.

Yes, if I am not mistaken, `#` is only allowed in comments or IRIs, so just looking at the `<`, `#`, `>` characters (maybe using `memchr::memchr3`) should be enough to detect comments, i.e. if there exist a `#` character that is not enclosed in `<` and `>`.

PR #1403, opened 2025-08-21
feat: evaluate expression

### LINKED ISSUE #1402: [Feat]: Publicly expose a `effective_boolean_value_expression_evaluator` function
I would like to have access to an `effective_boolean_value_expression_evaluator` in another package. Would you accept a PR which:
 - Makes the below function public *within* the module but not outside
 - Implements a new function which made public outside the module; and which has signature `pub fn(expression: &Expression, encoded_variables: &mut Vec<Variable>) -> Optional<bool>`

https://github.com/oxigraph/oxigraph/blob/a8cb4b7812a4715e1c72b4d9d8be693d9d4736ac/lib/spareval/src/eval.rs#L1875-L1894

jeswr:
Closes #1402

### DIFF
--- a/lib/spareval/src/lib.rs
+++ b/lib/spareval/src/lib.rs
@@ -399,7 +399,7 @@ impl QueryEvaluator {
     fn eval_expression_term_with_substitutions<'a>(
         &self,
         expression: &sparopt::algebra::Expression,
-        substitutions: impl IntoIterator<Item = (&'a Variable, &'a Term)>,
+        substitutions: impl IntoIterator<Item = (&'a Variable, Term)>,
     ) -> Option<ExpressionTerm> {
         // Empty dataset to support EXISTS evaluation without accessing data
         let dataset = Dataset::new();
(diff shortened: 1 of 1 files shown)


lib/spareval/src/lib.rs
```diff
@@ -394,6 +394,65 @@ impl QueryEvaluator {
         self.run_stats = true;
         self
     }
+
+    // Internal helper: evaluates a SPARQL expression to an ExpressionTerm against an empty dataset
+    fn eval_expression_term_with_substitutions<'a>(
+        &self,
+        expression: &sparopt::algebra::Expression,
+        substitutions: impl IntoIterator<Item = (&'a Variable, &'a Term)>,
```
Tpt, line 402: nit: Maybe better with `Term` instead of `&Term` to avoid the clone() call (same with public APIs) but I am happy to merge the MR in the current state if you prefer.

jeswr: Implemented in https://github.com/oxigraph/oxigraph/pull/1403/commits/8b3e6eb4b99550a0b29139ae69a36450e79f2ebd

PR #1437, opened 2025-09-12
Hdt storage

GregHanson:
POC for adding a new storage implementation that can be integrated with Oxigraph's CLI


Tpt:
>Would you be open to accepting a PR adding HDT as a format to oxigraph's convert subcommand?

I would tend to prefer for it to stay in a separated project while HDT implementation matures.

> but it sounds like I'd be better off with something custom and using the individual QueryableDataset implementations

Yes!

PR #1439, opened 2025-09-14
Bulk loader: wait for background threads before cleaning up the state

Tpt:


lib/oxigraph/src/storage/rocksdb.rs
```diff
@@ -1340,6 +1340,12 @@ pub struct RocksDbStorageBulkLoader<'a> {
 
 impl Drop for RocksDbStorageBulkLoader<'_> {
     fn drop(&mut self) {
+        // We wait for threads
+        while let Some(thread) = self.threads.pop_front() {
+            if let Ok(Ok(files)) = thread.join() {
```
Tpt, line 1345: This is a terrible idea, I am working on a fix

PR #1443, opened 2025-09-15
bulk loaders: Add support for partial commits

progval:


lib/oxigraph/src/storage/mod.rs
```diff
@@ -602,6 +602,19 @@ impl StorageBulkLoader<'_> {
...
+    #[cfg_attr(
+        any(target_family = "wasm", not(feature = "rocksdb")),
+        expect(clippy::unnecessary_wraps)
+    )]
+    pub fn partial_commit(&mut self) -> Result<(), StorageError> {
+        match &mut self.kind {
+            #[cfg(all(not(target_family = "wasm"), feature = "rocksdb"))]
+            StorageBulkLoaderKind::RocksDb(loader) => loader.partial_commit(),
+            StorageBulkLoaderKind::Memory(_loader) => {
+                todo!("MemoryStorageBulkLoader::partial_commit");
```
Tpt, line 614: I think it's ok to just not do anything in the memory store

lib/oxigraph/src/store.rs
```diff
@@ -2083,6 +2096,9 @@ impl BulkLoader<'_> {
             drop(sender);
             while let Ok(batch) = receiver.recv() {
                 self.storage.load_batch(batch, target_num_threads)?;
+                if self.partial_commits {
+                    self.storage.partial_commit()?;
```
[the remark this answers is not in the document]
Tpt: but it requires explicit calls in every methods of this bulk loader except of keeping all the loading/saving logic in `RocksDBBulkLoader` so not sure...

lib/oxigraph/src/store.rs
```diff
@@ -1710,6 +1712,15 @@ impl BulkLoader<'_> {
         }
     }
 
+    /// If `true` (the default), save to database only at the end instead of doing
+    /// it regularly while loading.
+    ///
+    /// The default value is `true`.
+    pub fn with_atomicity(mut self, atomicity: bool) -> Self {
```
Tpt, line 1719: If the default is `true` why not making this method `without_atomity(mut self) -> self` to avoid the ambiguity on the default behavior?

progval: done

PR #1452, opened 2025-09-18
Update serd testsuite copy

Tpt:
Thanks @drobilla for the great work on these tests


drobilla:
You're welcome. At one point I'd intended to put more effort into getting many of them incorporated into the official test suites, but that's pretty painful. I've been trying to document them a little better anyway, though, so it's good to see them useful elsewhere.

PR #1458, opened 2025-09-26
Generalize the Storage interface as an UpdatableDataset trait

progval:
This implements the trait for both the Memory and the Rocksdb traits, and Storage simply delegates to either of them.

Future commits will extend the trait so it covers Store as well, which will allow new backends as alternative to the Store (eg. because they do not use EncodedTerm)

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -124,7 +124,7 @@ pub fn main() -> anyhow::Result<()> {
             if file.is_empty() {
                 // We read from stdin
                 let start = Instant::now();
-                let mut loader = store.bulk_loader().on_progress(move |size| {
+                let mut loader = store.bulk_loader()?.on_progress(move |size| {
                     let elapsed = start.elapsed();
                     eprintln!(
                         "{size} triples loaded in {}s ({} t/s)",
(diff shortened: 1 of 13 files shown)


progval:
I kept the new trait in `lib/oxigraph/src/storage/` for the time being so it doesn't affect any public API while we're still working on it, but it will probably move to `spareval` in the future

Tpt:
If we move sparql update implementation to each blackend, do we still need these traits?

Tpt:
I am not sure to follow you. If the implementation is duplicated for each backend then each implementation can call directly the backend methods, no need for a trait?

progval:
yes, but you would need to call that implementation through the trait

Tpt:
> yes, but you could only call that implementation through the trait

Indeed, if we want traits on top of the different stores. I am a bit reluctant to go this way because it means all storages must implement the same API and I am not sure if it will be easily workable in the future (e.g. if we introduce storages that use async I/O...)

progval:
then we can add a different set of traits for that other API.

But if we want interchangeable storages then we need them to agree to an API

progval:
(or to one API among a small set, eg. the current one + one that returns futures)

Tpt:
> But if we want interchangeable storages then we need them to agree to an API

Indeed, but do we want that? I guess the key missing information is how much users want to swap backends while keeping their code unchanged. If they want to do so a lot having traits make sense even if it means having less nice APIs because of genericity. If not having traits mean constraining quite a bit the API for not much upside. My preference at the moment is more towards not having traits and see how much we miss them.

progval:
My motivation for this is to add my own backend to Oxigraph that does not go through `Storage` but still make `oxigraph serve` be able to query it. This means that there must be, somewhere between MemoryStorage/RocksdbStorage and `oxigraph serve`, an API backed by either `Storage` or my backend so `oxigraph serve` can run either.

Where do we draw that line if not at the `Storage` level?

progval:
yeah that's the plan

progval:
So my plan was to extend these traits so both Storage and Store implement it. This is silly because Store sits on top of Storage and has a small different in quads_for_pattern that is a dealbreaker: the Iterator it returns has a different lifetime that can't be reconciled.

I'll try a different approach.

PR #1459, opened 2025-09-26
Add Oxigraph build for WASM C sandboxes

nka11:
Some AI generated efforts to write a WASM plugin to allow SPARQL and oxigraph store in typst templates.

Key aspects:
- Pure wasm32-unknown-unknown target without JavaScript dependencies
- Memory-only storage backend for plugin sandbox compatibility
- C-compatible exports for Typst plugin system integration
- Custom time provider implementation for document generation

Addresses the incompatibility issues with current WASM builds to provides a Typst plugin integration (or any other need to deploy oxigraph in sandboxed environment that is not javascript).

Reference : https://typst.app/docs/reference/foundations/plugin/


Tpt:
I am not a Typst user so I don't feel able to maintain this code. I feel it would be better maintained by yourself in an other crate. You did not had to touch Oxigraph internal so I don't see a blocker for this. The other oxigraph bindings are currently shipped in other crates by myself for pyoxigraph and oxigraph-js and by others for other bindings.

nka11:
I'm ok with that, I will move interface code to another crate. It looks necessary to have the featured build working on the mainstream to be able to use oxigraph[wasm] as dependency in hte subcrate as you depends on oxigraph[js] in your js crate. Will this be ok this way (Will mean merge the Cargo.toml only changes) ?

lib/oxigraph/Cargo.toml
```diff
@@ -13,10 +13,14 @@ description = "SPARQL database and RDF toolkit"
 edition.workspace = true
 rust-version.workspace = true
 
+[lib]
+crate-type = ["rlib", "cdylib"]
```
Tpt, line 17: not useful if you don't build the `oxigraph` crate as a dynamic library directly but an other crate having Oxigraph in it's dependencies. It does not apply in this case.

Tpt (commented):
Thank you for your reply. I am not sure we actually need any of the change to the Cargo (see inline comments)

PR #1462, opened 2025-09-28
Add support for running PreparedSparqlQuery on any QueryableDataset

progval:


lib/oxigraph/src/sparql/mod.rs
```diff
@@ -578,26 +578,32 @@ impl PreparedSparqlQuery {
...
+    pub fn on_transaction<'b>(
+        self,
+        transaction: &'b Transaction<'_>,
+    ) -> BoundPreparedSparqlQuery<'b> {
+        let reader = transaction.inner().reader();
+        let dataset = DatasetView::new(reader, &self.dataset);
+        self.on_queryable_dataset(dataset)
+    }
+
+    /// Bind the prepared query to the [`QueryableDataset`] it should be evaluated on.
```
Tpt, line 596: There is a trap here: opposite to the other two methods, the `self.dataset` definition is not applied so `FROM`, `FROM NAMED` and mutations done in Rust will be ignored. It should be at least documented imho:

```suggestion
    /// Bind the prepared query to the [`QueryableDataset`] it should be evaluated on.
    ///
    ///  <div class="warning">The query dataset specification accessible using [`Self::dataset`] is ignored by this method. You must take care of it before calling this method. </div>
```

This is something that sucks in the current API, it would be great to fix it in Oxigraph 0.6

progval: good point. I noticed that but forgot to document it

PR #1470, opened 2025-10-01
spareval: add QueryEvaluator::prepare

Tpt:
The prepared evaluator properly handles the query dataset (FROM and FROM NAMED), opposite to the existing ones


lib/spareval/src/lib.rs
```diff
@@ -596,6 +818,155 @@ pub trait AggregateFunctionAccumulator {
...
+            && self.named.is_none()
+    }
+
+    /// Returns the list of the store graphs that are available to the query as the default graph or `None` if the union of all graphs is used as the default graph.
+    /// This list is by default only the store default graph.
+    pub fn default_graph_graphs(&self) -> Option<&[GraphName]> {
+        self.default.as_deref()
+    }
+
+    /// Sets if the default graph for the query should be the union of all the graphs in the queried store.
```
progval, line 853:
```suggestion
    /// Sets the default graph of the query to be the union of all the graphs in the queried store.
```

lib/spareval/src/lib.rs
```diff
@@ -596,6 +818,155 @@ pub trait AggregateFunctionAccumulator {
...
+    }
+
+    /// Checks if this dataset specification is the default one
+    /// (i.e., the default graph is the store default graph, and all named graphs included in the queried store are available)
+    pub fn is_default_dataset(&self) -> bool {
+        self.default
+            .as_ref()
+            .is_some_and(|t| t == &[GraphName::DefaultGraph])
+            && self.named.is_none()
+    }
```
progval, line 845:
```suggestion
    pub fn is_default_dataset(&self) -> bool {
        self == Self::default()
    }
```

simpler.

and shouldn't it be renamed to `is_default_dataset_specifically`? (or `is_default` if you want it to be short)

progval (approved):
very nice!

lib/spareval/src/lib.rs
Tpt: > simpler

But requires an alloc

> `is_default`

Good idea. Will add a TODO (I just moved this struct so I don't want to introduce breaking changes).

progval: Hmm I assumed the compiler would be smart enough to compile it to the same code, but I just checked it's not. I guess I have Rust code elsewhere to update, then...

PR #1527, opened 2025-11-30
New query evaluator using DataFusion

Tpt:
Very WIP, will stay a draft for a long time


ashleysommer:
Wow @Tpt this looks amazing. This looks like it was an enormous amount of work.
Do you think this will (eventually) become the default query evaluator?
Are there any downsides, eg, particularly with regard to DataFusion using Apache Arrow in-memory structures, and the whole Oxigraph->Arrow conversion overhead?

Tpt:
Some missing elements in this MR:
- RocksDB `iter` is quite slow, making large reads not very good. Tweaking RocksDB evaluation to do less hash/merge joins and more "for loop" joins is needed to keep the same performances on top of RocksDB (and nudges more toward rewriting the storage to be more read-friendly at the cost of maybe being less write-friendly)
- `EXISTS` and `LATERAL` are written as correlated queries. Sadly DataFusion does not know how to evaluate them without decorating the queries first. We should either change the emitted logical plan in something DataFusion is able to decorellate, write a better decorrelation algorithm or add correlated evaluation to DataFusion (this is sorted by personal order of preference).

PR #1619, opened 2026-02-22
Add Store::open_with_options to configure RocksDB open file limits

zeroxbt:
This adds a small API for configuring RocksDB FD behavior when opening a disk-backed store.

### What changed

- Added `StoreOptions`:
  - `max_open_files: Option<i32>`
  - `fd_reserve: Option<u32>`
- Added `Store::open_with_options(path, options)`.
- Kept `Store::open(path)` as-is (same default behavior).
- Wired options through storage to RocksDB setup.
- RocksDB behavior is now:
  - use explicit `max_open_files` if provided
  - otherwise compute from FD limit with reserve (`fd_reserve`, default `48`)
- Added tests for:
  - opening with options
  - FD -> `max_open_files` computation

### Why

In write-heavy setups, FD count can get very high because `max_open_files` follows `ulimit -n`.  
This gives users a direct way to cap it without changing global process limits.

### Compatibility

No breaking API changes. Existing `Store::open(path)` behavior is unchanged.

### Ran

- `cargo test -p oxigraph test_load_graph_on_disk_with_options -- --exact`
- `cargo test -p oxigraph max_open_files_from_fd_limit`

### DIFF
--- a/lib/oxigraph/src/storage/rocksdb_wrapper.rs
+++ b/lib/oxigraph/src/storage/rocksdb_wrapper.rs
@@ -1369,7 +1369,7 @@ mod tests {
     #[expect(clippy::panic_in_result_fn)]
     fn contains_key_handles_empty_and_non_empty_values() -> Result<(), StorageError> {
         let dir = TempDir::new()?;
-        let db = Db::open_read_write(dir.path(), vec![])?;
+        let db = Db::open_read_write(dir.path(), vec![], DbOptions::default())?;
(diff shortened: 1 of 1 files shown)


Tpt (approved):
Thank you for this!

I just have a nit an let's see if the CI gets green (ignore the debian failure and the cargo deny one)

Tpt (approved):
Thank you for this and sorry for the slow review. I have a final nit, sorry for that. Will merge as soon as it's fixed.

PR #1665, opened 2026-04-11
JSON-LD: do not set default language if a type mapping is set

Tpt:


Tpt:
CC @magbak should fix your bug

PR #1690, opened 2026-04-25
Wire GeoSPARQL functions natively into spareval (PR 2/5)

patrykorwat:
Add a geosparql cargo feature to lib/spareval that pulls spargeo as an optional dependency and pre-populates QueryEvaluator's CustomFunctionRegistry with the 43 GEOSPARQL_EXTENSION_FUNCTIONS at construction time. Propagate the feature through lib/oxigraph so SparqlEvaluator::new() inherits the registered functions.

Drop the manual registration loops from cli/src/main.rs, python/src/sparql.rs, js/src/store.rs, and testsuite/src/sparql_evaluator.rs in favor of just enabling the feature. The oxigraph_geosparql_testsuite continues to exercise the same 43 functions, now via the auto-registered registry rather than the manual loop.

cli/src/service_description.rs keeps its direct dependency on spargeo because it iterates GEOSPARQL_EXTENSION_FUNCTIONS to build the SPARQL service description document, which is a list operation independent of evaluator registration.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -24,7 +24,7 @@ rdf-12 = ["oxigraph/rdf-12"]
 rocksdb-pkg-config = ["oxigraph/rocksdb-pkg-config"]
 rustls-native = ["oxigraph/http-client-rustls-native"]
 rustls-webpki = ["oxigraph/http-client-rustls-webpki"]
-geosparql = ["dep:spargeo"]
+geosparql = ["dep:spargeo", "oxigraph/geosparql"]
 
 [dependencies]
 anyhow.workspace = true
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -299,7 +299,7 @@ pub fn main() -> anyhow::Result<()> {
                 io::read_to_string(stdin().lock())?
             };
             let store = Store::open_read_only(location)?;
-            let mut evaluator = default_sparql_evaluator();
+            let mut evaluator = SparqlEvaluator::new();
             if let Some(base) = query_base {
                 evaluator = evaluator.with_base_iri(&base)?;
             }
@@ -451,7 +451,7 @@ pub fn main() -> anyhow::Result<()> {
                 io::read_to_string(stdin().lock())?
             };
             let store = Store::open(location)?;
-            let mut evaluator = default_sparql_evaluator();
+            let mut evaluator = SparqlEvaluator::new();
             if let Some(base) = update_base {
                 evaluator = evaluator.with_base_iri(&base)?;
             }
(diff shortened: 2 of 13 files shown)


cli/src/main.rs
```diff
@@ -1481,13 +1479,7 @@ fn evaluate_sparql_query(
 }
 
 fn default_sparql_evaluator() -> SparqlEvaluator {
```
Tpt: nit: what about just dropping this function and inline `SparqlEvaluator::new`?

lib/spareval/src/lib.rs
```diff
@@ -75,6 +75,29 @@ pub struct QueryEvaluator {
...
+        }
+    }
+}
+
+fn default_custom_functions() -> CustomFunctionRegistry {
+    #[cfg_attr(not(feature = "geosparql"), expect(unused_mut))]
+    let mut registry = CustomFunctionRegistry::default();
+    #[cfg(feature = "geosparql")]
+    for (name, implementation) in spargeo::GEOSPARQL_EXTENSION_FUNCTIONS {
+        registry.insert(name.into_owned(), Arc::new(implementation));
```
Tpt, line 96: Do we need the `Arc`? If I am not mistaken none of these functions have an internal state so they are restricted to a small vtable that should be cheap to copy.

cli/Cargo.toml
```diff
@@ -24,7 +24,7 @@ rdf-12 = ["oxigraph/rdf-12"]
 rocksdb-pkg-config = ["oxigraph/rocksdb-pkg-config"]
 rustls-native = ["oxigraph/http-client-rustls-native"]
 rustls-webpki = ["oxigraph/http-client-rustls-webpki"]
-geosparql = ["dep:spargeo"]
+geosparql = ["dep:spargeo", "oxigraph/geosparql"]
```
Tpt, line 27: aside for a follow up (glad to do it): it might be nice to add a function to `SparqlEvaluator` that lists the registered custom functions and allows to drop this explicit dependency and makes the SPARQL service description generation code cleaner

Tpt (approved):
Thank you! Just some very minor nit

lib/spareval/src/lib.rs
patrykorwat: Required by the existing CustomFunctionRegistry signature (`HashMap<NamedNode, Arc<dyn Fn(&[Term]) -> Option<Term> + Send + Sync>>`).
The 43 spargeo entries are fn pointers, but the registry's value type is Arc<dyn Fn>, so we still need to wrap. To skip the Arc for stateless fn pointers we'd need to change the registry to an enum that distinguishes fn-pointer entries from boxed closures, which I'd rather do as a follow-up PR if that's worth pursuing.

Tpt: Thank you! My memory failed me.

cli/Cargo.toml
patrykorwat: Agreed and happy for you to take that one. service_description.rs would become a clean iterator over the evaluator's registered functions and the spargeo dep on the CLI could go away entirely.

Tpt (approved):
Will merge as soon `cargo fmt` is happy.

Tpt (approved):

PR #1692, opened 2026-04-25
Add oxreason crate (OWL 2 RL forward chainer + SHACL Core scaffold) (PR 3a)

patrykorwat:
Adds the oxreason crate as a standalone library for OWL 2 RL forward chaining and SHACL Core validation over an oxrdf::Graph. No integration with the rest of oxigraph in this PR — Store::reason wiring will follow in PR 3b once this lands and the API surface is reviewed.

Coverage: 39 OWL 2 RL inference rules (M1 through M4), semi-naive evaluation with shadow-set deduplication, term interner backing rule body lookups for cax-sco/cax-eqc, five inconsistency detectors, plus a SHACL Core scaffold with sh:minCount implemented. See lib/oxreason/DESIGN.md for the rule coverage matrix and lib/oxreason/TESTING.md for the test strategy.

Tracks oxigraph#130. North star (paired with PR #1690): OWL 2 RL reasoning and GeoSPARQL composing in a single SPARQL query, so schema entailments and spatial predicates evaluate together rather than via two separate systems.

The plan for oxreason introduction:

1. PR 3a: scaffold lib/oxreason + M1 rules + minimal forward chainer + integration tests for M1.
2. PR 3b: M2 property and class rules.
3. PR 3c: M3 schema rules + SHACL sh:minCount.
4. PR 3d: Store::reason method on oxigraph::Store, with a reasoning feature flag, end-to-end test that combines reasoning + GeoSPARQL in one query.


patrykorwat:
@Tpt sorry for a, yet again, larger PR.

This one is specifically big as I add OWL RL rules (1 SHACL rule as well) + reasoning engine and SHACL validator.

The most citical part of the engine is the evaluation loop. Once you understand it, the rest is obvious (at least, I hope).

The supportive material that can help with review:

- https://github.com/patrykorwat/oxigraph/tree/pr-split-3a-oxreason-scaffold-v2/lib/oxreason#engine-pipeline
- https://github.com/patrykorwat/oxigraph/blob/pr-split-3a-oxreason-scaffold-v2/lib/oxreason/src/engine.rs#L44-L102

Tpt:
Thank you for pushing this! I will try to have a careful look at it this week (the week might be a bit busy for me so no promises).

I ask myself if this should not be better maintained outside of Oxigraph if it's your design and do not require deep Oxigraph integration. What I would love is avoiding multiple execution system. My feeling is that if we add a least fixed point operator to the SPARQL evaluator we can the have OWL 2 RL using the same code as SPARQL (I wrote an experiment around that three years ago but never finished it). But it's very much an open question.

patrykorwat:
Thanks for the thought. I agree the SPARQL fixpoint operator is the cleaner long-term destination, and it matches the north-star here better than oxreason as a separate engine.

But oxreason is built, tested and benchmarked against reasonable, and a fixpoint operator is a much bigger effort. Keeping it in tree means there'a s single repo to evolve code, with one issue tracker, one CI, and a clear retirement path when the unified evaluator matures.

I propose to land oxreason here as transitional OWL 2 RL implementation and replace it once the fixpoint operator is in. Happy to help on the fixpoint design whenever your bandwidth allows.

Tpt:
Sorry for the slow answer.

> But oxreason is built, tested and benchmarked against reasonable, and a fixpoint operator is a much bigger effort. Keeping it in tree means there'a s single repo to evolve code, with one issue tracker, one CI, and a clear retirement path when the unified evaluator matures.

Yes! But it still more code for me to maintain, slowing my progress elsewhere. I would tend to prefer it to be maintained out of tree to leave me more bandwidth for everything else. I think we should prioritize the fixpoint operator (there is already one in DataFusion, I have an old prototype of one in spareval, the main missing thing is to port properly property paths on it).

patrykorwat:
@Tpt Let me have a look at the DF. For now, I'll close this PR. Btw, I've been testing the branch reasoning code with up to 100,000 triples and the results are as below:
obraz

Hope to see even lower values in the upcoming future :)

eljeffeg:
@patrykorwat were you planning to create a separate project for `oxreason`?  I was planning to integrate [reasonable](https://crates.io/crates/reasonable) into my project [Oxigraph Nova](https://github.com/eljeffeg/oxigraph-nova), but it's not a perfect fit, and your oxreason work looks like a better match for my architecture.

I like @Tpt's idea of using a least fixed-point operator in the evaluator.  I have a different evaluator (WCOJ/LFTJ) from Oxigraph (traditional iterator), but I expect that if I can make it work, it would be applicable to both.  

I could add oxreason to my repo (or pull in your crate).  As you mentioned above, we could start with materialization and work toward the fixed-point architecture.  It could be the experimental testing ground before it gets moved to Oxigraph.

For shacl, I'd be more inclined to pull in https://github.com/rudof-project/rudof than start from scratch with oxreason, but I'd have to look at the difficulty there.

eljeffeg:
@Tpt, your instinct was spot on. I ended up implementing OWL 2 RL reasoning by running it directly through Nova’s existing query evaluator instead of building a separate engine.  Rules are evaluated as normal joins using the same leapfrog triejoin machinery that the SPARQL evaluator already uses. There’s no second code path for reasoning — the fixpoint loop just repeatedly runs rule bodies as joins over the current data + newly derived facts until nothing new shows up.

The trickiest part was handling the delta efficiently during semi-naive evaluation. Since the main index is only rebuilt during compaction, I added a small transient in-memory structure that the join engine can read from alongside the stable index in the same pass. It turned out to be simpler than expected.

I took a lot of inspiration from @patrykorwat’s oxreason work on the rule representation and overall approach. Right now, Nova supports the core OWL 2 RL rules around subclassing, subproperties, domain/range, transitivity, and symmetry. I still need to add the owl:sameAs and the consistency rules.

It’s exposed as a simple decorator (ReasoningDataset) that layers inferred triples onto any existing dataset. On the server side you just pass --reasoning and it rebuilds the inferences after compaction. The service description also advertises OWL-RL entailment.

Overall, reusing the evaluator’s join engine for reasoning worked out well and didn’t require big changes to the core query code. Happy to share more details or compare notes if you pick the fixpoint operator work back up.

@patrykorwat would love if you could run your benchmarks on it.

PR #1715, opened 2026-05-12
Add option using blob file

maxiprogram:
StoreOptions
oxigraph::store::Store::open_with_options()
optional using blob file

### DIFF
--- a/lib/oxigraph/src/storage/rocksdb_wrapper.rs
+++ b/lib/oxigraph/src/storage/rocksdb_wrapper.rs
@@ -47,6 +47,8 @@ pub struct ColumnFamilyDefinition {
 pub struct DbOptions {
     pub max_open_files: Option<i32>,
     pub fd_reserve: Option<u32>,
+    pub is_enable_blob: Option<bool>,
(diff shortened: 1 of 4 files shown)


Tpt:
Thank you for this MR! Out of curiosity, have you seen any performance improvement with it? We only use values for the strings column families, the other ones are key-only.

lib/oxigraph/src/storage/rocksdb_wrapper.rs
```diff
@@ -350,6 +352,16 @@ impl Db {
                     })
                     .0,
             );
+            if let Some(is_enable_blob) = db_options.is_enable_blob {
+                if is_enable_blob {
+                    rocksdb_options_set_enable_blob_files(options, 1);
+                } else {
+                    rocksdb_options_set_enable_blob_files(options, 0);
+                }
```
Tpt, line 360: tip: this is a bit shorter: 
```suggestion
               rocksdb_options_set_enable_blob_files(options, is_enable_blob.into());
```

PR #1724, opened 2026-05-18
oxrdf: introduce our own OxString type

Tpt:
This is a combination of Arc<str> and &str

### DIFF
--- a/lib/oxrdf/src/string.rs
+++ b/lib/oxrdf/src/string.rs
@@ -1,0 +1,595 @@
+#![expect(unsafe_code)]
+
+// TODO: add safety guards with max refcount and max len
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
+use std::alloc::{Layout, alloc, dealloc};
+use std::borrow::Borrow;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::marker::PhantomData;
+use std::ops::Deref;
+use std::ptr::NonNull;
+use std::sync::atomic::{AtomicUsize, Ordering, fence};
+
+/// Owned variant of [`OxStr`]: A compact string type for reference-counted owned data or static slices.
+///
+/// `OxString` is conceptually a fusion of [`Arc<str>`](std::sync::Arc) and [`Cow<'static, str>`](std::borrow::Cow):
+/// it allows storing a static string slice ([`&'static str`](std::str))
+/// or a reference-counted fixed-sized string ([`Arc<str>`](std::sync::Arc)), enabling cheap clones of owned data.
+///
+/// When owned, cloning is cheap and increments an atomic reference count.
+///
+/// See [`OxStr`] for implementation details.
+///
+/// ```
+/// use oxrdf::OxString;
+///
+/// let value = OxString::new("hello");
+/// assert_eq!(value.as_str(), "hello");
+/// ```
+pub type OxString = OxStr<'static>;
+
+/// A compact string type that can be either borrowed or reference-counted owned data.
+///
+/// `OxStr` is conceptually a fusion of [`Arc<str>`](std::sync::Arc) and [`Cow<'a, str>`](std::borrow::Cow):
+/// it allows storing a string slice ([`&str`](std::str))
+/// or a reference-counted fixed-sized string ([`Arc<str>`](std::sync::Arc)), enabling cheap clones of owned data.
+///
+/// It is not relying on an enum but uses an optimized layout, storing only a pointer and a `usize` length.
+/// It relies on a magic bit in the length to know if the value is borrowed or owned.
+/// If borrowed, the pointer directly targets the string bytes.
+/// If owned, the pointer points to a memory allocation with first the reference counter, then the string bytes.
+///
+/// When owned, cloning is cheap and increments an atomic reference count.
+///
+/// ```
+/// use oxrdf::OxStr;
+///
+/// let borrowed = OxStr::new("hello");
+/// let owned = OxStr::new_owned("hello");
+///
+/// assert_eq!(borrowed.as_str(), "hello");
+/// assert_eq!(owned.as_str(), "hello");
+/// ```
+pub struct OxStr<'a> {
+    len: usize,
+    data: NonNull<u8>,
+    _marker: PhantomData<&'a ()>,
+}
+
+const OWNED_FLAG: usize = 1 << (usize::BITS - 1);
+
+impl<'a> OxStr<'a> {
+    /// Creates an `OxStr` borrowing from `value`.
+    ///
+    /// This does not allocate and keeps the lifetime of `value`.
+    ///
+    /// ```
+    /// use oxrdf::OxStr;
+    ///
+    /// let value = OxStr::new("abc");
+    /// assert_eq!(value.as_str(), "abc");
+    /// ```
+    #[inline]
+    pub const fn new(value: &'a str) -> Self {
+        Self {
+            len: value.len(),
+            // SAFETY: &str pointers are not null
+            data: unsafe { NonNull::new_unchecked(value.as_ptr().cast_mut().cast()) },
+            _marker: PhantomData,
+        }
+    }
+
+    /// Creates an owned `OxStr` by copying `value`.
+    ///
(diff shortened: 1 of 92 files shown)


lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+/// `OxStr` is conceptually a fusion of [`Arc<str>`](std::sync::Arc) and [`Cow<'a, str>`](std::borrow::Cow):
+/// it allows storing a string slice ([`&str`](std::str))
+/// or a reference-counted fixed-sized string ([`Arc<str>`](std::sync::Arc)), enabling cheap clones of owned data.
+///
+/// It is not relying on an enum but uses an optimized layout, storing only a pointer and a `usize` length.
+/// It relies on a magic bit in the length to know if the value is borrowed or owned.
+/// If borrowed, the pointer directly targets the string bytes.
+/// If owned, the pointer points to a memory allocation with first the reference counter, then the string bytes.
+///
+/// When owned, cloning is cheap and increments an atomic reference count.
```
adamreichold, line 45: This does not say anything about it, but it might suggest the cloning is expensive in the borrowed case. Maybe add that the borrowed variant is essentially `Copy`?

lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+            unsafe {
+                // Load and fence from Arc implementation
+                if self.owned_counter().fetch_sub(1, Ordering::Release) != 1 {
+                    return;
+                }
+                fence(Ordering::Acquire);
+                dealloc(
+                    self.data.as_mut(),
+                    Self::owned_layout_for_len(self.owned_len()),
+                );
```
adamreichold, line 350: Admittedly something to measure, but my first instinct here would be to put this into a separate function so it isn't inlined?

lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+    }
+}
+
+impl Clone for OxStr<'_> {
+    #[inline]
+    fn clone(&self) -> Self {
+        if self.is_owned() {
+            // SAFETY: we just checked it's the owned variant, we can call owned_counter
+            unsafe {
+                self.owned_counter().fetch_add(1, Ordering::Relaxed); // Arc is also using relaxed, I guess it's fine
```
adamreichold, line 362: These comments might be a bit cavalier for the amount of unsafe code. It should be possible to reproduce reasoning that went into `Arc` here, shouldn't it? Especially if the approach taken by `Arc` ever changes.

(I think the main point is that `drop` is basically the only synchronization point between different users of an owned-but-shared instance, so this is where acquire-release semantics are necessary. The split of these into release for the `fetch_sub` and `fence(Acquire)` only before deallocation is another micro-optimization to avoid the acquire ordering when only a reference is dropped.)

lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+
+#[cfg(feature = "serde")]
+impl Serialize for OxStr<'_> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.as_str().serialize(serializer)
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for OxStr<'_> {
```
adamreichold, line 501: Wouldn't it make sense for `OxStr` to optionally borrow from the deserializer input? Does implementing `impl<'de> Deserialize<'de> for OxStr<'de>` interact unfavourably with variance?

lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.as_str().fmt(f)
+    }
+}
+
+impl fmt::Display for OxStr<'_> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.as_str().fmt(f)
```
adamreichold: Are you sure these implementations will not accidentally cross-call between `<&str as fmt::Debug>::fmt` and `<&str as fmt::Display>::fmt`?

lib/oxrdf/src/string.rs
```diff
@@ -0,0 +1,595 @@
...
+    #[inline]
+    fn from(value: &'a str) -> Self {
+        Self::new(value)
+    }
+}
+
+impl From<String> for OxStr<'_> {
+    #[inline]
+    fn from(value: String) -> Self {
+        Self::new_owned(&value)
```
adamreichold, line 489: Here is a weird idea: In code bases where I used `Arc<str>`, these places where I would copy an intermediate `String` into the final, immutable, shared `Arc<str>` would often be visible as low-level "distributed fat" in CPU profiles.

Since you control the layout and the length is stored inline, if you would put the atomic reference behind instead of in front of the main allocation, you could "adopt" the backing buffer of the `String` and `realloc` it to add your reference count to the end, at least giving the allocator a chance to reuse the underlying allocation instead of always having to copy the data into a new one.

Then again, putting the reference count at the end might also kill spatial locality and thereby cache effectiveness and make the offset computation for `Clone`ing more expensive. As I wrote in the beginning, a bit of a weird idea...

(Another more involved yet simple option would be to provide a "builder" type that starts with the right memory layout and is then frozen into the immutable form. But this won't help with third-party code that insists on dealing in terms of `String`.)

Tpt: Thank you for reviewing this!

Yes to everything! I had also this idea. But then I would need to also store the length of the allocation to give the proper layout to `dealloc`. Also, it does not work all the time, ones need some extra bytes at the end of the input for that.

On the builder, I already build a poor man version of it with the `concat` function that is quite useful for prefixed names.

lib/oxrdf/src/string.rs
Tpt: Good call. Will make the trait explicit.

lib/oxrdf/src/string.rs
Tpt: > interact unfavourably with variance

Yes! Rust will scream if we provide both `impl<'de> Deserialize<'de> for OxStr<'static> {` and `impl<'de> Deserialize<'de> for OxStr<'de> {`. I don't see a good way out of it.

lib/oxrdf/src/string.rs
Tpt: Indeed. my hope is that fence + `dealloc` is compiled into slightly more than a method call but I might be wrong. Will benchmark that.

lib/oxrdf/src/string.rs
Tpt: Great point. Thanks! Will fix

lib/oxrdf/src/string.rs
adamreichold: > But then I would need to also store the length of the allocation to give the proper layout to dealloc.

I would assume the length of the allocation stays exactly the same as in the current approach, i.e. you would still `realloc` to drop any "excess" capacity as e.g. `String::into_boxed_str` does. I think that this is still often a no-op has more to do with the size classes used by common `malloc` implementations which means the block is effectively re-used even though the Rust-level allocation is reduced in length.

>  Also, it does not work all the time, ones need some extra bytes at the end of the input for that.

The idea is also that even though you do `realloc` to get an exact fit of the same size you use now, the underlying "blockiness" of the allocator makes it into an almost no-op. Also the `String`s will have often have some excess capacity anyway due to their power-of-two re-allocation scheme.

But yes, this is all "statistically" speaking, so you are probably better off starting with the straight forward approach and then give this a try and measure the effects as a separate change. No hurry, as long as the layout is an implementation detail you control.

lib/oxrdf/src/string.rs
adamreichold: One could give more control to downstream code by providing modules to be used with Serde's `with` field attribute, so downstream code can explicitly request optionally borrowing and/or always owning behaviour.

(The given impl would probably still remain a better default though.)

lib/oxrdf/src/string.rs
Tpt: Thanks! Will give it a try in a follow-up

PR #1733, opened 2026-05-26
bugfix/optimizer: fix quadratic scaling of OPTIONAL in a JOIN on FK

nka11:
I have investigated the reported OPTIONAL performance issue and confirmed it as a bug in the Oxigraph query optimizer. The issue causes quadratic scaling when an OPTIONAL clause contains a multi-pattern group (join)
  over a foreign-key shape.

  Root Cause Analysis
  My investigation revealed that:
   1. Poor Join Reordering in OPTIONAL RHS: When sep-0006 (LATERAL join) is enabled (which is the default in Oxigraph), the optimizer transforms OPTIONAL joins into LATERAL joins if the right-hand side (RHS) is "fit"
      (e.g., a join of simple patterns). However, it was reordering the RHS patterns using only global input bindings, ignoring the bindings that would be provided by the left-hand side (LHS) of the OPTIONAL.
   2. Unselective Starting Patterns: Because it ignored LHS bindings (like ?c in the reproducer), the optimizer chose a starting pattern for the RHS that matched many triples globally (e.g., ?order a schema:Order
      instead of ?order schema:customer ?c).
   3. Quadratic Evaluation: The resulting plan re-evaluated the full list of RHS candidate triples for every row of the LHS, leading to O(|LHS| × |RHS_Total|) complexity instead of the expected O(|LHS| + |RHS_Total|).
   4. Cost Model Ambiguity: The cost model assigned equal weight to rdf:type patterns and property patterns when both had one variable bound, making it susceptible to picking the less selective type pattern first if
      it appeared earlier in the query.

  Changes Applied
  I have implemented the following fixes in lib/sparopt/src/optimizer.rs:
   - Context-Aware RHS Reordering: Modified Optimizer::reorder_joins for LeftJoin to re-optimize the RHS using the variable types inferred from the LHS. This ensures that the RHS join order is optimal for the bindings
     actually provided during execution.
   - Property-First Cost Model: Added a small penalty to rdf:type patterns in the cost model (estimate_graph_pattern_size). This ensures that if a property join and a type join have otherwise equal costs, the property
     join is preferred as it is generally more selective.
   - Regression Test: Added a permanent regression test test_optional_lateral_reordering in lib/sparopt/src/optimizer.rs that verifies the optimized plan for the reported query avoids the quadratic pattern.

  Verification
  I verified the fix using the provided reproducer query. The optimized algebra now correctly starts the OPTIONAL RHS chain with the selective property join:
   - Before: RHS started with ?order a schema:Order (matched all orders for every customer).
   - After: RHS starts with ?order schema:customer ?c (matches only orders for the specific customer bound by the LHS).


Tpt:
Thank you for this! Ping me if when you want a review.

nka11:
> Thank you for this! Ping me if when you want a review.

Trying first to ge the CI green

nka11:
@codspeedbot fix this regression

nka11:
@Tpt The codespeed shows significant improvements, however some other checks are failing.
The last commit doesn´t change the codespeed. Removing the draft

lib/sparopt/src/optimizer.rs
```diff
@@ -1069,3 +1101,86 @@ fn is_named_node_pattern_bound(pattern: &NamedNodePattern, input_types: &Variabl
         NamedNodePattern::Variable(v) => !input_types.get(v).undef,
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Tpt: we have a way to write optimizer tests without writing rust code in `testsuite/oxigraph-tests/sparql-optimization`. You write to `.rq` files, one with the input query and one with the optimize query and then add test to `manifest.ttl`. Glad to do it in a follow up if you don't feel comfortable with it.

Tpt (commented):
Thank you! Looks great to me.

As you see the optimizer is not great at all at the moment.

Just a small suggestion on the way to write tests. Glad to do it in a follow up if you prefer.

nka11:
Have to write a runtime test to guard the regression with verification of the explain. Please let me know your preference for the seeding (as it is written...), or for a different way to write this test/guard.

Tpt:
I see. Maybe a benchmark would be nicer and less brittle? But let's not waste your time on this and ship this improvement. CI seems to fail on `serde_json` not available. Glad to merge when it's fixed.

nka11:
CI should pass now

Tpt:
@nka11 Sadly not, Clippy and fmt are still unhappy. `cargo clippy --all-features --all-targets` might help

nka11:
Hi @Tpt , 

There are some remaining fails on tests that changed (I added some), and the performance benchmark variations.
Tell me what's missing, so this can be shipped...

Regards

Tpt:
Thank you for these painful iterations. CI is now green \o/

On benchmarks, small variations are expected (it might just be sparopt that is slightly slower now). The benchmark tooling is also not fully stable. In any case, the benefits outweigh significantly the regressions.

PR #1742, opened 2026-05-31
OxStr: move dealloc call outside hot path

Tpt:


adamreichold:
Another day ending in y: I am bad at estimating CPU performance. 

Sorry for the noise over at #1724!

Issue #1745, opened 2026-06-04
Lexer uses a buffer with a fixed max `usize`, not supporting large WKT values in triples


C-Loftus:
Thank you very much for your great work on oxigraph

**Describe the bug**

In the `lexer.rs` file used for parsing rdf data there is currently a max buffer size which uses `usize`.
https://github.com/oxigraph/oxigraph/blob/25c89bb30320f3efb82d4ed5dc800b40cccd65f0/lib/oxttl/src/toolkit/lexer.rs#L64
If you reach this total, the parse will fail and not try to grow the buffer larger.
https://github.com/oxigraph/oxigraph/blob/25c89bb30320f3efb82d4ed5dc800b40cccd65f0/lib/oxttl/src/toolkit/lexer.rs#L121-L127

When trying to parse my large nquad data, I get the error:
`Reached the buffer maximal size of 16777216`

For knowledge graphs with geospatial information, you embed WKT strings which represent entire geometries. For large geometries with many vertices (i.e. detailed geometries for an entire country etc.), it is not hard to go above 16MB. As such, this is related to support for #1560 

**To Reproduce**

Generate an nquad with one very large line that is above the value of `usize` on the host.   Try to parse it with the `NQuadsParser::parse()` function.

I believe the error is coming from the `Result<>` here. 

```rs
    // assume the triples reader is just a reader for a file on disk or something similar, 
    // not read into memory all at once
    let parsed_quads = parser.for_reader(reader);
    for quad in parsed_quads {
        let quad = quad?
```

**Other Context**

It seems like this is a fairly common challenge in many rdf libraries, for instance, I raised this in goRDFlib and had this fixed recently as well: https://github.com/tggo/goRDFlib/issues/12

C-Loftus:
I would be happy to submit a PR for this as well, just let me know if that would be appropriate, and if so, what the PR should edit / refactor.

Tpt:
Hey! Thank you for this nicely written issue. I was expecting this to come up, it took 2 years.

I set this upper limit to avoid out of memory errors in case of invalid Turtle files (unclosed strings or IRIs...). But indeed, it leads to issues on very large literals.

I see two way forward:
1. Increase the constant. But it only mitigate the issue and not solve it
2. Add an option to the NQuad parsers (and the other similar parsers like Turtle, TriG...) to enable setting a larger limit.

Option 2 seems nicer to me. I would love an MR in this direction.

C-Loftus:
[...]

PR #1746, opened 2026-06-04
Customize the max buffer size for parsing

C-Loftus:
First pass at a PR for for customizing the max buffer size for parsing RDF data. Addresses #1745 

Once this seems initially reasonable I will go in and add tests and comments for anything non obvious.  Main questions I have:

1. Are there any other places I should put this? I sought to mainly just expose this inside the parser structs and nothing higher level. i.e. someone has to explicitly opt in to this.
2. Is this pattern of keeping the `new_parser` function and adding another `new_parser_with_custom_buffer_size` a reasonable choice? Or is that too verbose? I am trying to keep the parser buffer size immutable in the lower level lexer code and still allow the higher level structs to be created with both a sensible default implicitly and the custom size


lib/oxttl/src/n3.rs
Tpt: This means that `N3Parser::default` will have `max_buffer_size` set to `0`. I think we need to implement `Default` manually. Then we can move back `new` to call `default` or maybe the inverse, instantiate the structs fully in `new` and then make `default` call new? I actually prefer the later slightly, but does not care much.

lib/oxttl/src/lib.rs
```diff
@@ -23,6 +23,8 @@ pub use crate::trig::{TriGParser, TriGSerializer};
 pub use crate::turtle::{TurtleParser, TurtleSerializer};
 
 pub(crate) const MIN_BUFFER_SIZE: usize = 4096;
-pub(crate) const MAX_BUFFER_SIZE: usize = 4096 * 4096;
+/// Represents a reasonable default max buffer size for parsing to avoid OOM errors
+/// A different usize can be used or [`usize::MAX`] can be used to disable the limit
```
Tpt: this line would be better on the public method than on this private constant

lib/oxttl/src/n3.rs
```diff
@@ -194,13 +194,22 @@ pub struct N3Parser {
...
     #[inline]
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            max_buffer_size: DEFAULT_MAX_BUFFER_SIZE,
+            ..Self::default()
+        }
+    }
+
+    pub fn with_max_buffer_size(mut self, max_buffer_size: usize) -> Self {
```
Tpt, line 225: It might be nice to document the function a bit. Something like (not sure if it's great, tweaks welcome):

```suggestion
    /// Define an upper bound for the internal buffer of the parser in bytes
    ///
    /// This limits the memory consumption of the parser but also the maximum size of parsed IRIs and literals.
    ///
    /// The default is set to 16MB, use this function to change it (e.g. to [`usize::MAX`] to not set an upper bound).
    pub fn with_max_buffer_size(mut self, max_buffer_size: usize) -> Self {
```

Tpt (commented):
Thank you! Looks great! I found a bug in `default` and I would love some doc comments (glad to do it in a follow up if you prefer). I commented on N3 but the comments are also applicable elsewhere

lib/oxttl/src/n3.rs
C-Loftus: Agreed, was just waiting to add comments until general structure looked good 👍

lib/oxttl/src/n3.rs
```diff
@@ -188,19 +188,42 @@ impl From<Quad> for N3Quad {
...
+            max_buffer_size: DEFAULT_MAX_BUFFER_SIZE,
+        }
+    }
+
+    /// Define an upper bound for the internal buffer of the parser in bytes
+    ///
+    /// This limits the memory consumption of the parser but also the maximum size of parsed IRIs and literals.
+    ///
+    /// The default is set to 16MB, use this function to change it (e.g. to [`usize::MAX`] to not set an upper bound).
+    pub fn with_max_buffer_size(mut self, max_buffer_size: usize) -> Self {
```
Tpt, line 225: hyper-nit: make the function inline-able
```suggestion
    #[inline]
    pub fn with_max_buffer_size(mut self, max_buffer_size: usize) -> Self {
```

lib/oxttl/src/n3.rs
```diff
@@ -1488,3 +1520,18 @@ impl<'a> Iterator for N3PrefixesIter<'a> {
...
+
+    #[test]
+    fn test_default_buffer_size() {
+        // Ensure that the default function and new function both
+        // return the same non-zero default value for the max_buffer_size
+        let default_parser = N3Parser::default();
+        let new_parser = N3Parser::new();
+        assert_eq!(default_parser.max_buffer_size, DEFAULT_MAX_BUFFER_SIZE);
+        assert_eq!(new_parser.max_buffer_size, DEFAULT_MAX_BUFFER_SIZE);
+    }
```
Tpt: thanks! Imho I would not bother with tests, if  the max_buffer_size is 0, parsing will just not run.

Tpt (commented):
Thank you! Two very minor nits but happy to merge this MR in its current state if you don't feel like to work on them (just say so).

lib/oxttl/src/n3.rs
C-Loftus: Will do 👍

lib/oxttl/src/n3.rs
C-Loftus: Will remove these

C-Loftus:
Thanks for your comments @Tpt , I will change to inline and remove the `default` / `new` tests.

One question, I am trying to write one last test for parsing above the max default size and ensuring it errors when expected and won't when `usize::MAX` is set. (However maybe this isn't necessary either and in that case I can skip adding this)
However, this test with an expected error value, takes longer than 60 seconds to run. Do you have any idea why that might be the case? I wanted to ensure its something on the Rust toolchain side of things maybe getting slowed due down to a huge string in the test error message (or something similar) and not something about the internals of the parsing logic.

```rs
fn test_parse_nquads_above_max_default_size() {
        // Ensure that the parser can be configured to parse files above the default max buffer size
        // and that it does not return an error when parsing such files
        let unbound_parser = NQuadsParser::default().with_max_buffer_size(usize::MAX);
        let bounded_parser = NQuadsParser::default();

        // Create a literal > 16 MiB (16 MiB + 1 byte)
        let large_literal = "x".repeat(16 * 1024 * 1024 + 1);

        let file =
            format!(r#"<http://example.com/foo> <http://schema.org/name> "{large_literal}" ."#);

        for quad in bounded_parser.for_reader(file.as_bytes()) {
            match quad.unwrap_err() {
                TurtleParseError::Io(err) => {
                    assert_eq!(err.kind(), io::ErrorKind::OutOfMemory);
                }
                err => panic!("expected TurtleParseError::Io, got {err:?}"),
            }
        }

        for quad in unbound_parser.for_reader(file.as_bytes()) {
            assert!(quad.is_ok());
        }
    }
```

Tpt:
@C-Loftus Great! Point. I think this is because we grow the buffer linearly: https://github.com/oxigraph/oxigraph/blob/fd46987d25c55d31d0bb6ef6385673601965530c/lib/oxttl/src/toolkit/lexer.rs#L130

I think we should grow it exponentially (double it each time) to avoid this. If you do the change here, please do it also for `AsyncRead`: https://github.com/oxigraph/oxigraph/blob/fd46987d25c55d31d0bb6ef6385673601965530c/lib/oxttl/src/toolkit/lexer.rs#L158

C-Loftus:
Sounds good will work on that. It is slightly confusing to me since it worked just fine and completed quickly when the parser was created with usize::max parsing the same 16mb + 1 data, but only caused timeout issues in tests when it failed with the default. But it probably makes sense to make the buffer grow by doubling each time regardless so will try that.

Do you want me to bring back a test for parsing an nquad above the 16mb default? Or skip that test? Just wanted to align on what you prefer. Thanks!

Tpt:
@C-Loftus I think we can skip the test and add one if you change the buffer growing behavior.

C-Loftus:
Ok I have
- Changed the buffer from using `+ min_buffer_size` to `* 2`. Since `min_buffer_size` is no longer used, I removed it. 
- I added a test for the buffer using a mock parser, to ensure the lexer is tested separate from the upstream parsers for specific data formats.
- I kept in the test for `test_parse_nquads_above_max_default_size`. As I've written more, I think some of the buffer logic is non trivial and would be easy to accidentally cause regressions in the upstream parsers if the lexer functionality changes, so I think it is best to keep one test at the parser level if that would be reasonable to you.

lib/oxttl/src/toolkit/lexer.rs
```diff
@@ -431,3 +428,59 @@ impl<B: Deref<Target = [u8]>, R: TokenRecognizer> Lexer<B, R> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn buffer_grows_exponentially() {
```
Tpt, line 480: The old test checking the unbouneded buffer support might be an easier way to smoke test this. WDYT ?

Tpt (commented):
Parsing seems to be broken, not sure why.

Tpt:
>  I've written more, I think some of the buffer logic is non trivial and would be easy to accidentally cause regressions in the upstream parsers if the lexer functionality changes, so I think it is best to keep one test at the parser level if that would be reasonable to you.

Make sense

lib/oxttl/src/turtle.rs
```diff
@@ -44,19 +44,43 @@ use tokio::io::{AsyncRead, AsyncWrite};
...
+            base: None,
+            prefixes: HashMap::new(),
+        }
+    }
+
+    /// Define an upper bound for the internal buffer of the parser in bytes
+    ///
+    /// This limits the memory consumption of the parser and the maximum size of parsed IRIs and literals.
+    ///
+    /// The default is set to [`DEFAULT_MAX_BUFFER_SIZE`] bytes, use this function to change it
```
Tpt: The constant is not public. Why not mention 16MB like the other parers?

lib/oxttl/src/toolkit/lexer.rs
```diff
@@ -122,12 +119,12 @@ impl<R: TokenRecognizer> Lexer<Vec<u8>, R> {
...
                 io::ErrorKind::OutOfMemory,
                 format!(
-                    "Reached the buffer maximal size of {}",
+                    "Reached the buffer maximal size of {}. The buffer size can be increased at the cost of higher memory use if large data is required",
                     self.max_buffer_size
                 ),
             ));
         }
-        let min_end = min(self.data.len() + self.min_buffer_size, self.max_buffer_size);
+        let min_end = min(self.data.len() * 2, self.max_buffer_size);
```
Tpt: I found the issue: if `self.data` is empty it will stay this way. I think we need to be smarter like ensuring we read at least `min_buffer_size` all the time.

lib/oxttl/src/turtle.rs
C-Loftus: I tried to avoid hard coding a specific number in case the value changed. That way the docs are always in sync. I think honestly the default max buffer size may make sense to be public since even though it isn't necessary to use in parser logic directly, it does affect how clients choose to parse data. (i.e. `if x > DEFAULT_MAX_BUFFER_SIZE skip` would be a reasonable operation for a client to perform`)

lib/oxttl/src/toolkit/lexer.rs
C-Loftus: Hmmm interesting that makes sense. Sorry I missed this. I will not get to this today but can take a look tomorrow most likely.

I tried to run the tests locally but was getting this. (Figured I'd save this as a follow up; think I may be missing some system dependencies). 

```rs
error: couldn't read `oxrocksdb-sys/rocksdb/src.mk`: No such file or directory (os error 2)
  --> oxrocksdb-sys/build.rs:75:27
   |
75 |     let mut lib_sources = include_str!("rocksdb/src.mk")
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0282]: type annotations needed
  --> oxrocksdb-sys/build.rs:84:18
   |
84 |         .filter(|p| !p.is_empty())
   |                  ^   - type must be known at this point
   |
help: consider giving this closure parameter an explicit type
   |
84 |         .filter(|p: /* Type */| !p.is_empty())
   |                   ++++++++++++
```

Tpt: `git submodule init && git submodule update` should fix that.

Tpt: Something like this that makes sure we use the full buffer and, if it's full, increase its size:
```rust
let new_start = self.data.len();
if self.data.len() == self.data.capacity() {
    // We are at full capacity, we need to reallocate and we increase 2x the buffer size
    let new_len = min(self.data.len() * 2, self.max_buffer_size);
    self.data.resize(new_len, 0);
} else {
    // We are not at capacity, we just make sure to use the full buffer
    self.data.resize(self.data.capacity(), 0);
}
```

C-Loftus:
Should be all set to review again now. Thanks for guidance on this. I changed a bit of the logic on the buffer resizing and added some comments / variable renames for clarity. There was a few edge cases both when the data is empty but also if multiplying by 2 would silent cause a wrap around. I have addressed both scenarios with a new helper function.

Regarding CI failures:
- Codspeed says it is slower but sure how important that is. ~5ms which I presume is simply because we default to increase the buffer by 2x or it could just be statistical noise.
- Codecov, Linux linking, and semver appear to be failing to reasons outside of this PR.

lib/oxttl/src/toolkit/lexer.rs
```diff
@@ -155,19 +158,41 @@ impl<R: TokenRecognizer> Lexer<Vec<u8>, R> {
...
+            self.min_buffer_size
+        } else {
+            // Otherwise, we double the buffer size. If the
+            // size of the buffer would have caused an overflow,
+            // we fall back to the maximal buffer size and
+            // take the minimum
+            self.data
+                .len()
+                .checked_mul(2)
+                .map_or(self.max_buffer_size, |len| min(len, self.max_buffer_size))
```
Tpt: We might still end-up with a small buffer here is `self.data.len() < self.min_buffer_size / 2`. What about using something like (sorry, ugly formula):
```rust
min(
    self.max_buffer_size,
    max(
        self.data.len() + self.min_buffer_size,
        self.data.len().saturating_mul(2)
    )
)
```
This way we ensure we can read at least `min_buffer_size` bytes and that the buffer is not larger than `max_buffer_size`

lib/oxttl/src/toolkit/lexer.rs
```diff
@@ -155,19 +158,41 @@ impl<R: TokenRecognizer> Lexer<Vec<u8>, R> {
...
-        }
-        let read = reader.read(&mut self.data[new_start..]).await?;
-        self.data.truncate(new_start + read);
-        self.is_ending = read == 0;
+        // The current buffer size is the lower bound
+        let lower_bound = self.data.len();
+
+        let upper_bound = self.resized_buffer_len();
+        // Fill the buffer until the upper bound with 0s
+        self.data.resize(upper_bound, 0);
```
Tpt, line 171: hyper-nit: I would keep:
``rust
        if self.data.len() < self.data.capacity() {
            // We keep extending to have as much space as available without reallocation
            self.data.resize(self.data.capacity(), 0);
        }
```
because `upper_bound` is not always a multiple of two so we might have some allocated but not accessible memory.

File: lib/oxttl/src/toolkit/lexer.rs (hunk 8 above)
Comment (C-Loftus) [reply lib/oxttl/src/toolkit/lexer.rs]: Addressed, thank you. Thinking out loud (correct me if I am thinking wrong but feel free to ignore if not), I guess I didn't realize that we always needed at least `min_buffer_size` more in this context since at least as the buffer isn't 0 it will still be able to grow, but thinking about it more, it makes sense since we don't want to have a buffer of size 1 for instance and need to be doubling that over and over to get up to the

File: lib/oxttl/src/toolkit/lexer.rs (hunk 9 above)
Comment (C-Loftus) [reply lib/oxttl/src/toolkit/lexer.rs:171]: Hmmm could you provide a bit of extra context on this? I inserted that logic but it causes a few test errors. I must be misunderstanding your ask.

```rs
        // The current buffer size is the lower bound
        let lower_bound = self.data.len();

        let upper_bound = self.resized_buffer_len();
        // Fill the buffer until the upper bound with 0s
        if self.data.len() < self.data.capacity() {
            self.data.resize(upper_bound, 0);
        }
```


```
failures:

---- test_load_graph stdout ----

thread 'test_load_graph' (274278) panicked at lib/oxigraph/tests/store.rs:135:9:
assertion failed: store.contains(&q)?

---- test_load_dataset stdout ----

thread 'test_load_dataset' (274277) panicked at lib/oxigraph/tests/store.rs:260:9:
assertion failed: store.contains(&q)?
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test_load_graph_generates_new_blank_nodes stdout ----

thread 'test_load_graph_generates_new_blank_nodes' (274279) panicked at lib/oxigraph/tests/store.rs:291:5:
assertion `left == right` failed
  left: 0
 right: 2

---- test_load_graph_on_disk stdout ----

thread 'test_load_graph_on_disk' (274280) panicked at lib/oxigraph/tests/store.rs:148:9:
assertion failed: store.contains(&q)?

---- test_load_graph_on_disk_with_options stdout ----

thread 'test_load_graph_on_disk_with_options' (274281) panicked at lib/oxigraph/tests/store.rs:166:9:
assertion failed: store.contains(&q)?


failures:
    test_load_dataset
    test_load_graph
    test_load_graph_generates_new_blank_nodes
    test_load_graph_on_disk
    test_load_graph_on_disk_with_options

test result: FAILED. 19 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

I think the issue is that when `self.data.len() == self.data.capacity()` then the resize is not triggered and the lower bound is equivalent to the end (i.e. `lower_bound.. == self.data.len()`). The tests seem to indicate that this is not reading enough data although I am slightly confused at the same time since I thought this would signal a termination condition to not resize further. 

Let me know what you would like on this, thanks!

Tpt: Yes, we indeed need to keep the first resize to make sure the buffer is at least of `upper_bound`. The other case should be done later.

We need something like:


```
        // The current buffer size is the lower bound
        let lower_bound = self.data.len();

        // Fill the buffer until the upper bound with 0s
        let upper_bound = self.resized_buffer_len();
        self.data.resize(upper_bound, 0);

        // We keep extending to have as much space as available without reallocation
        if self.data.len() < self.data.capacity() {
            self.data.resize(upper_bound, 0);
        }
```

Thank you! Sorry for all these iterations.

C-Loftus: Ahh ok that makes sense. Done. All good, thanks for your patience!

Tpt:
Thank you! Sadly there is a conflict. May you do a rebase :pleading_face:

C-Loftus:
> Thank you! Sadly there is a conflict. May you do a rebase 🥺

Hmmm I did not see any merge conflicts on my end. That being said, I did pull it all the changes from upstream. Hopefully that should get all the latest code. Let me know if that looks reasonable (Wasn't sure if you want git history looking different or if you will squash)

Tpt:
Thank you! I will squash on merge.

testsuite/rdf-tests
Tpt, line 1: you reverted this submodule update (and the other ones). This might break CI.

lib/oxttl/src/lib.rs
```diff
@@ -23,6 +23,8 @@ pub use crate::trig::{TriGParser, TriGSerializer};
 pub use crate::turtle::{TurtleParser, TurtleSerializer};
 
 pub(crate) const MIN_BUFFER_SIZE: usize = 4096;
-pub(crate) const MAX_BUFFER_SIZE: usize = 4096 * 4096;
+/// Represents a reasonable default max buffer size in bytes for parsing
+/// rdf data
+pub const DEFAULT_MAX_BUFFER_SIZE: usize = 4096 * 4096;
```
Tpt: Sorry to revisit this topic: I don't think we should make this public. I would rephrase the doc comment to not state the default value. I would be something like:

```
    /// The default is set conservatively, use this function to change it (e.g. to [`usize::MAX`] to not set an upper bound).
```

WDYT?

C-Loftus: Happy to defer to you. I have used your wording and made it private outside of the crate again.

testsuite/rdf-tests
C-Loftus: Hmmm sorry I am not exactly sure what went wrong here. I did not run any submodule commands. I don't often use submodules so I must have done something wrong. I just ran `git pull` and pushed the merge commits. Sorry, could you advise on what I should be doing to get the submodules in the proper state again?  I tried using `git restore --source=HEAD --staged --worktree testsuite` which did not seem to help.

Tpt: Not sure. Do not bother with it, I will fix it after merging.

Tpt:
Thank you for the iterations! Squash in #1757

Tpt:
Merged! Thank you for all this work!

C-Loftus:
Thank you very much for your guidance and feedback. It is much appreciated and look forward to contributing again in the future 👍

Tpt:
Thnak you! Would be amazing

PR #1756, opened 2026-06-09
OxStr: adds an enum to encode the container type (owned vs borrowed)

Tpt:


lib/oxrdf/src/string.rs
```diff
@@ -7,12 +7,15 @@ use std::borrow::{Borrow, Cow};
...
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
+use std::mem::transmute;
 use std::ops::Deref;
 use std::process::abort;
 use std::ptr::NonNull;
 use std::sync::atomic::{AtomicUsize, Ordering, fence};
 
 const MAX_REF_COUNTER: usize = isize::MAX as usize;
+const KIND_SHIFT: u32 = usize::BITS - 1;
```
adamreichold, line 17: Just out of curiosity if you have the time to answer: Why the bit shift instead of the mask? Shouldn't both compile down to single (bit test) instruction, at least on x86-64?

Tpt: Great question! Mostly because I do the shift in the two directions in order to build my enum. But Indeed I could have set the enum variants value to the shifted values and have a mask instead.

This enum is to leave the room to further experiments around inlining small strings (can be useful for language tags) and maybe have references to owned value with a `into_owned()` method that bump there counter instead of cloning the buffer.

adamreichold: Thanks!

> This enum is to leave the room to further experiments around inlining small strings (can be useful for language tags)

This sounds promising. As a potentially relevant piece of anecdata, `string_cache` changed their code to prefer inline over interned storage for performance reasons, c.f. https://github.com/servo/string-cache/pull/278

Tpt: Thanks! Good to know

PR #1760, opened 2026-06-11
Store: Ensure that `Transaction` is `Send`

### LINKED ISSUE #1759: Transactions are `!Send`, impeding async use
Currently (as of 0.5.8, and still as of de4dc5f04379109aed08bfdff587235a647fb9b5), `oxigraph::store::Transaction` is `!Send`, meaning it cannot be sent between threads safely.

I figure this is likely unintentional, and originates from a [`std::sync::MutexGuard` field](https://github.com/oxigraph/oxigraph/blob/v0.5.8/lib/oxigraph/src/storage/memory.rs#L421) (still there on [main](https://github.com/oxigraph/oxigraph/blob/main/lib/oxigraph/src/storage/memory.rs#L420)) included in the `MemoryStorageTransaction` struct which then contaminates everything, all the way up to the top-level `Transaction`, with `!Send`.

The net effect is to considerably impede using `Store`/`Transaction` in async-heavy code bases. The minimal reproduction is the following, below, which just tries to move a transaction handle to another thread. The resulting compiler output, further below, is pretty clear.

The standard library's `MutexGuard` is [always](https://users.rust-lang.org/t/how-can-a-mutexguard-possibly-be-sync/108263) `!Send`, for cross-platform conservatism. The quickest fix here might be to change to using [parking_lot's](https://docs.rs/parking_lot) functionally equivalent [`Send`-enabled](https://docs.rs/crate/parking_lot/latest/features#send_guard) guards.

# Minimal Example

```oxistore.rs```:

```rust
#!/usr/bin/env rust-script
//! ```cargo
//! [dependencies]
//! oxigraph = "0.5.8"
//! ```

use oxigraph::store::Store;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let store = Store::new()?;

    let tx = store.start_transaction()?;

    let handle = std::thread::spawn(move || {
        eprintln!("is_empty={:?}", tx.is_empty().unwrap());
    });
    handle.join().unwrap();

    Ok(())
}
```

# Compiler Errors

<details>
<summary>Expand the compiler output</summary>

Rust 1.93.0:

```console
$ rust-script oxistore.rs
error[E0277]: `std::sync::MutexGuard<'_, usize>` cannot be sent between threads safely
    --> oxistore.rs:14:37
     |
  14 |       let handle = std::thread::spawn(move || {
     |                    ------------------ ^------
     |                    |                  |
     |  __________________|__________________within this `{closure@oxistore.rs:14:37: 14:44}`
     | |                  |
     | |                  required by a bound introduced by this call
  15 | |         eprintln!("is_empty={:?}", tx.is_empty().unwrap());
  16 | |     });
     | |_____^ `std::sync::MutexGuard<'_, usize>` cannot be sent between threads safely
     |
     = help: within `{closure@oxistore.rs:14:37: 14:44}`, the trait `Send` is not implemented for `std::sync::MutexGuard<'_, usize>`
note: required because it appears within the type `oxigraph::storage::memory::MemoryStorageTransaction<'_>`
    --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/oxigraph-0.5.8/src/storage/memory.rs:416:12
     |
 416 | pub struct MemoryStorageTransaction<'a> {
     |            ^^^^^^^^^^^^^^^^^^^^^^^^
note: required because it appears within the type `oxigraph::storage::StorageReadableTransactionKind<'_>`
    --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/oxigraph-0.5.8/src/storage/mod.rs:449:6
     |
 449 | enum StorageReadableTransactionKind<'a> {
     |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: required because it appears within the type `oxigraph::storage::StorageReadableTransaction<'_>`
    --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/oxigraph-0.5.8/src/storage/mod.rs:445:12
     |
 445 | pub struct StorageReadableTransaction<'a> {
     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
note: required because it appears within the type `Transaction<'_>`
    --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/oxigraph-0.5.8/src/store.rs:1127:12
     |
1127 | pub struct Transaction<'a> {
     |            ^^^^^^^^^^^
note: required because it's used within this closure
    --> oxistore.rs:14:37
     |
  14 |     let handle = std::thread::spawn(move || {
     |                                     ^^^^^^^
note: required by a bound in `spawn`
    --> ~/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/thread/functions.rs:128:8
     |
 125 | pub fn spawn<F, T>(f: F) -> JoinHandle<T>
     |        ----- required by a bound in this function
...
 128 |     F: Send + 'static,
     |        ^^^^ required by this bound in `spawn`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `main_f312a0f25950e8d05e5b4f06` (bin "main_f312a0f25950e8d05e5b4f06") due to 1 previous error
error: Could not execute cargo
```
</details>

artob:
Fixes #1759


Tpt:
Thank you for this! The transaction does not actually need to access the data guarded by the mutex so we can just use a `Mutex<bool>` and a `Condvar` to lock at the beginning of the transaction and release when it's dropped. I have just implemented it in #1761 This enables to avoid an extra dependency and also makes the `Transaction` object `Sync`. What do you think?

artob:
@Tpt Makes sense, though note that `parking_lot` is a transitive dependency in any case. Anyway, will close this in favor of #1761

PR #1761, opened 2026-06-11
Store: make Transaction and BulkLoader Send and Sync

Tpt:
Use a Mutex + Condvar combo to avoid keeping the Mutex guard during the transaction


artob (approved):

artob:
This resolves #1759 once backported to and released from the 0.5 branch

PR #1787, opened 2026-06-29
SPARQL parser: guess token array size

Tpt:


Tpt:
not really significant perf improvement and maybe significant memory consumption increase, closing

PR #1803, opened 2026-07-07
sparopt: expose join/variable-elimination order via GraphPattern::join_order_variables

eljeffeg:
## What

Adds a new public method, `GraphPattern::join_order_variables()`, to `sparopt::algebra::GraphPattern`. It returns the variables used by a pattern in left-to-right, depth-first order (deduplicated, keeping the first occurrence).

## Why

`Optimizer::optimize_graph_pattern` already performs join reordering (`reorder_joins`) based on structural/statistical heuristics, producing a left-deep nested `Join` tree. However, the resulting join order is only *implicit* in the shape of that tree — there's currently no public way to read it back out as an ordered list of variables.

Execution engines that implement their own join algorithm — for example worst-case-optimal join / Leapfrog Triejoin (LFTJ) executors, which need a variable-elimination order (VEO) — would like to use sparopt's structural join-ordering decision as an initial seed or tie-breaker for their own ordering, rather than ignoring it and recomputing an order from scratch. This method exposes exactly that.

Because `Join { left, right, .. }` nodes are always visited `left` before `right` in the existing `lookup_used_variables` traversal, calling `join_order_variables()` on the *output* of `Optimizer::optimize_graph_pattern` yields variables in the order the optimizer chose to introduce them.

## Implementation notes

- Purely additive: no existing public API, struct, or behavior is changed.
- Reuses the existing `lookup_used_variables` traversal internally (just wraps it with a `HashSet` for dedup while preserving first-occurrence order), so there's no risk of divergent traversal logic or maintenance burden from a second walk implementation.
- Added a `#[cfg(test)] mod tests` block in `algebra.rs` (previously the crate had no test infrastructure) covering:
  - left-to-right join order over a left-deep `Join` tree,
  - deduplication when a variable is shared across both sides of a join,
  - the empty/variable-free pattern case.
- Added a `CHANGELOG.md` entry under a new `[Unreleased]` section.

## Testing

```
cargo test -p sparopt --all-features   # 3 passed
cargo test -p sparopt                  # 3 passed (default features)
cargo clippy -p sparopt --all-features # clean (pre-existing unrelated lint-name warning only)
cargo fmt -p sparopt -- --check        # clean
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# [Unreleased]
+
+### Added
+- `sparopt`: `GraphPattern::join_order_variables`, exposing the join / variable-elimination order chosen by `Optimizer::optimize_graph_pattern` for consumption by external execution engines (e.g. worst-case-optimal join executors).
+
+
 # [0.5.7] - 2026-04-19
 
 ### Changed
--- a/lib/sparopt/src/algebra.rs
+++ b/lib/sparopt/src/algebra.rs
@@ -1112,6 +1112,38 @@ impl GraphPattern {
         }
     }
 
+    /// Returns the variables used by this pattern, in the order they are
+    /// first encountered during a left-to-right, depth-first walk of the
+    /// pattern tree (duplicates removed, keeping the first occurrence).
+    ///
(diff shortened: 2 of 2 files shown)


Tpt:
Thank you!

PR #1804, opened 2026-07-08
sparopt: fix GRAPH variable scoping bug in MINUS

eljeffeg:
## Summary

Fixes a GRAPH variable scoping bug in `sparopt` that caused `MINUS` to incorrectly use the ambient `GRAPH ?g { ... }` variable as an implicit join key between its two branches, breaking the SPARQL disjointness semantics that `MINUS` requires.

This makes the W3C SPARQL 1.1 evaluation test **"outer GRAPH operator does not affect MINUS disjointness"** (`negation/manifest#graph-minus`) pass.

## Root cause

`GraphPattern::from_sparql_algebra()` threads an ambient `graph_name: Option<&NamedNodePattern>` parameter through the recursive conversion from `spargebra`'s algebra to `sparopt`'s algebra, in order to "dissolve" `GRAPH ?g { ... }` wrappers into the `graph_name` field of leaf `QuadPattern`/`Path` nodes.

For `MINUS { left, right }`, the *same* `graph_name` value (including the same variable, when it's a variable) was passed down into **both** `left` and `right`. This is incorrect: per the SPARQL evaluation semantics for `MINUS`, the outer `GRAPH` operator's variable is not supposed to be considered when checking whether the two operands are disjoint. Sharing the same variable in both branches effectively turns it into a natural join/equality key between the branches, which defeats the disjointness check and produces wrong results.

## Fix

When converting `MINUS`, if the ambient `graph_name` is a variable, the right-hand branch is given a **fresh** variable instead of reusing the same one as the left-hand branch. This severs the accidental correlation between the two branches while still correctly scoping each side's pattern to the appropriate (dynamically-bound) graph.

```rust
AlGraphPattern::Minus { left, right } => {
    let right_graph_name = if let Some(NamedNodePattern::Variable(_)) = graph_name {
        Some(NamedNodePattern::Variable(new_var()))
    } else {
        graph_name.cloned()
    };
    Self::Minus {
        left: Box::new(Self::from_sparql_algebra(left, graph_name, blank_nodes)),
        right: Box::new(Self::from_sparql_algebra(
            right,
            right_graph_name.as_ref(),
            blank_nodes,
        )),
        algorithm: MinusAlgorithm::default(),
    }
}
```

## Testing

Verified against the live W3C SPARQL 1.1 query evaluation testsuite:

```
cargo test -p oxigraph-testsuite --test sparql sparql11_query_w3c_evaluation_testsuite -- --nocapture
```

- `negation/manifest#graph-minus` now passes (previously in the skip-list).
- No other tests regress; full `sparql.rs` test file (8 suites) passes.

## Note on related (separate) issue

While investigating this, I found two related skip-list entries that look superficially similar (also `GRAPH`-scoping related) but turned out to be a **different, unrelated root cause**:

- `aggregates/manifest#agg-empty-group-count-graph`
- `bindings/manifest#graph`

Both of these depend on an *empty* (zero-triple) named graph (e.g. `empty.ttl`) being visible during query evaluation (e.g. `SELECT ?g WHERE { GRAPH ?g {} }` needs to enumerate it). I confirmed via a standalone `spareval`/`oxrdf` repro (bypassing `sparopt` entirely) that `oxrdf::Dataset` has no way to represent a named graph that contains zero quads — it derives the set of existing named graphs purely by scanning stored quads (see `QueryableDataset::internal_named_graphs()`'s default impl). This is a structural limitation of `oxrdf::Dataset` itself, not a `sparopt` algebra bug, and is out of scope for this PR. I've left those two entries in the skip-list with an updated comment, and will file a separate issue to track it.

### DIFF
--- a/lib/sparopt/src/algebra.rs
+++ b/lib/sparopt/src/algebra.rs
@@ -1199,11 +1199,32 @@ impl GraphPattern {
                 expression: Expression::from_sparql_algebra(expression, graph_name),
                 variable: variable.clone(),
             },
-            AlGraphPattern::Minus { left, right } => Self::Minus {
-                left: Box::new(Self::from_sparql_algebra(left, graph_name, blank_nodes)),
-                right: Box::new(Self::from_sparql_algebra(right, graph_name, blank_nodes)),
-                algorithm: MinusAlgorithm::default(),
-            },
+            AlGraphPattern::Minus { left, right } => {
+                // The right-hand side of MINUS must not structurally share the
+                // ambient GRAPH variable with the left-hand side: per SPARQL
+                // semantics, the outer GRAPH operator's variable is not
+                // supposed to be considered when checking whether the two
+                // sides of a MINUS are disjoint (see W3C test
+                // "outer GRAPH operator does not affect MINUS disjointness").
+                // Using the same variable for both sides' graph_name would
+                // make it a natural join key between the branches, which
+                // breaks that disjointness check. We give the right branch a
+                // fresh variable instead.
+                let right_graph_name = if let Some(NamedNodePattern::Variable(_)) = graph_name {
+                    Some(NamedNodePattern::Variable(new_var()))
+                } else {
+                    graph_name.cloned()
+                };
+                Self::Minus {
+                    left: Box::new(Self::from_sparql_algebra(left, graph_name, blank_nodes)),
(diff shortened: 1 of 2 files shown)


Tpt:
Thank you for this MR #1758 is fixing this issue and a bunch of other GRAPH-related ones by fixing wrong assumptions on how `GRAPH` must be processed.

PR #1812, opened 2026-07-10
fix: instantiate ground DELETE/INSERT templates once instead of once per WHERE solution

### LINKED ISSUE #1802: Inexplicably slow DELETE INSERT WHERE
**Describe the bug**
The following (atomic replacement) query takes about 14 seconds(!) to complete, consuming a full CPU core during that time:

```sparql
PREFIX pinballshop-nl: <urn:sk:pinballshop-nl:>
PREFIX pinball: <urn:skx:pinball:>

DELETE {
  pinballshop-nl:product-5901-12784-00 pinball:url ?value_pinball_url . 
  pinballshop-nl:product-5901-12784-00 pinball:name ?value_pinball_name . 
  pinballshop-nl:product-5901-12784-00 pinball:inStock ?value_pinball_inStock . 
  pinballshop-nl:product-5901-12784-00 pinball:partType ?value_pinball_partType . 
  pinballshop-nl:product-5901-12784-00 pinball:country ?value_pinball_country . 
  pinballshop-nl:product-5901-12784-00 pinball:shipsTo ?value_pinball_shipsTo . 
  pinballshop-nl:product-5901-12784-00 pinball:forMachine ?value_pinball_forMachine . 
  pinballshop-nl:product-5901-12784-00 pinball:partNumber ?value_pinball_partNumber
}
INSERT {
  pinballshop-nl:product-5901-12784-00
    pinball:url "https://www.pinballshop.nl/electronics-parts/dmd/dot-matrix-128-x-32-plasma-display-red/" ;
    pinball:name "Dot Matrix DMD 128 x 32 Plasma display - Red" ;
    pinball:inStock true ;
    pinball:partType "ball" ;
    pinball:country "NL" ;
    pinball:shipsTo "AF" ;
    pinball:shipsTo "AX" ;
    pinball:shipsTo "AL" ;
    pinball:shipsTo "DZ" ;
    pinball:shipsTo "AS" ;
    pinball:shipsTo "AD" ;
    pinball:shipsTo "AO" ;
    pinball:shipsTo "AI" ;
    pinball:shipsTo "AQ" ;
    pinball:shipsTo "AG" ;
    pinball:shipsTo "AR" ;
    pinball:shipsTo "AM" ;
    pinball:shipsTo "AW" ;
    pinball:shipsTo "AU" ;
    pinball:shipsTo "AT" ;
    pinball:shipsTo "AZ" ;
    pinball:shipsTo "BS" ;
    pinball:shipsTo "BH" ;
    pinball:shipsTo "BD" ;
    pinball:shipsTo "BB" ;
    pinball:shipsTo "BY" ;
    pinball:shipsTo "BE" ;
    pinball:shipsTo "BZ" ;
    pinball:shipsTo "BJ" ;
    pinball:shipsTo "BM" ;
    pinball:shipsTo "BT" ;
    pinball:shipsTo "BO" ;
    pinball:shipsTo "BQ" ;
    pinball:shipsTo "BA" ;
    pinball:shipsTo "BW" ;
    pinball:shipsTo "BV" ;
    pinball:shipsTo "BR" ;
    pinball:shipsTo "IO" ;
    pinball:shipsTo "BN" ;
    pinball:shipsTo "BG" ;
    pinball:shipsTo "BF" ;
    pinball:shipsTo "BI" ;
    pinball:shipsTo "KH" ;
    pinball:shipsTo "CM" ;
    pinball:shipsTo "CA" ;
    pinball:shipsTo "CV" ;
    pinball:shipsTo "KY" ;
    pinball:shipsTo "CF" ;
    pinball:shipsTo "TD" ;
    pinball:shipsTo "CL" ;
    pinball:shipsTo "CN" ;
    pinball:shipsTo "CX" ;
    pinball:shipsTo "CC" ;
    pinball:shipsTo "CO" ;
    pinball:shipsTo "KM" ;
    pinball:shipsTo "CG" ;
    pinball:shipsTo "CD" ;
    pinball:shipsTo "CK" ;
    pinball:shipsTo "CR" ;
    pinball:shipsTo "CI" ;
    pinball:shipsTo "HR" ;
    pinball:shipsTo "CU" ;
    pinball:shipsTo "CW" ;
    pinball:shipsTo "CY" ;
    pinball:shipsTo "CZ" ;
    pinball:shipsTo "DK" ;
    pinball:shipsTo "DJ" ;
    pinball:shipsTo "DM" ;
    pinball:shipsTo "DO" ;
    pinball:shipsTo "EC" ;
    pinball:shipsTo "EG" ;
    pinball:shipsTo "SV" ;
    pinball:shipsTo "GQ" ;
    pinball:shipsTo "ER" ;
    pinball:shipsTo "EE" ;
    pinball:shipsTo "SZ" ;
    pinball:shipsTo "ET" ;
    pinball:shipsTo "FK" ;
    pinball:shipsTo "FO" ;
    pinball:shipsTo "FJ" ;
    pinball:shipsTo "FI" ;
    pinball:shipsTo "FR" ;
    pinball:shipsTo "GF" ;
    pinball:shipsTo "PF" ;
    pinball:shipsTo "TF" ;
    pinball:shipsTo "GA" ;
    pinball:shipsTo "GM" ;
    pinball:shipsTo "GE" ;
    pinball:shipsTo "DE" ;
    pinball:shipsTo "GH" ;
    pinball:shipsTo "GI" ;
    pinball:shipsTo "GR" ;
    pinball:shipsTo "GL" ;
    pinball:shipsTo "GD" ;
    pinball:shipsTo "GP" ;
    pinball:shipsTo "GU" ;
    pinball:shipsTo "GT" ;
    pinball:shipsTo "GG" ;
    pinball:shipsTo "GN" ;
    pinball:shipsTo "GW" ;
    pinball:shipsTo "GY" ;
    pinball:shipsTo "HT" ;
    pinball:shipsTo "HM" ;
    pinball:shipsTo "VA" ;
    pinball:shipsTo "HN" ;
    pinball:shipsTo "HK" ;
    pinball:shipsTo "HU" ;
    pinball:shipsTo "IS" ;
    pinball:shipsTo "IN" ;
    pinball:shipsTo "ID" ;
    pinball:shipsTo "IR" ;
    pinball:shipsTo "IQ" ;
    pinball:shipsTo "IE" ;
    pinball:shipsTo "IM" ;
    pinball:shipsTo "IL" ;
    pinball:shipsTo "IT" ;
    pinball:shipsTo "JM" ;
    pinball:shipsTo "JP" ;
    pinball:shipsTo "JE" ;
    pinball:shipsTo "JO" ;
    pinball:shipsTo "KZ" ;
    pinball:shipsTo "KE" ;
    pinball:shipsTo "KI" ;
    pinball:shipsTo "KP" ;
    pinball:shipsTo "KR" ;
    pinball:shipsTo "KW" ;
    pinball:shipsTo "KG" ;
    pinball:shipsTo "LA" ;
    pinball:shipsTo "LV" ;
    pinball:shipsTo "LB" ;
    pinball:shipsTo "LS" ;
    pinball:shipsTo "LR" ;
    pinball:shipsTo "LY" ;
    pinball:shipsTo "LI" ;
    pinball:shipsTo "LT" ;
    pinball:shipsTo "LU" ;
    pinball:shipsTo "MO" ;
    pinball:shipsTo "MG" ;
    pinball:shipsTo "MW" ;
    pinball:shipsTo "MY" ;
    pinball:shipsTo "MV" ;
    pinball:shipsTo "ML" ;
    pinball:shipsTo "MT" ;
    pinball:shipsTo "MH" ;
    pinball:shipsTo "MQ" ;
    pinball:shipsTo "MR" ;
    pinball:shipsTo "MU" ;
    pinball:shipsTo "YT" ;
    pinball:shipsTo "MX" ;
    pinball:shipsTo "FM" ;
    pinball:shipsTo "MD" ;
    pinball:shipsTo "MC" ;
    pinball:shipsTo "MN" ;
    pinball:shipsTo "ME" ;
    pinball:shipsTo "MS" ;
    pinball:shipsTo "MA" ;
    pinball:shipsTo "MZ" ;
    pinball:shipsTo "MM" ;
    pinball:shipsTo "NA" ;
    pinball:shipsTo "NR" ;
    pinball:shipsTo "NP" ;
    pinball:shipsTo "NL" ;
    pinball:shipsTo "NC" ;
    pinball:shipsTo "NZ" ;
    pinball:shipsTo "NI" ;
    pinball:shipsTo "NE" ;
    pinball:shipsTo "NG" ;
    pinball:shipsTo "NU" ;
    pinball:shipsTo "NF" ;
    pinball:shipsTo "MK" ;
    pinball:shipsTo "MP" ;
    pinball:shipsTo "NO" ;
    pinball:shipsTo "OM" ;
    pinball:shipsTo "PK" ;
    pinball:shipsTo "PW" ;
    pinball:shipsTo "PS" ;
    pinball:shipsTo "PA" ;
    pinball:shipsTo "PG" ;
    pinball:shipsTo "PY" ;
    pinball:shipsTo "PE" ;
    pinball:shipsTo "PH" ;
    pinball:shipsTo "PN" ;
    pinball:shipsTo "PL" ;
    pinball:shipsTo "PT" ;
    pinball:shipsTo "PR" ;
    pinball:shipsTo "QA" ;
    pinball:shipsTo "RE" ;
    pinball:shipsTo "RO" ;
    pinball:shipsTo "RU" ;
    pinball:shipsTo "RW" ;
    pinball:shipsTo "SH" ;
    pinball:shipsTo "BL" ;
    pinball:shipsTo "KN" ;
    pinball:shipsTo "LC" ;
    pinball:shipsTo "MF" ;
    pinball:shipsTo "PM" ;
    pinball:shipsTo "VC" ;
    pinball:shipsTo "WS" ;
    pinball:shipsTo "SM" ;
    pinball:shipsTo "ST" ;
    pinball:shipsTo "SA" ;
    pinball:shipsTo "SN" ;
    pinball:shipsTo "RS" ;
    pinball:shipsTo "SC" ;
    pinball:shipsTo "SL" ;
    pinball:shipsTo "SG" ;
    pinball:shipsTo "SX" ;
    pinball:shipsTo "SK" ;
    pinball:shipsTo "SI" ;
    pinball:shipsTo "SB" ;
    pinball:shipsTo "SO" ;
    pinball:shipsTo "ZA" ;
    pinball:shipsTo "GS" ;
    pinball:shipsTo "SS" ;
    pinball:shipsTo "ES" ;
    pinball:shipsTo "LK" ;
    pinball:shipsTo "SD" ;
    pinball:shipsTo "SR" ;
    pinball:shipsTo "SJ" ;
    pinball:shipsTo "SE" ;
    pinball:shipsTo "CH" ;
    pinball:shipsTo "SY" ;
    pinball:shipsTo "TW" ;
    pinball:shipsTo "TJ" ;
    pinball:shipsTo "TZ" ;
    pinball:shipsTo "TH" ;
    pinball:shipsTo "TL" ;
    pinball:shipsTo "TG" ;
    pinball:shipsTo "TK" ;
    pinball:shipsTo "TO" ;
    pinball:shipsTo "TT" ;
    pinball:shipsTo "TN" ;
    pinball:shipsTo "TR" ;
    pinball:shipsTo "TM" ;
    pinball:shipsTo "TC" ;
    pinball:shipsTo "TV" ;
    pinball:shipsTo "UG" ;
    pinball:shipsTo "UA" ;
    pinball:shipsTo "AE" ;
    pinball:shipsTo "GB" ;
    pinball:shipsTo "US" ;
    pinball:shipsTo "UM" ;
    pinball:shipsTo "UY" ;
    pinball:shipsTo "UZ" ;
    pinball:shipsTo "VU" ;
    pinball:shipsTo "VE" ;
    pinball:shipsTo "VN" ;
    pinball:shipsTo "VG" ;
    pinball:shipsTo "VI" ;
    pinball:shipsTo "WF" ;
    pinball:shipsTo "EH" ;
    pinball:shipsTo "YE" ;
    pinball:shipsTo "ZM" ;
    pinball:shipsTo "ZW" ;
    pinball:forMachine "Cactus Canyon" ;
    pinball:forMachine "Congo" ;
    pinball:forMachine "Dirty Harry" ;
    pinball:forMachine "Guns N\' Roses" ;
    pinball:forMachine "Hurricane" ;
    pinball:forMachine "Mario Andretti" ;
    pinball:forMachine "NBA Fastbreak" ;
    pinball:forMachine "Nascar" ;
    pinball:forMachine "Shrek" ;
    pinball:forMachine "Spider-Man" ;
    pinball:forMachine "Twilight Zone" ;
    pinball:forMachine "Twister" ;
    pinball:forMachine "White Water" ;
    pinball:forMachine "World Poker Tour" ;
    pinball:forMachine "High Roller Casino" ;
    pinball:partNumber "williams:5901-12784-00" ;
    pinball:partNumber "stern:520-5052-00" .
}
WHERE {
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:url ?value_pinball_url } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:name ?value_pinball_name } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:inStock ?value_pinball_inStock } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:partType ?value_pinball_partType } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:country ?value_pinball_country } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:shipsTo ?value_pinball_shipsTo } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:forMachine ?value_pinball_forMachine } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:partNumber ?value_pinball_partNumber } .
}
```

While there are admittedly many triples involved, it doesn't seem like the kind of amount that would explain such a slow query.

More interestingly, when removing the `INSERT` component from the query, it is much faster (190ms):

```sparql
PREFIX pinballshop-nl: <urn:sk:pinballshop-nl:>
PREFIX pinball: <urn:skx:pinball:>

DELETE {
  pinballshop-nl:product-5901-12784-00 pinball:url ?value_pinball_url . 
  pinballshop-nl:product-5901-12784-00 pinball:name ?value_pinball_name . 
  pinballshop-nl:product-5901-12784-00 pinball:inStock ?value_pinball_inStock . 
  pinballshop-nl:product-5901-12784-00 pinball:partType ?value_pinball_partType . 
  pinballshop-nl:product-5901-12784-00 pinball:country ?value_pinball_country . 
  pinballshop-nl:product-5901-12784-00 pinball:shipsTo ?value_pinball_shipsTo . 
  pinballshop-nl:product-5901-12784-00 pinball:forMachine ?value_pinball_forMachine . 
  pinballshop-nl:product-5901-12784-00 pinball:partNumber ?value_pinball_partNumber
}
WHERE {
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:url ?value_pinball_url } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:name ?value_pinball_name } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:inStock ?value_pinball_inStock } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:partType ?value_pinball_partType } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:country ?value_pinball_country } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:shipsTo ?value_pinball_shipsTo } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:forMachine ?value_pinball_forMachine } .
  OPTIONAL { pinballshop-nl:product-5901-12784-00 pinball:partNumber ?value_pinball_partNumber } .
}
```

I initially thought this meant that the slowness came from the `INSERT` component. But this hypothesis was then immediately defeated when I ran the original query *with* the `INSERT` component again - which, this time, completed in 17ms(!). But *yet another* run of the original query, without being preceded by the `INSERT`-less version of the query, got me right back to 13 seconds.

In conclusion, it appears that this `DELETE/INSERT/WHERE` pattern is slow *if and only if* there is some amount of matches to be deleted prior to the insertion; but if the deletion phase is done separately first, then the total query time of both the preparation query *and* the original query is an order of magnitude lower than the original query alone would've been.

I've done some further testing by doing a run of the full query, and then doing repeated runs of the same query but with ~10 `shipsTo` entries removed every time, which (accounting for timing jitter) seems to show __exponential__ performance changes; the sequence of timings is below, ordered from most to least `shipsTo` entries:

```
13.5s
11.0s
10.6s
9.2s
8.9s
9.0s
7.0s
7.3s
5.5s
5.5s
4.6s
4.7s
3.8s
3.0s
2.6s
2.3s
1.9s
1.4s
1.2s
0.9s
0.7s
0.5s
0.4s
0.2s
0.1s
```

When re-running the full query with all `shipsTo` entries after this test (ie. a very small amount of `shipsTo` entries remained in the database, but not zero), it completed in 61ms but a second run took 14.3 seconds again, which suggests that the performance of the query scales with the amount of entries already in the database, not the amount in the query.

The state of the database is the same whether the full query is run at once (slow) or whether it's preceded by a `DELETE WHERE`-only query first (fast), so it seems that the performance issue lies specifically in some interaction between a) a large amount of matching entries in the database, and b) an `INSERT` clause being present in the `DELETE WHERE`. Is it perhaps possible that a join occurs internally between the `INSERT` and the `DELETE WHERE` despite there being no actual reference between them?

**To Reproduce**

Oxigraph version 0.5.7 from packages on NixOS 26.05. Above query should reproduce the bug, though I haven't tested it on an empty database (my current database folder is about 1.6GB).

mvanhorn:
## Summary
In `DeleteInsertIter` (lib/spareval/src/update.rs), partition the `delete` and `insert` template lists at construction time into ground templates (no variables anywhere, checked recursively through sparql-12 nested triple patterns; for INSERT templates also no blank nodes, since blank nodes must be re-instantiated freshly per solution) and variable-dependent templates. In `Iterator::next`, when the first solution arrives, emit each ground template exactly once (pre-converted to a concrete `Quad`) in addition to that solution's variable-dependent fills; subsequent solutions only fill the variable-dependent templates. Zero solutions still emit nothing, preserving SPARQL Update semantics exactly (identical final store state, since duplicate quad inserts/deletes are set-semantics no-ops).

## Why this matters
A `DELETE ... INSERT ... WHERE` update whose INSERT clause contains only constant (fully ground) triples takes ~14 seconds on a store where the WHERE clause matches many rows, while the equivalent DELETE-only query plus a separate INSERT completes in ~200ms. The reporter's WHERE clause is 8 unrelated OPTIONALs that cross-join into thousands of solutions, and the ~250 constant INSERT triples are re-instantiated and re-inserted once per solution, producing millions of redundant (no-op after the first) insert operations; timings scale with the number of matching rows already in the database.

See https://github.com/oxigraph/oxigraph/issues/1802.

## Testing
- Happy path: DELETE (variable patterns) + INSERT (all-ground triples) + WHERE producing N>1 solutions -> each ground insert quad emitted exactly once, per-solution deletes still emitted for every solution, final store state identical to before the change. - WHERE produces zero solutions -> ground templates emit nothing (no insert happens). - INSERT template containing a blank node -> NOT hoisted; a fresh blank node is instantiated per solution (spec behavior preserved). - Mixed INSERT template (some ground triples, some with variables) -> ground ones emitted once, variable ones once per solution.

Fixes #1802

### DIFF
--- a/.osc-metadata/sync.json
+++ b/.osc-metadata/sync.json
@@ -1,0 +1,5 @@
+{
+  "fork_synced_at": "2026-07-10T09:52:32.116353+00:00",
+  "commits_behind_before_sync": 0,
+  "action_taken": "noop"
+}
--- a/.osc-metadata/sync.json
+++ b/.osc-metadata/sync.json
@@ -1,5 +1,0 @@
-{
-  "fork_synced_at": "2026-07-10T09:52:32.116353+00:00",
-  "commits_behind_before_sync": 0,
-  "action_taken": "noop"
-}
--- a/lib/oxigraph/tests/store.rs
+++ b/lib/oxigraph/tests/store.rs
@@ -4,7 +4,6 @@
 use oxigraph::io::RdfFormat;
 use oxigraph::model::vocab::{rdf, xsd};
 use oxigraph::model::*;
-use oxigraph::sparql::SparqlEvaluator;
 use oxigraph::store::Store;
 #[cfg(all(not(target_family = "wasm"), feature = "rocksdb"))]
 use oxigraph::store::StoreOptions;
@@ -290,83 +289,6 @@ fn test_load_graph_generates_new_blank_nodes() -> Result<(), Box<dyn Error>> {
         )?;
     }
     assert_eq!(store.len()?, 2);
-    Ok(())
-}
-
-#[test]
-fn test_update_ground_insert_with_multi_solution_where() -> Result<(), Box<dyn Error>> {
-    let row1 = NamedNode::new_unchecked("http://example.com/row1");
-    let row2 = NamedNode::new_unchecked("http://example.com/row2");
-    let match_predicate = NamedNode::new_unchecked("http://example.com/match");
-    let old_predicate = NamedNode::new_unchecked("http://example.com/old");
-    let inserted_predicate = NamedNode::new_unchecked("http://example.com/inserted");
-    let row1_old = Quad::new(
-        row1.clone(),
-        old_predicate.clone(),
-        Literal::new_simple_literal("old1"),
-        GraphName::DefaultGraph,
-    );
-    let row2_old = Quad::new(
-        row2.clone(),
-        old_predicate,
-        Literal::new_simple_literal("old2"),
-        GraphName::DefaultGraph,
-    );
-    let inserted1 = Quad::new(
-        NamedNode::new_unchecked("http://example.com/inserted1"),
-        inserted_predicate.clone(),
-        Literal::new_simple_literal("value1"),
-        GraphName::DefaultGraph,
-    );
-    let inserted2 = Quad::new(
-        NamedNode::new_unchecked("http://example.com/inserted2"),
-        inserted_predicate,
-        Literal::new_simple_literal("value2"),
-        GraphName::DefaultGraph,
-    );
-    let store = Store::new()?;
-
-    store.insert(Quad::new(
-        row1.clone(),
-        match_predicate.clone(),
-        Literal::from(1),
-        GraphName::DefaultGraph,
-    ))?;
-    store.insert(row1_old.clone())?;
-    store.insert(Quad::new(
-        row2.clone(),
-        match_predicate,
-        Literal::from(2),
-        GraphName::DefaultGraph,
-    ))?;
-    store.insert(row2_old.clone())?;
-
-    SparqlEvaluator::new()
-        .parse_update(
-            r#"
-            PREFIX ex: <http://example.com/>
-            DELETE {
-                ?s ex:old ?old .
-            }
-            INSERT {
-                ex:inserted1 ex:inserted "value1" .
-                ex:inserted2 ex:inserted "value2" .
-            }
-            WHERE {
-                ?s ex:match ?value .
-                ?s ex:old ?old .
-            }
-            "#,
-        )?
-        .on_store(&store)
-        .execute()?;
-
-    assert!(!store.contains(&row1_old)?);
-    assert!(!store.contains(&row2_old)?);
-    assert!(store.contains(&inserted1)?);
-    assert!(store.contains(&inserted2)?);
-    assert_eq!(store.len()?, 4);
-    store.validate()?;
     Ok(())
 }
 
--- a/lib/oxigraph/tests/store.rs
+++ b/lib/oxigraph/tests/store.rs
@@ -4,6 +4,7 @@
 use oxigraph::io::RdfFormat;
 use oxigraph::model::vocab::{rdf, xsd};
 use oxigraph::model::*;
+use oxigraph::sparql::SparqlEvaluator;
 use oxigraph::store::Store;
 #[cfg(all(not(target_family = "wasm"), feature = "rocksdb"))]
 use oxigraph::store::StoreOptions;
@@ -289,6 +290,83 @@ fn test_load_graph_generates_new_blank_nodes() -> Result<(), Box<dyn Error>> {
         )?;
     }
     assert_eq!(store.len()?, 2);
+    Ok(())
+}
+
+#[test]
+fn test_update_ground_insert_with_multi_solution_where() -> Result<(), Box<dyn Error>> {
+    let row1 = NamedNode::new_unchecked("http://example.com/row1");
+    let row2 = NamedNode::new_unchecked("http://example.com/row2");
+    let match_predicate = NamedNode::new_unchecked("http://example.com/match");
+    let old_predicate = NamedNode::new_unchecked("http://example.com/old");
+    let inserted_predicate = NamedNode::new_unchecked("http://example.com/inserted");
+    let row1_old = Quad::new(
+        row1.clone(),
+        old_predicate.clone(),
+        Literal::new_simple_literal("old1"),
+        GraphName::DefaultGraph,
+    );
+    let row2_old = Quad::new(
+        row2.clone(),
+        old_predicate,
+        Literal::new_simple_literal("old2"),
+        GraphName::DefaultGraph,
+    );
+    let inserted1 = Quad::new(
+        NamedNode::new_unchecked("http://example.com/inserted1"),
+        inserted_predicate.clone(),
+        Literal::new_simple_literal("value1"),
+        GraphName::DefaultGraph,
+    );
+    let inserted2 = Quad::new(
+        NamedNode::new_unchecked("http://example.com/inserted2"),
+        inserted_predicate,
+        Literal::new_simple_literal("value2"),
+        GraphName::DefaultGraph,
+    );
+    let store = Store::new()?;
+
+    store.insert(Quad::new(
+        row1.clone(),
+        match_predicate.clone(),
+        Literal::from(1),
+        GraphName::DefaultGraph,
+    ))?;
+    store.insert(row1_old.clone())?;
+    store.insert(Quad::new(
+        row2.clone(),
+        match_predicate,
+        Literal::from(2),
+        GraphName::DefaultGraph,
+    ))?;
+    store.insert(row2_old.clone())?;
+
+    SparqlEvaluator::new()
+        .parse_update(
+            r#"
+            PREFIX ex: <http://example.com/>
+            DELETE {
+                ?s ex:old ?old .
+            }
+            INSERT {
+                ex:inserted1 ex:inserted "value1" .
+                ex:inserted2 ex:inserted "value2" .
+            }
+            WHERE {
+                ?s ex:match ?value .
+                ?s ex:old ?old .
+            }
+            "#,
+        )?
+        .on_store(&store)
+        .execute()?;
+
+    assert!(!store.contains(&row1_old)?);
+    assert!(!store.contains(&row2_old)?);
+    assert!(store.contains(&inserted1)?);
+    assert!(store.contains(&inserted2)?);
+    assert_eq!(store.len()?, 4);
+    store.validate()?;
     Ok(())
 }
 
--- a/lib/spareval/src/update.rs
+++ b/lib/spareval/src/update.rs
@@ -11,6 +11,7 @@ use spargebra::term::{
 #[cfg(feature = "sparql-12")]
 use spargebra::term::{GroundTriplePattern, TriplePattern};
 use std::collections::VecDeque;
+use std::mem::take;
 
 /// A [`Quad`] to delete or insert.
 #[derive(Eq, PartialEq, Debug, Clone, Hash)]
@@ -28,9 +29,6 @@ pub struct DeleteInsertIter<'a, 'b> {
     variable_insert: Vec<&'b QuadPattern>,
     buffer: VecDeque<DeleteInsertQuad>,
     bnodes: FxHashMap<BlankNode, BlankNode>,
-    ground_deletes_emitted: bool,
-    ground_inserts_emitted: bool,
-    had_solution: bool,
 }
 
 impl<'a, 'b> DeleteInsertIter<'a, 'b> {
@@ -38,16 +36,17 @@ impl<'a, 'b> DeleteInsertIter<'a, 'b> {
         solutions: QuerySolutionIter<'a>,
         delete: &'b [GroundQuadPattern],
         insert: &'b [QuadPattern],
+        without_optimizations: bool,
     ) -> Self {
         let empty_solution =
             QuerySolution::from((Vec::<Variable>::new(), Vec::<Option<Term>>::new()));
         let mut ground_delete = Vec::new();
         let mut variable_delete = Vec::new();
         for quad in delete {
-            if ground_quad_pattern_has_variable(quad) {
+            if without_optimizations || ground_quad_pattern_has_variable(quad) {
                 variable_delete.push(quad);
-            } else if let Some(quad) = fill_ground_quad_pattern(quad, &empty_solution) {
-                ground_delete.push(quad);
+            } else {
+                ground_delete.extend(fill_ground_quad_pattern(quad, &empty_solution));
             }
         }
 
@@ -55,10 +54,10 @@ impl<'a, 'b> DeleteInsertIter<'a, 'b> {
         let mut variable_insert = Vec::new();
         let mut bnodes = FxHashMap::default();
         for quad in insert {
-            if insert_quad_pattern_needs_solution(quad) {
+            if without_optimizations || quad_pattern_needs_solution(quad) {
                 variable_insert.push(quad);
-            } else if let Some(quad) = fill_quad_pattern(quad, &empty_solution, &mut bnodes) {
-                ground_insert.push(quad);
+            } else {
+                ground_insert.extend(fill_quad_pattern(quad, &empty_solution, &mut bnodes));
             }
         }
 
@@ -70,9 +69,6 @@ impl<'a, 'b> DeleteInsertIter<'a, 'b> {
             variable_insert,
             buffer: VecDeque::new(),
             bnodes: FxHashMap::default(),
-            ground_deletes_emitted: false,
-            ground_inserts_emitted: false,
-            had_solution: false,
         }
     }
 }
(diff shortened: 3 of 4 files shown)


.osc-metadata/sync.json
Tpt, line 1: too aggressive `git add`?

lib/spareval/src/update.rs
```diff
@@ -34,12 +39,40 @@ impl<'a, 'b> DeleteInsertIter<'a, 'b> {
...
         insert: &'b [QuadPattern],
     ) -> Self {
+        let empty_solution =
+            QuerySolution::from((Vec::<Variable>::new(), Vec::<Option<Term>>::new()));
+        let mut ground_delete = Vec::new();
+        let mut variable_delete = Vec::new();
+        for quad in delete {
+            if ground_quad_pattern_has_variable(quad) {
+                variable_delete.push(quad);
+            } else if let Some(quad) = fill_ground_quad_pattern(quad, &empty_solution) {
```
Tpt: can just be an else
```suggestion
            } else {
```

lib/spareval/src/update.rs
```diff
@@ -34,12 +39,40 @@ impl<'a, 'b> DeleteInsertIter<'a, 'b> {
...
+            }
+        }
+
+        let mut ground_insert = Vec::new();
+        let mut variable_insert = Vec::new();
+        let mut bnodes = FxHashMap::default();
+        for quad in insert {
+            if insert_quad_pattern_needs_solution(quad) {
+                variable_insert.push(quad);
+            } else if let Some(quad) = fill_quad_pattern(quad, &empty_solution, &mut bnodes) {
```
Tpt: same
```suggestion
            } else {
```

lib/spareval/src/update.rs
```diff
@@ -52,16 +85,35 @@ impl Iterator for DeleteInsertIter<'_, '_> {
...
                 return Some(Ok(quad));
             }
-            let solution = match self.solutions.next()? {
-                Ok(solution) => solution,
-                Err(e) => return Some(Err(e)),
+            let solution = match self.solutions.next() {
+                Some(Ok(solution)) => solution,
+                Some(Err(e)) => return Some(Err(e)),
+                None => {
+                    if self.had_solution && !self.ground_inserts_emitted {
```
Tpt: what about moving this alongside the ground_delete case, it would avoid the need for `had_solution` boolean

lib/spareval/src/update.rs
```diff
@@ -52,16 +85,35 @@ impl Iterator for DeleteInsertIter<'_, '_> {
...
             };
-            for quad in self.delete {
+            self.had_solution = true;
+            if !self.ground_deletes_emitted {
+                for quad in &self.ground_delete {
+                    self.buffer
+                        .push_back(DeleteInsertQuad::Delete(quad.clone()));
+                }
+                self.ground_deletes_emitted = true;
+            }
```
Tpt: I don't think you need `ground_deletes_emitted`. Just emptying the `ground_delete` triples works
```suggestion
            for quad in take(&mut self.ground_delete) {
                self.buffer
                                .push_back(DeleteInsertQuad::Delete(quad));
```
(same for insertions)

lib/spareval/src/update.rs
```diff
@@ -71,6 +123,60 @@ impl Iterator for DeleteInsertIter<'_, '_> {
...
+}
+
+#[cfg(feature = "sparql-12")]
+fn ground_triple_pattern_has_variable(triple: &GroundTriplePattern) -> bool {
+    ground_term_pattern_has_variable(&triple.subject)
+        || named_node_pattern_has_variable(&triple.predicate)
+        || ground_term_pattern_has_variable(&triple.object)
+}
+
+fn insert_quad_pattern_needs_solution(quad: &QuadPattern) -> bool {
```
Tpt: hyper-nit: for consistency?
```suggestion
fn quad_pattern_needs_solution(quad: &QuadPattern) -> bool {
```
(same for `insert_term_pattern_needs_solution`)

Tpt (commented):
Thank you for this.

Due to the code verbosity, I guess it's LLM generated (apologies if I am wrong). Some ideas to make it less verbose to avoid exploding the codebase size.

lib/oxigraph/tests/store.rs
```diff
@@ -292,6 +293,83 @@ fn test_load_graph_generates_new_blank_nodes() -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
+#[test]
+fn test_update_ground_insert_with_multi_solution_where() -> Result<(), Box<dyn Error>> {
```
Tpt: we already have a bunch of SPARQL update tests, I am not sure this one is useful.

lib/spareval/src/update.rs
```diff
@@ -196,3 +302,254 @@ fn fill_ground_triple_pattern(
 fn convert_blank_node(node: &BlankNode, bnodes: &mut FxHashMap<BlankNode, BlankNode>) -> BlankNode {
     bnodes.entry(node.clone()).or_default().clone()
 }
+
+#[cfg(test)]
+mod tests {
```
Tpt, line 291: To be lazy, we might disable the "ground term" optimization if `QueryEvaluator::without_optimization` is set ([example](https://github.com/oxigraph/oxigraph/blob/8183c2efd6a3b6dfb8dc0bc1cd7cf754339249a7/lib/spareval/src/lib.rs#L601)). Then the update fuzzer will ensure that results are always the same with and without optimization, skipping the need to write explicit tests.

Tpt:
Aside: ignore the codspeed message, it seems to be just noise

mvanhorn:
Thanks, all applied: collapsed both arms to `else` with `.extend`, dropped `had_solution` and the `ground_*_emitted` flags in favour of draining `ground_delete`/`ground_insert` with `take()`, renamed to `quad_pattern_needs_solution`, and gated the ground-term hoisting on `without_optimizations` so the update fuzzer can compare optimized vs unoptimized. Also removed the store test you flagged and the stray `.osc-metadata/sync.json` (bad `git add` on my end).

One thing to sanity-check: draining ground inserts eagerly means they now emit before later solutions' variable deletes rather than being deferred to the end, so I dropped the `ground_insert_is_emitted_after_later_variable_deletes` test that pinned the old ordering. That's fine if the store applies the delete/insert sets as a batch, but let me know if emission order matters downstream and I'll restore the deferral.

Tpt:
thank you!

PR #1824, opened 2026-07-16
Unstable canonicalization: output blank node ids based on hashes and …

Tpt:
…not a counter

Also expose the blank node mapping to the user

### DIFF
--- a/lib/oxrdf/src/dataset.rs
+++ b/lib/oxrdf/src/dataset.rs
@@ -610,7 +610,9 @@ impl Dataset {
         algorithm: CanonicalizationAlgorithm,
     ) -> HashMap<InternedBlankNode, BlankNode> {
         let hash_algorithm = match algorithm {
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
Tested this branch against our LinkML use case (canonicalize → sorted serialize → diff). Two concrete results:

- **Stability:** isomorphic graphs with different blank-node ids produce byte-identical output. ✅
- **Change-locality:** 5 blank-node records, then inserting a 6th whose value sorts in the middle → exactly the 2 new triples appear in the diff; the 5 existing records are byte-unchanged. Under the old counter labels that middle insert would have renumbered everything after it. This is precisely the diff-locality we were missing. ✅

On the first-degree-collision drift I mentioned: in practice the existing node kept its id and the new one took the counter suffix, and even for genuinely automorphic nodes the *sorted serialization* is invariant to the tie-break — so it doesn't actually show up in diffs. Happy with this from our end.

PR #1825, opened 2026-07-17
Fixed JS sample code to use '.quad()'

pmcb55:
There is no '.triple()' method on the RDF/JS Data Model, so this sample JavaScript fails whenever anyone tries it. The fix is to use '.quad()' instead (the 'graph' parameter can be 'undefined', so nothing else needs to change in this sample code).


Tpt:
thanks!

PR #1858, opened 2026-08-09
Fix CLI load exit status on file errors

### LINKED ISSUE #1857: CLI file loader exits successfully after load failures
**Describe the bug**

When `oxigraph load --file` encounters a file open, RDF parsing, or bulk-loader commit error, the worker prints the error but the command still exits successfully. This differs from loading the same invalid input through stdin, where the error is propagated.

A successful exit can cause scripts and data pipelines to accept and publish a partially loaded store.

**To reproduce**

Tested on Oxigraph `main` at `db207116c1a3ad1e3b80498a21d4508dfaf83361`.

1. Create an invalid N-Triples file:

       printf 'not valid N-Triples\n' > invalid.nt

2. Run:

       cargo run --bin oxigraph --no-default-features -- load --location store --file invalid.nt

3. Observe an `Error while loading file ...` message followed by exit status 0.

Loading the same content through stdin returns a nonzero status.

**Expected behavior**

`oxigraph load --file` should return a nonzero exit status if any input file cannot be opened, parsed, or committed.

mattfaltyn:
## Summary

- return a failure from file-based CLI loads when any worker cannot open, parse, or commit a file
- preserve parallel loading and the existing per-file error diagnostics
- cover concurrent parse and gzip-open failures with a regression test

Closes #1857

## Testing

- `cargo fmt -- --check`
- `cargo clippy -p oxigraph-cli --all-targets --no-default-features -- -D warnings -D clippy::all`
- `cargo clippy -p oxigraph-cli --all-targets -- -D warnings -D clippy::all`
- `cargo test -p oxigraph-cli --no-default-features` (63 passed)
- `cargo test -p oxigraph-cli cli_load_file_errors_fail_the_command`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -35,6 +35,7 @@ use std::os::unix::net::UnixDatagram;
 use std::path::Path;
 use std::rc::Rc;
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread::available_parallelism;
 use std::time::{Duration, Instant};
 use std::{fmt, fs, str, thread};
@@ -153,6 +154,7 @@ pub fn main() -> anyhow::Result<()> {
                 )?;
                 loader.commit()?;
             } else {
+                let has_error = AtomicBool::new(false);
                 ThreadPoolBuilder::new()
                     .num_threads(max(1, available_parallelism()?.get() / 2))
                     .thread_name(|i| format!("Oxigraph bulk loader thread {i}"))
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
adamreichold, line 163: It might be preferable to use one of Rayon's combinators like [`try_for_each`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.try_for_each) here instead of manually spawning tasks. (Or `.map(..).collect::<Vec<Result<_>>>()` if all errors should be collected instead of best-effort short-circuiting.)

Tpt (approved):
Thank you!

PR #1860, opened 2026-08-09
RDF/XML: reject predicates without QName local parts

### LINKED ISSUE #1859: RDF/XML serializer emits malformed XML for predicates without a QName local part
**Describe the bug**

`RdfXmlSerializer` accepts predicate IRIs that cannot be represented as XML QNames and emits malformed XML. For example, a predicate ending in a numeric-only path segment produces an empty QName local part while serialization still succeeds.

**To Reproduce**

Tested on `main` at `94688a245115887dd5b10f1c33030502e087a2f5` on macOS:

```console
$ printf '<http://example.com/s> <http://example.com/123> <http://example.com/o> .\n' \
    | oxigraph convert --from-format nt --to-format rdf
<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF ...>
	<rdf:Description rdf:about="http://example.com/s">
		<oxprefix: xmlns:oxprefix="http://example.com/123" rdf:resource="http://example.com/o"/>
	</rdf:Description>
</rdf:RDF>
```

The command exits successfully, but an XML parser rejects the result with `not well-formed (invalid token)` at the `<oxprefix:` element.

The same behavior is exposed by the Rust `RdfXmlSerializer` API because `split_iri` returns an empty local name and `uri_to_qname_and_xmlns` formats it as `oxprefix:`.

**Expected behavior**

Serialization should return an `InvalidInput` error when an IRI cannot be encoded as an XML QName instead of successfully producing malformed RDF/XML.

This affects valid RDF graphs: the predicate IRI is valid RDF, but RDF/XML cannot represent every possible IRI as a property-element QName.

mattfaltyn:
## Summary

- return `InvalidInput` instead of emitting malformed RDF/XML when a predicate IRI has no valid QName local part
- fall back to an explicit `rdf:type` property when a type IRI cannot be used as an XML element name
- cover both paths with focused regression tests

Closes #1859.

## Validation

- `cargo test test_invalid_qname --locked`
- `cargo test --all-features --locked` (5 unit tests and 20 doctests)
- `cargo clippy --all-targets --all-features --locked -- -D warnings -D clippy::all`

### DIFF
--- a/lib/oxrdfxml/src/serializer.rs
+++ b/lib/oxrdfxml/src/serializer.rs
@@ -365,14 +365,15 @@ impl InnerRdfXmlWriter {
                 if let Term::NamedNode(t) = &triple.object {
                     if RESERVED_SYNTAX_TERMS.contains(&t.as_str()) {
                         (BytesStart::new("rdf:Description"), false)
-                    } else {
-                        let (prop_qname, prop_xmlns) = self.uri_to_qname_and_xmlns(t);
+                    } else if let Some((prop_qname, prop_xmlns)) = self.uri_to_qname_and_xmlns(t) {
                         let mut description_start = BytesStart::new(prop_qname.clone());
                         if let Some(prop_xmlns) = prop_xmlns {
                             description_start.push_attribute(prop_xmlns);
                         }
                         self.current_resource_tag = Some(prop_qname.into_owned());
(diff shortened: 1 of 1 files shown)


Tpt (approved):
Thank you for catching this and the nice fix!

PR #1864, opened 2026-08-12
Add --fail-on-named-graphs to load

### LINKED ISSUE #1863: CLI: add an option to reject named graphs during load
## Motivation

`oxigraph load --graph TARGET` remaps statements from the input default graph to `TARGET`, but named graph statements keep their original graph names. This is useful for dataset-capable formats such as JSON-LD, which is commonly used for a single graph despite not having a separate graph-only media type.

However, callers that require all loaded data to stay in one graph currently have no way to reject unexpected named graph statements. A dataset can therefore write outside `TARGET` without the caller noticing.

## Reproduction

1. Oxigraph: `0.6.0-dev` at `a40e00f74765a96f53b47cd34266b526f58fe5c4`
2. Platform: macOS/Darwin 25.6.0 arm64; Rust 1.97.1
3. Run:

```console
$ store=$(mktemp -d)/store
$ printf '%s\n' "<http://example.com/s> <http://example.com/p> <http://example.com/o> <http://example.com/original> ." | oxigraph load --location "$store" --format nq --graph http://example.com/target
$ oxigraph dump --location "$store" --format nq
<http://example.com/s> <http://example.com/p> <http://example.com/o> <http://example.com/original> .
```

## Proposed behavior

Add a `load --fail-on-named-graphs` option that:

- enables `RdfParser::without_named_graphs` for both stdin and file loading;
- causes the input load to fail when a named graph is encountered;
- remains compatible with `--graph`, so default-graph JSON-LD statements can still be loaded into a requested graph; and
- conflicts with `--lenient`, which would otherwise allow the named-graph parse error to be ignored.

The default behavior remains unchanged: dataset formats are accepted with `--graph`, default graph statements are remapped, and named graph statements are preserved.

## Existing coverage

I searched open and closed issues and pull requests using combinations of `--graph`, N-Quads, TriG, JSON-LD, dataset format, target graph, named graph, `without_named_graphs`, and `with_default_graph`. No issue or active pull request currently tracks a strict named-graph mode for the CLI loader. The closest item, #1657, concerns the Rust API rather than CLI loading.

mattfaltyn:
Fixes #1863.

## Summary

- Add `oxigraph load --fail-on-named-graphs` for callers that require graph-only input semantics.
- Apply `RdfParser::without_named_graphs` to both stdin and file loading paths.
- Keep dataset formats compatible with `--graph`, including loading default-graph JSON-LD into the requested target graph.
- Make `--fail-on-named-graphs` conflict with `--lenient` so the requested failure cannot be swallowed as a recoverable parse error.

## Rationale

`RdfParser::with_default_graph` intentionally remaps only default-graph statements; named graph statements retain their graph names. Rejecting all dataset-capable formats would prevent common single-graph JSON-LD use cases because JSON-LD has no separate graph-only media type.

The new opt-in flag expresses the actual strictness requirement directly: reject content containing named graphs while continuing to support dataset-capable formats whose content only uses the default graph.

## Validation

- `CARGO_BUILD_JOBS=8 cargo test --features rdf-12` (full native workspace suite)
- `cargo test -p oxigraph-cli cli_load_can_fail_on_named_graphs --features rdf-12`
- `cargo clippy -p oxigraph-cli --features rdf-12 --all-targets -- -D warnings`
- `cargo clippy -p oxigraph-cli --no-default-features --all-targets -- -D warnings`
- `cargo llvm-cov --package oxigraph-cli --features rdf-12 -- cli_load_can_fail_on_named_graphs`

The regression test covers rejection from file and stdin input, verifies the store stays empty after each atomic failure, and confirms default-graph JSON-LD still loads successfully into `--graph` when strict mode is enabled.

## Documentation

The `load --help` text now documents dataset-format `--graph` behavior and the new strict named-graph option.

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -125,9 +125,13 @@ pub enum Command {
         ///
         /// By default, the default graph is used.
         ///
-        /// Only available when loading a graph file (N-Triples, Turtle...) and not a dataset file (N-Quads, TriG...).
+        /// Default graph statements from dataset formats are loaded into this graph.
+        /// Named graph statements keep their original graph names.
         #[arg(long, value_hint = ValueHint::Url)]
         graph: Option<String>,
+        /// Fail if the input contains named graph statements
+        #[arg(long, conflicts_with = "lenient")]
+        fail_on_named_graphs: bool,
     },
     /// Dump the store content into a file
     Dump {
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -103,6 +103,7 @@ pub fn main() -> anyhow::Result<()> {
             format,
             base,
             graph,
+            fail_on_named_graphs,
         } => {
             let store = Store::open(&location)?;
             let format = if let Some(format) = format {
@@ -151,6 +152,7 @@ pub fn main() -> anyhow::Result<()> {
                     base.as_deref(),
                     graph,
                     lenient,
+                    fail_on_named_graphs,
                 )?;
                 loader.commit()?;
             } else {
@@ -209,6 +211,7 @@ pub fn main() -> anyhow::Result<()> {
                                             base.as_deref(),
                                             graph,
                                             lenient,
+                                            fail_on_named_graphs,
                                         )
                                     } else {
                                         bulk_load_file(
@@ -220,6 +223,7 @@ pub fn main() -> anyhow::Result<()> {
                                             base.as_deref(),
                                             graph,
                                             lenient,
+                                            fail_on_named_graphs,
                                         )
                                     }
                                 } {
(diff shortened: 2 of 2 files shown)


Tpt:
Thank you for the MR and the bug report.

The motivation for the change has been JSON-LD: it is a dataset format that is most often used for a single graph (there is no single-graph variant on JSON-LD with its own media-type/extension). So, rejecting all dataset formats means the `--graph` option can't be used with JSON-LD. It does not seems great to me.

If I understand the underlying "user requirement" properly, it's to fail when the input file(s) contains named graphs. What we can do is add a new option named something like `--fails-on-named-graphs` that enables `RdfParser::without_named_graphs` option and raise an error if a named graph is included in the input file. It has the downside of failing during parsing and not just from the format but covers JSON-LD nicely. What do you think about it?

mattfaltyn:
Thank you — agreed. I have updated the PR to add `--fail-on-named-graphs` using `RdfParser::without_named_graphs` instead of rejecting dataset-capable formats.

The revised behavior keeps `--graph` available for JSON-LD and other dataset formats, while callers that need strict graph-only input can opt into rejecting named graph statements. I also made the new option conflict with `--lenient`, since lenient loading would otherwise turn the requested named-graph failure into a recoverable parse error.

The regression test now covers file and stdin rejection with an unchanged store, plus successful default-graph JSON-LD loading into the requested graph.

Tpt (approved):
Thank you!

PR #1867, opened 2026-08-13
[oxrocksdb-sys] Surface more rocksdb features through the C API to Rust

ashleysommer:
This is an Optional PR.

I have been using oxrocksdb-sys in a different database project, it has some advantages over the vanilla librocksdb-sys for my use case.

I got to the point in my project that I needed some additional RocksDB features exposed through the C API. That includes

* attach callbacks on compaction events, and on flush events
* get table_properties iterator from SSTs for a given CF
* get user_collected_properties kv's from table_properties
* collect user_properties during compaction events using a callback.

The implementations for these wrappers were a combination of taking inspiration from upstream RocksDB API (for the parts that exist there) and `crocksdb` [library](https://github.com/tikv/rust-rocksdb/blob/master/librocksdb_sys/crocksdb/c.cc) from the TiKV project.

I'm not sure how useful this is for the oxigraph project. If not useful now, they could be useful in the future. The obvious use case (and what I'm using it for) is for collecting stats on each CF index at compaction time, storing these stats profiles in the user_collected_properties section of each SST, then using them to feed into SPARQL query optimisation rules.

I'd prefer these changes are adopted in this repo rather than creating and maintaining my own fork, but that's not my call to make.

### DIFF
--- a/oxrocksdb-sys/api/c.cc
+++ b/oxrocksdb-sys/api/c.cc
@@ -1,19 +1,34 @@
 #include "c.h"
 
 #include <rocksdb/db.h>
+#include <rocksdb/listener.h>
+#include <rocksdb/table_properties.h>
 #include <rocksdb/utilities/write_batch_with_index.h>
 
 #include <cstring>
+#include <memory>
 #include <vector>
 
 using ROCKSDB_NAMESPACE::ColumnFamilyHandle;
+using ROCKSDB_NAMESPACE::CompactionJobInfo;
+using ROCKSDB_NAMESPACE::CompactionServiceOptionsOverride;
 using ROCKSDB_NAMESPACE::DB;
+using ROCKSDB_NAMESPACE::EntryType;
+using ROCKSDB_NAMESPACE::EventListener;
+using ROCKSDB_NAMESPACE::FlushJobInfo;
 using ROCKSDB_NAMESPACE::IngestExternalFileOptions;
 using ROCKSDB_NAMESPACE::Iterator;
+using ROCKSDB_NAMESPACE::Options;
 using ROCKSDB_NAMESPACE::PinnableSlice;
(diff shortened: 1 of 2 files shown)


ashleysommer:
Here is a sample Rust file showing how each of these features is used, and proving their correctness in implementation
https://gist.github.com/ashleysommer/2e149e8103f704fb7d91970984b870b4

Tpt:
Sorry for the slow answer.

That's a large API footprint and I am a bit reluctant to add it in this repo without any test coverage.

> The obvious use case (and what I'm using it for) is for collecting stats on each CF index at compaction time, storing these stats profiles in the user_collected_properties section of each SST, then using them to feed into SPARQL query optimisation rules.

Thank you! That's a great idea.

> I'd prefer these changes are adopted in this repo rather than creating and maintaining my own fork, but that's not my call to make.

I would tend to suggest you to do your own fork, it is quite cheap to maintain (just need to update the submodule when there is a new rocksdb release) and you are not tied to Oxigraph release timing. The only downside I see is if you want to have both Oxigraph and your database in the same binary. Is it something that might happen?

Tpt:
Sorry for the slow answer.

That's a large API footprint and I am a bit reluctant to add it in this repo without any test coverage.

> The obvious use case (and what I'm using it for) is for collecting stats on each CF index at compaction time, storing these stats profiles in the user_collected_properties section of each SST, then using them to feed into SPARQL query optimisation rules.

Thank you! That's a great idea.

> I'd prefer these changes are adopted in this repo rather than creating and maintaining my own fork, but that's not my call to make.

I would tend to suggest you to do your own fork, it is quite cheap to maintain (just need to update the submodule when there is a new rocksdb release) and you are not tied to Oxigraph release timing. The only downside I see is if you want to have both Oxigraph and your database in the same binary. Is it something that might happen?

PR #1869, opened 2026-08-13
Fix HTTP content negotiation specificity

### LINKED ISSUE #1868: HTTP content negotiation can return explicitly unacceptable media types
**Describe the bug**

The CLI HTTP server can return a representation that is outside the media range accepted by the client. It can also return a representation that a more-specific media range explicitly excludes with `q=0`.

This affects both SPARQL query-result negotiation and RDF graph/dataset negotiation. For example, `Accept: text/*` may produce `application/sparql-results+json` from `/query` and `application/n-quads` from `/store`.

The same negotiation logic ranks a broad wildcard before applying a more-specific exclusion. Consequently, this header can still produce JSON even though JSON is explicitly unacceptable:

```http
Accept: */*;q=1, application/sparql-results+json;q=0, text/csv;q=0.5
```

[RFC 9110 section 12.5.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-accept) requires the most specific matching media range to determine a representation's quality. A `q=0` value means that representation is not acceptable.

**To Reproduce**

1. Build Oxigraph CLI from `main` at `2461755a9f14d7a1402379ec40957c21f35d88ce` (also reproducible with v0.5.9). Reproduced on macOS arm64.
2. Start an in-memory server:

   ```console
   cargo run --release --manifest-path cli/Cargo.toml -- serve
   ```

3. Request a SELECT result using a text wildcard:

   ```console
   curl -i -G http://localhost:7878/query \
     -H 'Accept: text/*' \
     --data-urlencode 'query=SELECT * WHERE { ?s ?p ?o }'
   ```

   Actual response: `200 OK` with `Content-Type: application/sparql-results+json`.

4. Request the default graph using the same wildcard:

   ```console
   curl -i 'http://localhost:7878/store?default' -H 'Accept: text/*'
   ```

   Actual response: `200 OK` with `Content-Type: application/n-quads`.

5. Explicitly exclude JSON while accepting a lower-quality text representation:

   ```console
   curl -i -G http://localhost:7878/query \
     -H 'Accept: */*;q=1, application/sparql-results+json;q=0, text/csv;q=0.5' \
     --data-urlencode 'query=SELECT * WHERE { ?s ?p ?o }'
   ```

   Actual response: `200 OK` with `Content-Type: application/sparql-results+json`.

**Expected behavior**

- `text/*` only selects a representation whose media type is in the `text` type, such as CSV for query results and Turtle for RDF.
- A more-specific range overrides a broader range for the same representation, so the explicitly excluded JSON representation is not selected. In the third example, another representation accepted by `*/*`, such as SPARQL Results XML, should be returned.
- Existing defaults for a missing `Accept` header and `*/*` should remain unchanged.

**Root cause**

`rdf_content_negotiation` and `query_results_content_negotiation` map `text/*` to their application-typed defaults. The shared `content_negotiation` function also keeps only the highest quality encountered globally, so it skips a lower-quality, more-specific range before that range can override the wildcard match for the same representation.

The fix should remain in the shared negotiation path and include regressions for both media-type families and the `q=0` specificity case.

mattfaltyn:
Closes #1868.

## Summary

- evaluate `Accept` quality and media-range specificity per supported representation
- make `text/*` select text-typed query-result and RDF formats
- honor exact and parameter-specific `q=0` exclusions over broader ranges
- preserve exact parameterized formats such as the JSON-LD streaming profile

Missing `Accept` and `*/*` keep their existing JSON/N-Quads defaults. Ties retain the server preference order declared by each endpoint.

## Validation

- `git diff --check`
- `cargo test -p oxigraph-cli accept --features rdf-12 -- --nocapture` (12 passed)
- `cargo clippy --manifest-path cli/Cargo.toml --all-targets --features rdf-12 -- -D warnings -D clippy::all`
- `cargo clippy --manifest-path cli/Cargo.toml --all-targets --no-default-features -- -D warnings -D clippy::all`
- `cargo test --features rdf-12 -- --skip cli_ --test-threads=1` (all executed workspace tests and doctests passed)

The 22 unchanged `cli_*` subprocess tests were excluded from the final workspace run because locally built macOS executables repeatedly stalled in `_dyld_start` on this external-volume workspace. All in-process CLI HTTP tests, including the new regressions, were executed. Upstream CI exercises the subprocess tests on Linux, macOS, and Windows.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1673,9 +1673,17 @@ fn rdf_content_negotiation(request: &Request<Body>) -> Result<RdfFormat, HttpErr
         RdfFormat::from_media_type,
         RdfFormat::NQuads,
         &[
-            ("application", RdfFormat::NQuads),
-            ("text", RdfFormat::NQuads),
+            RdfFormat::NQuads,
+            RdfFormat::Turtle,
+            RdfFormat::NTriples,
+            RdfFormat::RdfXml,
+            RdfFormat::TriG,
+            RdfFormat::N3,
+            RdfFormat::JsonLd {
+                profile: JsonLdProfileSet::empty(),
+            },
         ],
+        RdfFormat::media_type,
         "application/n-quads or text/turtle",
     )
 }
@@ -1688,18 +1696,22 @@ fn query_results_content_negotiation(
         QueryResultsFormat::from_media_type,
         QueryResultsFormat::Json,
         &[
-            ("application", QueryResultsFormat::Json),
-            ("text", QueryResultsFormat::Json),
+            QueryResultsFormat::Json,
+            QueryResultsFormat::Xml,
+            QueryResultsFormat::Csv,
+            QueryResultsFormat::Tsv,
         ],
+        QueryResultsFormat::media_type,
         "application/sparql-results+json or text/tsv",
     )
 }
 
-fn content_negotiation<F: Copy>(
+fn content_negotiation<F: Copy + Eq>(
     request: &Request<Body>,
     parse: impl Fn(&str) -> Option<F>,
     default: F,
-    default_by_base: &[(&str, F)],
+    supported: &[F],
+    media_type: impl Fn(F) -> &'static str,
     example: &str,
 ) -> Result<F, HttpError> {
     let header = request
@@ -1713,8 +1725,7 @@ fn content_negotiation<F: Copy>(
     if header.is_empty() {
         return Ok(default);
     }
-    let mut result = None;
-    let mut result_score = 0_f32;
+    let mut candidates: Vec<_> = supported.iter().map(|format| (*format, None)).collect();
     for mut possible in header.split(',') {
         let mut score = 1.;
         if let Some((possible_type, last_parameter)) = possible.rsplit_once(';') {
@@ -1727,9 +1738,6 @@ fn content_negotiation<F: Copy>(
                 }
             }
         }
-        if score <= result_score {
-            continue;
-        }
         let (possible_base, possible_sub) = possible
             .split_once(';')
             .unwrap_or((possible, ""))
@@ -1739,21 +1747,45 @@ fn content_negotiation<F: Copy>(
         let possible_base = possible_base.trim();
         let possible_sub = possible_sub.trim();
 
-        let mut format = None;
-        if possible_base == "*" && possible_sub == "*" {
-            format = Some(default);
+        let parameter_count = possible.split(';').skip(1).count();
+        let (format, specificity, wildcard) = if possible_base == "*" && possible_sub == "*" {
+            (None, (0, parameter_count), true)
         } else if possible_sub == "*" {
-            for (base, sub_format) in default_by_base {
-                if *base == possible_base {
-                    format = Some(*sub_format);
+            (None, (1, parameter_count), true)
+        } else {
+            (parse(possible), (2, parameter_count), false)
+        };
+        if let Some(format) = format {
+            if !candidates.iter().any(|(candidate, _)| *candidate == format) {
+                candidates.push((format, None));
+            }
+        }
+        for (candidate, candidate_score) in &mut candidates {
+            let candidate_base = media_type(*candidate).split_once('/').unwrap().0;
+            if format == Some(*candidate)
+                || (wildcard && (possible_base == "*" || possible_base == candidate_base))
+            {
+                if candidate_score
+                    .is_none_or(|(_, previous_specificity)| specificity > previous_specificity)
+                {
+                    *candidate_score = Some((score, specificity));
+                } else if let Some((previous_score, previous_specificity)) = candidate_score {
+                    if specificity == *previous_specificity && score > *previous_score {
+                        *previous_score = score;
+                    }
                 }
             }
-        } else {
-            format = parse(possible);
         }
-        if let Some(format) = format {
-            result = Some(format);
-            result_score = score;
+    }
+
+    let mut result = None;
+    let mut result_score = 0_f32;
+    for (format, candidate) in candidates {
+        if let Some((score, _)) = candidate {
+            if score > result_score {
+                result = Some(format);
+                result_score = score;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -1727,8 +1739,7 @@ fn content_negotiation<F: Copy>(
     if header.is_empty() {
         return Ok(default);
     }
-    let mut result = None;
-    let mut result_score = 0_f32;
+    let mut candidates: Vec<_> = supported.iter().map(|format| (*format, None)).collect();
```
Tpt: This would simplify the code and I think provide the same behavior (`f32::EPSILON` is the smallest positive float, this way other formats are scored higher than the explicitly excluded ones)
```suggestion
    let mut candidates: Vec<_> = supported.iter().map(|format| (*format, (f32::EPSILON, (0,0))).collect();
```

cli/src/main.rs
```diff
@@ -1687,9 +1687,17 @@ fn rdf_content_negotiation(request: &Request<Body>) -> Result<RdfFormat, HttpErr
...
-            ("text", RdfFormat::NQuads),
+            RdfFormat::NQuads,
+            RdfFormat::Turtle,
+            RdfFormat::NTriples,
+            RdfFormat::RdfXml,
+            RdfFormat::TriG,
+            RdfFormat::N3,
+            RdfFormat::JsonLd {
+                profile: JsonLdProfileSet::empty(),
+            },
```
Tpt, line 1680: nit: my personal preference
```suggestion
            RdfFormat::NQuads,
            RdfFormat::TriG,
            RdfFormat::JsonLd {
                profile: JsonLdProfileSet::empty(),
            },
            RdfFormat::Turtle,
            RdfFormat::NTriples,
            RdfFormat::RdfXml,
            RdfFormat::N3,
```

Tpt (approved):
Thank you for this! Just a minor improvement idea

cli/src/main.rs
mattfaltyn: Thanks! Simplified this with a zero-score sentinel so unmatched formats still return 406.

cli/src/main.rs
mattfaltyn: Thanks! Applied this ordering.

mattfaltyn:
Thanks! The CodSpeed report used stale base 848ae80 instead of PR base 2461755, plus a different environment. Since this PR only changes CLI negotiation, the reported store benchmark changes appear unrelated.

cli/src/main.rs
```diff
@@ -1687,9 +1687,17 @@ fn rdf_content_negotiation(request: &Request<Body>) -> Result<RdfFormat, HttpErr
         RdfFormat::from_media_type,
         RdfFormat::NQuads,
         &[
-            ("application", RdfFormat::NQuads),
-            ("text", RdfFormat::NQuads),
+            RdfFormat::NQuads,
+            RdfFormat::TriG,
+            RdfFormat::JsonLd {
+                profile: JsonLdProfileSet::empty(),
```
Tpt, line 1679: CI fails because `JsonLdProfileSet` is not imported

cli/src/main.rs
```diff
@@ -1753,19 +1776,37 @@ fn content_negotiation<F: Copy>(
...
+        for (candidate, candidate_score) in &mut candidates {
+            let candidate_base = media_type(*candidate).split_once('/').unwrap().0;
+            if format == Some(*candidate)
+                || (wildcard && (possible_base == "*" || possible_base == candidate_base))
+            {
+                if specificity > candidate_score.1 {
+                    *candidate_score = (score, specificity);
+                } else if specificity == candidate_score.1 && score > candidate_score.0 {
+                    candidate_score.0 = score;
                 }
```
Tpt: other idea (please ignore) to simplify the code:
```suggestion
                if (specificity, score) > (candidate_score.1, candidate_score.0) {
                    *candidate_score = (score, specificity);
                }
```
to avoid the `(candidate_score.1, candidate_score.0)` swap, you can switch the `specificity` and `score` order in `candidates` to use `(specificity, score)` instead of `(score, specificity)`, this way the code here is the simple:

```suggestion
                if (specificity, score) > candidate_score {
                    *candidate_score = (specificity, score);
                }
```

cli/src/main.rs
```diff
@@ -1753,19 +1776,37 @@ fn content_negotiation<F: Copy>(
...
+        };
+        if let Some(format) = format {
+            if !candidates.iter().any(|(candidate, _)| *candidate == format) {
+                candidates.push((format, (0., (0, 0))));
+            }
+        }
+        for (candidate, candidate_score) in &mut candidates {
+            let candidate_base = media_type(*candidate).split_once('/').unwrap().0;
+            if format == Some(*candidate)
+                || (wildcard && (possible_base == "*" || possible_base == candidate_base))
```
Tpt: idea to avoid the need for the `wildcard` variable:
```suggestion
                || (possible_sub == "*" && (possible_base == "*" || possible_base == candidate_base))
```

Tpt (commented):
Thank you! Ignore the benchmarks they are a bit noisy.

CI is failing because of a missing import and I took the occasion to suggest other simplifications. Feel free to ignore them

cli/src/main.rs
mattfaltyn: Thanks! Rebased onto current `main` and restored the `JsonLdProfileSet` import.

cli/src/main.rs
mattfaltyn: Thanks! Applied this and now store and compare `(specificity, score)` directly.

cli/src/main.rs
mattfaltyn: Thanks! Applied this and removed the extra `wildcard` flag.

mattfaltyn:
Thanks for the helpful review! I rebased onto current `main`, fixed the import, applied both simplifications, and added regression coverage for nonmatching media parameters. All local checks and PR CI pass.

Tpt:
Thank you!

PR #1876, opened 2026-08-15
spareval: avoid expensive work if errors are raised

### LINKED ISSUE #1872: spareval: `OrderBy` ignores the cancellation token
`OrderBy` spareval execution operator keep pulling the inner iterator even when the cancellation token error is raised, effectively ignoring it. It should abort execution when the error is raised.

More globally, all execution operator should halt execution when an error is returned

Tpt:
Allows `CancellationToken` to work properly

Close #1872


Tpt:
There seems to be a regression somewhere #1877 should help investigating it

PR #1879, opened 2026-08-16
Propagate CONSTRUCT externalization errors

### LINKED ISSUE #1878: CONSTRUCT silently drops triples when term externalization fails
## Describe the bug

`CONSTRUCT` silently drops a template triple when decoding one of its bound variables fails. The graph iterator reports successful completion, so callers can receive and persist an incomplete graph instead of seeing the dataset error.

The same `QueryableDataset::externalize_term` failure is propagated by `SELECT` and `DESCRIBE`.

## To reproduce

Tested on:

- Oxigraph `main` at `180adf2272b292b19782eaea42f5ec02b5c15db3` (the same error-erasing code is present in v0.5.9)
- macOS 26.6.1 arm64
- Rust 1.97.1

Use a `QueryableDataset` that returns one quad and fails while externalizing its object:

```rust
use oxrdf::{NamedNode, Term};
use spareval::{InternalQuad, QueryEvaluator, QueryResults, QueryableDataset};
use spargebra::SparqlParser;
use std::io;

struct FailingDataset;

impl<'a> QueryableDataset<'a> for FailingDataset {
    type InternalTerm = u8;
    type Error = io::Error;

    fn internal_quads_for_pattern(
        &self,
        _subject: Option<&u8>,
        _predicate: Option<&u8>,
        _object: Option<&u8>,
        _graph_name: Option<Option<&u8>>,
    ) -> impl Iterator<Item = Result<InternalQuad<u8>, io::Error>> + use<'a> {
        std::iter::once(Ok(InternalQuad {
            subject: 1,
            predicate: 2,
            object: 3,
            graph_name: None,
        }))
    }

    fn internalize_term(&self, _term: Term) -> Result<u8, io::Error> {
        Ok(2)
    }

    fn externalize_term(&self, term: u8) -> Result<Term, io::Error> {
        match term {
            1 => Ok(NamedNode::new_unchecked("http://example.com/s").into()),
            2 => Ok(NamedNode::new_unchecked("http://example.com/p").into()),
            _ => Err(io::Error::other("dictionary read failed")),
        }
    }
}

fn main() {
    let query = SparqlParser::new()
        .parse_query("CONSTRUCT { ?s <http://example.com/p> ?o } WHERE { ?s ?p ?o }")
        .unwrap();
    let QueryResults::Graph(results) = QueryEvaluator::new()
        .prepare(&query)
        .execute(FailingDataset)
        .unwrap()
    else {
        unreachable!()
    };
    println!("{:?}", results.collect::<Result<Vec<_>, _>>());
}
```

Actual output:

```text
Ok([])
```

As controls, a `SELECT ?s ?o WHERE { ?s ?p ?o }` against the same dataset returns `Err(QueryEvaluationError::Dataset(...))`, while a constant-only `CONSTRUCT` succeeds.

## Expected behavior

The graph result iterator should yield the dataset error. It should skip a template triple only for normal SPARQL cases such as an unbound variable or a value that cannot occupy the requested RDF position.

## Cause

`get_triple_template_value` calls:

```rust
dataset.externalize_term(t.clone()).ok() // TODO: raise error
```

This converts a `QueryEvaluationError` to `None`. `ConstructIterator` cannot distinguish that failure from an unbound or invalid template value and silently skips the triple.

## Suggested fix

Return `Result<Option<Term>, QueryEvaluationError>` from `get_triple_template_value`, propagate externalization errors from `ConstructIterator`, and retain `Ok(None)` for legitimate template omissions.

mattfaltyn:
Fixes #1878.

## Summary

- propagate dataset externalization failures from CONSTRUCT graph iterators
- keep skipping triples for legitimate unbound or invalid template values
- clear per-solution blank-node state before yielding an error
- add a regression test using a fallible `QueryableDataset`

## Testing

- `cargo test -p spareval --lib --no-default-features`
- `cargo test -p spareval --lib --features sparql-12`
- `cargo test -p spareval --lib --features sep-0002`
- `cargo test -p spareval --lib --features sep-0006`
- `cargo clippy --all-targets -- -D warnings -D clippy::all` (`lib/spareval`)
- `cargo clippy --all-targets --features sparql-12 -- -D warnings -D clippy::all` (`lib/spareval`)
- `cargo clippy --all-targets --features sep-0002 -- -D warnings -D clippy::all` (`lib/spareval`)
- `cargo clippy --all-targets --features sep-0006 -- -D warnings -D clippy::all` (`lib/spareval`)
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p spareval`

### DIFF
--- a/lib/spareval/src/eval.rs
+++ b/lib/spareval/src/eval.rs
@@ -3458,21 +3458,19 @@ impl<'a, D: QueryableDataset<'a>> Iterator for ConstructIterator<'a, D> {
                     Err(error) => return Some(Err(error)),
                 };
                 for template in &self.template {
-                    if let (Some(subject), Some(predicate), Some(object)) = (
+                    let (subject, predicate, object) = match (
                         get_triple_template_value(
                             &template.subject,
                             &tuple,
                             &mut self.bnodes,
                             &self.eval.dataset,
-                        )
-                        .and_then(|t| t.try_into().ok()),
+                        ),
                         get_triple_template_value(
                             &template.predicate,
                             &tuple,
                             &mut self.bnodes,
                             &self.eval.dataset,
-                        )
-                        .and_then(|t| t.try_into().ok()),
+                        ),
                         get_triple_template_value(
                             &template.object,
                             &tuple,
@@ -3480,6 +3478,17 @@ impl<'a, D: QueryableDataset<'a>> Iterator for ConstructIterator<'a, D> {
                             &self.eval.dataset,
                         ),
                     ) {
+                        (Ok(Some(subject)), Ok(Some(predicate)), Ok(Some(object))) => {
+                            (subject, predicate, object)
+                        }
+                        (Err(error), _, _) | (_, Err(error), _) | (_, _, Err(error)) => {
+                            self.bnodes.clear();
+                            return Some(Err(error));
+                        }
+                        _ => continue,
+                    };
+                    if let (Ok(subject), Ok(predicate)) = (subject.try_into(), predicate.try_into())
+                    {
                         let triple = Triple {
                             subject,
                             predicate,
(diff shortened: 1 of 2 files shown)


Tpt:
Thanks!

PR #1882, opened 2026-08-19
Preserve SPARQL XML literal boundary whitespace

### LINKED ISSUE #1881: SPARQL XML parser trims literal boundary whitespace
**Describe the bug**

The SPARQL XML results parser silently removes leading and trailing XML whitespace characters from literal values. This also removes whitespace encoded as numeric character references, even though the serializer emits those references specifically to preserve value boundaries.

This is a regression on `main`: PR #1049 preserved boundary whitespace with the previous parser, but the Quick-XML migration in PR #1711 started decoding text and general references into one buffer and trimming that buffer before dispatching on `State::Literal`.

**To Reproduce**

1. Oxigraph revision: `main@6757cd8fba447eaefd43cc32d437b3fd37606827` (`0.6.0-dev` workspace)
2. Platform: macOS 26.6.1 arm64; Rust 1.97.1
3. Run:

```rust
use sparesults::{
    QueryResultsFormat, QueryResultsParser, SliceQueryResultsParserOutput,
};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let xml = r#"<sparql xmlns="http://www.w3.org/2005/sparql-results#"><head><variable name="x"/></head><results><result><binding name="x"><literal>&#32;padded&#32;</literal></binding></result></results></sparql>"#;
    let SliceQueryResultsParserOutput::Solutions(mut solutions) =
        QueryResultsParser::from_format(QueryResultsFormat::Xml).for_slice(xml)?
    else {
        unreachable!()
    };
    let solution = solutions.next().unwrap()?;
    assert_eq!(solution.get("x").unwrap().to_string(), r#"" padded ""#);
    Ok(())
}
```

The assertion fails because the parsed value is `"padded"` rather than `" padded "`.

**Expected behavior**

Literal boundary whitespace encoded using XML character references should be preserved after decoding. Oxigraph's serializer emits these references specifically to distinguish significant boundary whitespace while retaining the parser's existing behavior of trimming unescaped boundary whitespace.

**Root cause**

`XmlInnerSolutionsParser::read_event` resolves `Event::Text` and `Event::GeneralRef` into `text_buffer`. On every `Event::End`, it then applies `trim_matches` before examining the parser state, so significant literal whitespace is lost.

The same parser is used by the public `sparesults` APIs, Python `parse_query_results`, and federated `SERVICE` responses.

PR #1882 tracks the first and last unescaped XML text events, trimming only their outer boundaries while preserving character-reference output, and covers both behaviors with focused regression tests.

mattfaltyn:
## Summary

- preserve SPARQL XML literal boundary whitespace produced by character references
- retain trimming for unescaped boundary whitespace by tracking the first and last text events
- cover the existing typed-literal behavior and leading/trailing space, tab, LF, and CR references

Fixes #1881.

## Testing

- `cargo test --all-targets --features async-tokio,sparql-12` (`lib/sparesults`)
- `cargo clippy --all-targets -- -D warnings -D clippy::all` (`lib/sparesults`)
- `cargo clippy --all-targets --features async-tokio,sparql-12 -- -D warnings -D clippy::all` (`lib/sparesults`)
- `cargo fmt -- --check`

### DIFF
--- a/lib/sparesults/src/xml.rs
+++ b/lib/sparesults/src/xml.rs
@@ -672,7 +672,13 @@ impl XmlInnerSolutionsParser {
         event: Event<'_>,
     ) -> Result<Option<Vec<Option<Term>>>, QueryResultsParseError> {
         match event {
-            Event::Start(event) => match self.state_stack.last().ok_or_else(|| {
+            Event::Start(event) => match {
+                self.text_buffer.clear();
+                self.first_text_event_end = None;
+                self.last_text_event_start = None;
+                self.state_stack.last()
+            }
+            .ok_or_else(|| {
                 QueryResultsSyntaxError::msg("Extra XML is not allowed at the end of the document")
             })? {
                 State::Start => {
@@ -727,9 +733,6 @@ impl XmlInnerSolutionsParser {
                         self.state_stack.push(State::BNode);
                         Ok(None)
                     } else if event.local_name().as_ref() == b"literal" {
-                        self.text_buffer.clear();
-                        self.first_text_event_end = None;
-                        self.last_text_event_start = None;
                         for attr in event.attributes() {
                             let attr = attr.map_err(Error::from)?;
                             if attr.key.as_ref() == b"xml:lang" {
--- a/lib/sparesults/src/xml.rs
+++ b/lib/sparesults/src/xml.rs
@@ -567,6 +567,8 @@ impl XmlInnerQueryResultsParser {
                                 predicate_stack: Vec::new(),
                                 object_stack: Vec::new(),
                                 text_buffer: String::new(),
+                                first_text_event_end: None,
+                                last_text_event_start: None,
                                 xml_version: self.xml_version,
                             },
                         }))
@@ -659,6 +661,8 @@ struct XmlInnerSolutionsParser {
     predicate_stack: Vec<Term>,
     object_stack: Vec<Term>,
     text_buffer: String,
+    first_text_event_end: Option<usize>,
+    last_text_event_start: Option<usize>,
     xml_version: XmlVersion,
 }
 
@@ -723,9 +727,9 @@ impl XmlInnerSolutionsParser {
                         self.state_stack.push(State::BNode);
                         Ok(None)
                     } else if event.local_name().as_ref() == b"literal" {
-                        let value = take(&mut self.text_buffer);
-                        self.text_buffer
-                            .push_str(value.trim_start_matches(['\t', '\n', '\r', ' ']));
+                        self.text_buffer.clear();
+                        self.first_text_event_end = None;
+                        self.last_text_event_start = None;
                         for attr in event.attributes() {
                             let attr = attr.map_err(Error::from)?;
                             if attr.key.as_ref() == b"xml:lang" {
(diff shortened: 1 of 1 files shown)


lib/sparesults/src/xml.rs
```diff
@@ -723,6 +723,9 @@ impl XmlInnerSolutionsParser {
                         self.state_stack.push(State::BNode);
                         Ok(None)
                     } else if event.local_name().as_ref() == b"literal" {
+                        let value = take(&mut self.text_buffer);
+                        self.text_buffer
+                            .push_str(value.trim_start_matches(['\t', '\n', '\r', ' ']));
```
Tpt: If I am not mistaken, this is a no-op. `self.text_buffer` is always empty here

Tpt:
Thank you for this and caching the issue! I am not sure this is the best approach, it means that now
`<literal datatype="http://www.w3.org/2001/XMLSchema#boolean"> true </literal>` is parsed as `" true "^^<http://www.w3.org/2001/XMLSchema#>`. My guess is that we should go back to the previous approach and trim only the unescaped whitespaces. An approach is maybe to save in the parser state where is the end of the first text event in `self.text_buffer` and where is the start of the last text event and only trim before and after these positions.

lib/sparesults/src/xml.rs
mattfaltyn: Agreed. I removed this block in 6162a1fe. The parser now resets the literal event bounds on entry and trims using the first and last `Event::Text` offsets, leaving `Event::GeneralRef` output intact.

mattfaltyn:
@Tpt Implemented the suggested text-event boundary approach in 6162a1fe. Unescaped outer whitespace keeps the previous trimming behavior (including the typed-boolean example), while whitespace decoded from character references is preserved. The focused regression test covers both cases, and the full `cargo test --features rdf-12` suite passes.

lib/sparesults/src/xml.rs
```diff
@@ -813,20 +820,41 @@ impl XmlInnerSolutionsParser {
...
-                match self.state_stack.pop().ok_or_else(|| {
+                let first_text_event_end = self.first_text_event_end.take();
+                let last_text_event_start = self.last_text_event_start.take();
+                let state = self.state_stack.pop().ok_or_else(|| {
                     QueryResultsSyntaxError::msg(
                         "Extra XML is not allowed at the end of the document",
                     )
-                })? {
+                })?;
+                let value = OxString::new_owned(if matches!(state, State::Literal) {
```
Tpt: Do we want to do this only for literals? I would apply this logic for all states to keep behaviors more predictible

Tpt (commented):
Thank you! One last question

lib/sparesults/src/xml.rs
mattfaltyn: Agreed. Updated in 5283b0e0 to apply the tracked text-event boundary logic consistently for all parser states. I also added a regression test showing that a boundary character reference in a URI is preserved and rejected as invalid instead of being silently trimmed. `cargo test -p sparesults --all-targets --features async-tokio,sparql-12`, clippy with the same features, formatting, and diff checks pass.

Tpt (approved):
Thanks!

lib/sparesults/src/xml.rs
```diff
@@ -723,6 +727,9 @@ impl XmlInnerSolutionsParser {
                         self.state_stack.push(State::BNode);
                         Ok(None)
                     } else if event.local_name().as_ref() == b"literal" {
+                        self.text_buffer.clear();
+                        self.first_text_event_end = None;
+                        self.last_text_event_start = None;
```
Tpt: I think we need to move these lines to line 674 to fix the CI: we now clear the text buffer only for literals and not for all events

PR #1890, opened 2026-08-23
spareval: propagate keyless MINUS dataset errors

### LINKED ISSUE #1889: Keyless MINUS silently discards dataset read errors
**Describe the bug**

When the two operands of a SPARQL `MINUS` share no variables, an error while reading the right-hand side from a `QueryableDataset` is silently discarded. If every right-hand item is an error, the evaluator treats the right-hand side as empty and returns the left-hand solutions successfully.

This masks storage and cancellation failures and can return incomplete query results as if evaluation succeeded.

The keyless branch currently collects the right-hand side with:

```rust
let right: Vec<_> = right(from.clone()).filter_map(Result::ok).collect();
```

**To Reproduce**

1. Oxigraph `main` at `2799f720bc3f7fe414d3f594ebe8876b119b7256` (the same implementation is present in v0.5.9), tested on macOS arm64 with Rust 1.97.1.
2. Save the following as `lib/spareval/examples/minus_read_error.rs`:

```rust
use oxrdf::{NamedNode, Term};
use spareval::{InternalQuad, QueryEvaluator, QueryResults, QueryableDataset};
use spargebra::SparqlParser;
use std::io;

struct Dataset;

impl<'a> QueryableDataset<'a> for Dataset {
    type InternalTerm = u8;
    type Error = io::Error;

    fn internal_quads_for_pattern(
        &self,
        _subject: Option<&u8>,
        predicate: Option<&u8>,
        _object: Option<&u8>,
        _graph_name: Option<Option<&u8>>,
    ) -> impl Iterator<Item = Result<InternalQuad<u8>, io::Error>> + use<'a> {
        match predicate {
            Some(2) => Box::new(std::iter::once(Ok(InternalQuad {
                subject: 1,
                predicate: 2,
                object: 3,
                graph_name: None,
            }))) as Box<dyn Iterator<Item = _>>,
            Some(5) => Box::new(std::iter::once(Err(io::Error::other(
                "right-side read failed",
            )))),
            _ => Box::new(std::iter::empty()),
        }
    }

    fn internalize_term(&self, term: Term) -> Result<u8, io::Error> {
        Ok(match term {
            Term::NamedNode(node) if node.as_str().ends_with("/left") => 2,
            Term::NamedNode(node) if node.as_str().ends_with("/right") => 5,
            _ => 1,
        })
    }

    fn externalize_term(&self, term: u8) -> Result<Term, io::Error> {
        Ok(NamedNode::new_unchecked(format!("http://example.com/{term}")).into())
    }
}

fn main() {
    let query = SparqlParser::new()
        .parse_query(
            "SELECT ?s WHERE {
                ?s <http://example.com/left> ?o
                MINUS { ?x <http://example.com/right> ?y }
            }",
        )
        .unwrap();
    let QueryResults::Solutions(solutions) = QueryEvaluator::new()
        .prepare(&query)
        .execute(Dataset)
        .unwrap()
    else {
        panic!("expected solutions")
    };

    let result = solutions.collect::<Result<Vec<_>, _>>();
    assert_eq!(result.unwrap_err().to_string(), "right-side read failed");
}
```

3. Run `cargo run -p spareval --example minus_read_error`.

**Actual behavior**

The query returns the left-hand solution and `unwrap_err()` panics because the result is `Ok`.

**Expected behavior**

The solution iterator returns the `right-side read failed` dataset error.

The equivalent `MINUS` with a shared variable on both sides already propagates the error. PR #1876 added that behavior to the keyed branch, but the separate keyless branch still filters errors out.

**Proposed fix**

Collect the keyless right-hand iterator into `Result<Vec<_>, _>` and return the error before evaluating the left-hand side, with a regression test using a failing `QueryableDataset`.

mattfaltyn:
## Summary

- propagate right-hand dataset errors from keyless `MINUS` evaluation
- cover the disjoint-variable path with a failing `QueryableDataset` regression test

The keyless branch previously used `filter_map(Result::ok)`, so a right-hand storage or cancellation error could be treated as an empty result and the left-hand solutions returned successfully. Collecting into `Result<Vec<_>, _>` now yields the error, matching the keyed `MINUS` branch.

Closes #1889.

## Validation

- `cargo test -p spareval minus_propagates_right_side_read_errors`
- `cargo test -p spareval`
- `cargo fmt -- --check`
- `taplo check`
- `cargo clippy --all-targets -- -D warnings -D clippy::all` in `lib/spareval`
- `cargo clippy --all-targets --features sparql-12 -- -D warnings -D clippy::all` in `lib/spareval`
- `cargo clippy --all-targets --features sep-0002 -- -D warnings -D clippy::all` in `lib/spareval`
- `cargo clippy --all-targets --features sep-0006 -- -D warnings -D clippy::all` in `lib/spareval`

### DIFF
--- a/lib/spareval/src/eval.rs
+++ b/lib/spareval/src/eval.rs
@@ -1280,7 +1280,10 @@ impl<'a, D: QueryableDataset<'a>> SimpleEvaluator<'a, D> {
             MinusAlgorithm::HashBuildRightProbeLeft { keys } => {
                 if keys.is_empty() {
                     Ok(Rc::new(move |from| {
-                        let right: Vec<_> = right(from.clone()).filter_map(Result::ok).collect();
+                        let right = match right(from.clone()).collect::<Result<Vec<_>, _>>() {
+                            Ok(right) => right,
+                            Err(error) => return Box::new(once(Err(error))),
+                        };
                         if right.is_empty() {
                             return left(from);
                         }
--- a/lib/spareval/src/lib.rs
+++ b/lib/spareval/src/lib.rs
@@ -1027,29 +1027,38 @@ mod tests {
     use spargebra::vocab::sparql;
     use sparopt::algebra::{Expression, QueryExpression};
 
-    struct FailingExternalizationDataset;
+    struct FailingDataset {
+        fail_reads: bool,
+    }
 
-    impl<'a> QueryableDataset<'a> for FailingExternalizationDataset {
+    impl<'a> QueryableDataset<'a> for FailingDataset {
         type InternalTerm = u8;
         type Error = io::Error;
 
         fn internal_quads_for_pattern(
             &self,
             _subject: Option<&u8>,
-            _predicate: Option<&u8>,
+            predicate: Option<&u8>,
             _object: Option<&u8>,
             _graph_name: Option<Option<&u8>>,
         ) -> impl Iterator<Item = Result<InternalQuad<u8>, io::Error>> + use<'a> {
-            std::iter::once(Ok(InternalQuad {
-                subject: 1,
-                predicate: 2,
-                object: 3,
-                graph_name: None,
-            }))
+            std::iter::once(if self.fail_reads && predicate == Some(&5) {
+                Err(io::Error::other("dataset read failed"))
+            } else {
+                Ok(InternalQuad {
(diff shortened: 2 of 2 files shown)


Tpt (approved):
Great catch! Thanks!

PR #1896, opened 2026-08-24
Make Graph Store PUT replacements atomic

### LINKED ISSUE #1895: Graph Store PUT deletes existing data when parsing fails
## Describe the bug

The Graph Store HTTP Protocol `PUT` handler clears the target graph before parsing and loading the request body in a separate transaction. If parsing fails, the request returns `400 Bad Request`, but the previously stored graph has already been deleted.

This affects replacement of the default graph, named graphs, and the whole dataset. In the last case, a malformed `PUT /store` can clear every graph in the store.

The default loading path is documented as transactional. The explicit `no_transaction` option is not involved.

## To reproduce

Reproduced on macOS arm64 using `main` at `8dcfb6b66cbb077bb2406379abb280d2471970d7`. Source inspection confirms that v0.5.9 has the same handler sequence.

Start a persistent server:

```console
$ oxigraph serve --location /tmp/oxigraph-put-repro --bind 127.0.0.1:7878
```

Store a valid default graph:

```console
$ curl -i -X PUT \
    -H 'Content-Type: application/n-triples' \
    --data-binary '<http://example.com/s> <http://example.com/p> <http://example.com/o> .' \
    'http://127.0.0.1:7878/store?default'
HTTP/1.1 204 No Content
```

Verify that the triple exists:

```console
$ curl -H 'Accept: application/n-triples' \
    'http://127.0.0.1:7878/store?default'
<http://example.com/s> <http://example.com/p> <http://example.com/o> .
```

Attempt to replace it with malformed Turtle:

```console
$ curl -i -X PUT \
    -H 'Content-Type: text/turtle' \
    --data-binary '<http://example.com/s> <http://example.com/p> .' \
    'http://127.0.0.1:7878/store?default'
HTTP/1.1 400 Bad Request

Syntax error at line 1 column 47: . is not a valid RDF object
```

Read the graph again:

```console
$ curl -H 'Accept: application/n-triples' \
    'http://127.0.0.1:7878/store?default'
```

The response is empty: the valid graph was cleared despite the failed replacement. The same sequence with N-Quads and `PUT /store` clears the entire dataset.

As a control, a malformed `POST /store?default` returns `400` and preserves existing data.

## Expected behavior

A failed transactional `PUT` should leave the target graph or dataset unchanged. Clearing the old data and loading the replacement should commit atomically only after the request body has parsed successfully.

## Root cause

The [`PUT` handler](https://github.com/oxigraph/oxigraph/blob/8dcfb6b66cbb077bb2406379abb280d2471970d7/cli/src/main.rs#L1097-L1143) commits `clear_graph` or `clear` before calling `web_load_graph` or `web_load_dataset`.

Those helpers call [`Store::load_from_reader`](https://github.com/oxigraph/oxigraph/blob/8dcfb6b66cbb077bb2406379abb280d2471970d7/lib/oxigraph/src/store.rs#L416-L426), which opens an independent transaction. Its rollback cannot restore data removed by the earlier transaction.

## Suggested fix

For the normal loading path, perform target creation/clearing and `load_from_reader` in one `Transaction`, then commit only after parsing succeeds. Keep the explicit `no_transaction` bulk-loading path unchanged.

A regression test should seed the target, send a malformed `PUT`, assert `400`, and verify that the original data remains present. Covering both a selected graph and the whole dataset would protect the two destructive branches.

mattfaltyn:
## Summary

- run Graph Store `PUT` clear and load operations in the same transaction
- preserve the existing `no_transaction` bulk-loading behavior
- add a regression test for failed default-graph and whole-dataset replacements

If parsing a transactional `PUT` fails, dropping the transaction now rolls back the preceding clear and leaves the existing data unchanged.

Closes #1895.

## Validation

- `cargo fmt -- --check`
- `taplo check`
- `cargo test --package oxigraph-cli graph_store_put_parse_error_preserves_existing_data`
- `cargo test --package oxigraph-cli graph_store`
- `cd cli && cargo clippy --all-targets -- -D warnings -D clippy::all`
- `cd cli && cargo clippy --all-targets --no-default-features -- -D warnings -D clippy::all`
- `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1126,19 +1126,13 @@ fn handle_request(
                         false
                     }
                 };
-                if RequestParams::from_request_url(request).contains("no_transaction") {
-                    transaction.commit().map_err(internal_server_error)?;
-                    web_load_graph(store, request, format, &GraphName::from(target), None)?;
-                } else {
-                    web_load_graph(
-                        store,
-                        request,
-                        format,
-                        &GraphName::from(target),
-                        Some(&mut transaction),
-                    )?;
-                    transaction.commit().map_err(internal_server_error)?;
-                }
+                web_load_graph(
+                    store,
+                    request,
+                    format,
+                    &GraphName::from(target),
+                    Some(transaction),
+                )?;
                 Response::builder()
                     .status(if new {
                         StatusCode::CREATED
@@ -1152,13 +1146,7 @@ fn handle_request(
                     .ok_or_else(|| unsupported_media_type(&content_type))?;
                 let mut transaction = store.start_transaction().map_err(internal_server_error)?;
                 transaction.clear().map_err(internal_server_error)?;
-                if RequestParams::from_request_url(request).contains("no_transaction") {
-                    transaction.commit().map_err(internal_server_error)?;
-                    web_load_dataset(store, request, format, None)?;
-                } else {
-                    web_load_dataset(store, request, format, Some(&mut transaction))?;
-                    transaction.commit().map_err(internal_server_error)?;
-                }
+                web_load_dataset(store, request, format, Some(transaction))?;
                 Response::builder()
                     .status(StatusCode::NO_CONTENT)
                     .body(Body::empty())
@@ -1902,7 +1890,7 @@ fn web_load_graph(
     request: &mut Request<Body>,
     format: RdfFormat,
     to_graph_name: &GraphName,
-    transaction: Option<&mut Transaction<'_>>,
+    transaction: Option<Transaction<'_>>,
 ) -> Result<(), HttpError> {
     let args = RequestParams::from_request_url(request);
     let base_iri = if let GraphName::NamedNode(graph_name) = to_graph_name {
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
Tpt, line 1910: idea: what about commiting and dropping the given `transaction` here. This would allow to drop the related-if/else code path in callers.

```suggestion
    if args.contains("no_transaction") {
        if let Some(transaction) = transaction.take() {
            transaction.commit().map_err(internal_server_error)?;
        }
```

you would need to take `transaction` by value in the function and `commit` at the end.

Tpt (commented):
Great catch! Thank you. A refactoring idea but not sure if it's a good one. Feel free to reject it if you feel it's not a good idea

cli/src/main.rs
mattfaltyn: Great idea, thank you! Implemented in 692e4afb: the load helpers now own and commit the transaction, which removes the duplicated `no_transaction` branching from the PUT callers.

Tpt:
Thank you!

PR #1899, opened 2026-08-27
Fix Graph Store POST creation status

### LINKED ISSUE #1898: Graph Store POST returns inverted creation statuses
**Describe the bug**

The CLI Graph Store handler returns the creation statuses in reverse for a targeted `POST`:

- Posting to a missing named graph creates it but returns `204 No Content`.
- Posting to an existing named graph merges the payload but returns `201 Created`.

The handler currently computes `new` with `assert_that_graph_exists(...).is_ok()`, so `new` is `true` precisely when the graph already exists. This can make clients treat a merge as resource creation and miss actual graph creation.

The documented `/store?graph=...` workflow allows a targeted `POST` to add data to a named graph. The status behavior before the graph-existence refactor also used `201` for a missing target and `204` for an existing target.

**To Reproduce**

1. Build current `main` at `348f8e02082b0ad3b72c8d605bab38eaf887cd3d` on macOS arm64 with the repository submodules initialized.
2. Start a clean server:

   ```console
   cargo run -p oxigraph-cli -- serve --location /tmp/oxigraph-post-repro
   ```

3. Post the same valid graph payload twice:

   ```console
   curl -i -X POST \
     -H 'Content-Type: application/n-triples' \
     --data '<http://example.com/s> <http://example.com/p> <http://example.com/o> .' \
     'http://localhost:7878/store?graph=http://example.com/g'

   curl -i -X POST \
     -H 'Content-Type: application/n-triples' \
     --data '<http://example.com/s2> <http://example.com/p> <http://example.com/o> .' \
     'http://localhost:7878/store?graph=http://example.com/g'
   ```

The first request returns `204 No Content`; the second returns `201 Created`.

Expected:

- The first request returns `201 Created` because the named graph did not exist.
- The second request returns `204 No Content` because it merges into an existing graph.
- Both payloads remain present in the target graph.

The same inversion applies to indirect graph IRIs under `/store/...`.

**Suggested validation**

Cover both sides of the invariant in the existing Graph Store tests: a targeted `POST` to a missing graph returns `201`, and a subsequent targeted `POST` to that existing graph returns `204`.

mattfaltyn:
Closes #1898.

## Summary

- Return `201 Created` when a targeted Graph Store `POST` creates a missing named graph.
- Return `204 No Content` when a targeted `POST` merges into an existing graph.
- Exercise both states in the existing Graph Store tests, including lenient and `no_transaction` paths.

## Why

The targeted `POST` handler used `assert_that_graph_exists(...).is_ok()` as its `new` flag. That made the flag true for an existing graph and false for a missing graph, reversing the two success statuses. The replacement treats only the helper's `404 Not Found` result as a new graph and still propagates storage errors.

## Validation

- [x] `cargo test -p oxigraph-cli graph_store -- --nocapture`
- [x] `cargo test -p oxigraph-cli lenient_load -- --nocapture`
- [x] `cargo fmt -- --check`
- [x] `cargo clippy --all-targets -- -D warnings -D clippy::all` (from `cli/`)
- [x] `cargo clippy --all-targets --no-default-features -- -D warnings -D clippy::all` (from `cli/`)
- [x] `cargo test --workspace --exclude sparql-smith`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1195,7 +1195,11 @@ fn handle_request(
             if let Some(target) = store_target(request)? {
                 let format = RdfFormat::from_media_type(&content_type)
                     .ok_or_else(|| unsupported_media_type(&content_type))?;
-                let new = assert_that_graph_exists(store, &target).is_ok();
+                let new = match assert_that_graph_exists(store, &target) {
+                    Ok(()) => false,
+                    Err((StatusCode::NOT_FOUND, _)) => true,
+                    Err(error) => return Err(error),
+                };
                 web_load_graph(store, request, format, &GraphName::from(target), None)?;
                 Response::builder()
                     .status(if new {
@@ -3345,7 +3349,7 @@ mod tests {
             .uri("http://localhost/store?graph=http://example.com")
             .header(CONTENT_TYPE, "text/turtle")
             .body("<> <http://example.com/p> <http://example.com/o1> .")?;
-        server.test_status(request, StatusCode::NO_CONTENT)?;
+        server.test_status(request, StatusCode::CREATED)?;
 
         // GET
         let request = Request::builder()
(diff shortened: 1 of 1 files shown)


Tpt (approved):
thank you for catching this!

Issue #1901, opened 2026-08-28
oxjsonld: JSON-LD toRDF conformance gaps in processing modes, scoped contexts, and RDF direction options


eljeffeg:
## Summary

`oxjsonld` fails multiple cases in the current W3C JSON-LD toRDF and Streaming JSON-LD toRDF suites. The failures reproduce from Oxigraph commit `ec68e3dd` using these corpora:

- JSON-LD API: `289ebf37c30846ba2dab8a33559cc4d348161e0d`
- JSON-LD streaming: `948d24ce66a129734a97efba01cb5d6d8552f4f5`

After excluding four relative-IRI cases already fixed on Oxigraph main, a diagnostic run with the remaining ignored cases enabled reports:

| Suite | Passing | Failing |
| --- | ---: | ---: |
| JSON-LD toRDF | 446/467 | 21 |
| Streaming JSON-LD toRDF | 456/483 | 27 |

The failures fall into related implementation and harness clusters rather than independent result mismatches.

## Reproduction

1. Check out Oxigraph commit `ec68e3dd`.
2. In `testsuite/tests/parser.rs`, temporarily remove the affected JSON-LD cases from the two `check_testsuite` ignore lists.
3. Run:

```console
cargo test -p oxigraph-testsuite jsonld_to_rdf_ -- --nocapture
```

### Processing-mode and JSON-LD 1.0 behavior

Cases: `te026`, `te071`, `te115`, `te116`, `ter02`, `ter03`, `ter24`, and `ter32`.

The harness does not use `specVersion` to select JSON-LD 1.0 processing. Once the correct mode is selected, the parser also needs to:

- apply the term-consistency rule only in JSON-LD 1.1;
- reject relative `@vocab` mappings in JSON-LD 1.0;
- reject lists of lists in JSON-LD 1.0; and
- report recursive remote-context inclusion during scoped-context validation.

### Scoped contexts, type maps, and invalid bases

Cases: `tc013` and `tli12`.

Type-map values are expanded without first applying the type key's scoped context. Previous-context rollback is also applied too broadly. Separately, an invalid absolute-form `@base` aborts expansion or reuses a stale base instead of leaving relative values unresolved so invalid RDF terms can be discarded.

### Expansion context option

Case: `te077`.

The public parser and W3C harness do not expose the JSON-LD `expandContext` option.

### RDF direction output modes

[...]

PR #1902, opened 2026-08-28
Fix JSON-LD scoped contexts and expansion options

eljeffeg:
Addresses the supported subset of #1901.

## Summary

This fixes three JSON-LD expansion and toRDF conformance gaps in `oxjsonld` and its W3C harness:

- apply a type key's scoped context before expanding its type-map value, while limiting previous-context rollback to `@id` and `@type` maps;
- recover from an invalid absolute-form `@base` without reusing a stale base, allowing invalid relative RDF terms to be discarded; and
- support the JSON-LD API `expandContext` option through `JsonLdParser::with_expand_context` and the W3C harness.

The one-time expansion-context initialization uses a dedicated converter state, so the ordinary parsing path does not check an unused option for every JSON event.

The `oxjsonld` README now documents two intentional support boundaries confirmed in #1901:

- `JsonLdProcessingMode::JsonLd1_0` gates JSON-LD 1.1-only features but is not a strict backward-compatibility mode for JSON-LD 1.0 algorithm edge cases; and
- RDF direction uses native RDF 1.2 directional literals with the `rdf-12` feature and is otherwise ignored; the older `i18n-datatype` and `compound-literal` representations are not supported.

## Conformance results

Corpora:

- JSON-LD API: `289ebf37c30846ba2dab8a33559cc4d348161e0d`
- JSON-LD streaming: `948d24ce66a129734a97efba01cb5d6d8552f4f5`

| Suite | Before | After | Newly executed and passing |
| --- | ---: | ---: | ---: |
| JSON-LD toRDF | 446/467 | 449/467 | 3 |
| Streaming JSON-LD toRDF | 456/483 | 459/483 | 3 |

Newly passing standard cases: `tc013`, `te077`, and `tli12`.

Newly executed and passing streaming cases: `tc013`, `te077`, and the previously stale skip `te068`.

## Scope boundaries

The JSON-LD 1.0 compatibility and alternate RDF-direction cases remain explicitly skipped and documented as intentional non-goals. This PR also does not broaden `oxrdf` or the streaming parser's buffering contract:

- generalized-RDF cases require blank-node predicates that cannot be represented by the current `oxrdf::Quad` API; and
- remaining streaming cases require buffering for late `@id`, `@type`, or graph identity.

## Compatibility and performance

The new parser option is additive. When no expansion context is supplied, parser state and event processing remain on the existing default path.

## Verification

The following pass on branch head `9f325e1`:

```console
cd lib/oxjsonld
cargo test --all-targets
cargo test --all-targets --all-features
cargo clippy --all-targets --all-features -- -D warnings -D clippy::all
rustup run 1.87.0 cargo test --all-targets

cd ../..
cargo test -p oxigraph-testsuite --test parser --all-features
cargo test -p oxigraph-testsuite --test parser jsonld_to_rdf_ --no-default-features
cargo fmt --all --check
git diff --check
```

### DIFF
--- a/lib/oxjsonld/src/context.rs
+++ b/lib/oxjsonld/src/context.rs
@@ -86,6 +86,7 @@ pub struct JsonLdLoadDocumentOptions {
 }
 
 /// Returned information about a remote JSON-LD document or context.
+#[derive(Clone)]
 pub struct JsonLdRemoteDocument {
     /// The retrieved document
     pub document: Vec<u8>,
@@ -318,12 +319,17 @@ impl JsonLdContextProcessor {
                         }
                         // 5.7.3) and 5.7.4)
                         JsonNode::String(value) => {
+                            let has_absolute_iri_form =
+                                IriRef::parse_unchecked(value.clone()).is_absolute();
                             match self.resolve_iri(
                                 value,
                                 result.base_iri.as_ref(),
                                 JsonLdErrorCode::InvalidBaseIri,
                             ) {
                                 Ok(base_iri) => result.base_iri = Some(base_iri),
+                                // Invalid IRIs are discarded during RDF conversion. Do not
+                                // fall back to the previous base for relative values.
+                                Err(_) if has_absolute_iri_form => result.base_iri = None,
                                 Err(e) => errors.push(e),
                             }
                         }
@@ -1499,7 +1505,7 @@ pub fn is_keyword(value: &str) -> bool {
     )
 }
 
-fn json_slice_to_node(data: &[u8]) -> Result<JsonNode, JsonSyntaxError> {
+pub(crate) fn json_slice_to_node(data: &[u8]) -> Result<JsonNode, JsonSyntaxError> {
     let mut parser = SliceJsonParser::new(data);
     json_node_from_events(std::iter::from_fn(|| match parser.parse_next() {
         Ok(JsonEvent::Eof) => None,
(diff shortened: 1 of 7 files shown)


lib/oxjsonld/src/context.rs
```diff
@@ -318,12 +319,17 @@ impl JsonLdContextProcessor {
...
                             match self.resolve_iri(
                                 value,
                                 result.base_iri.as_ref(),
                                 JsonLdErrorCode::InvalidBaseIri,
                             ) {
                                 Ok(base_iri) => result.base_iri = Some(base_iri),
+                                // Invalid IRIs are discarded during RDF conversion. Do not
+                                // fall back to the previous base for relative values.
+                                Err(_) if has_absolute_iri_form => result.base_iri = None,
                                 Err(e) => errors.push(e),
```
Tpt, line 333: Out of curiosity, why the `if has_absolute_iri_form` constraint and why hiding the error here? I fear this is just hiding the problem but not fixing it.

`IriRef::parse_unchecked(value.clone()).is_absolute()` is very unreliable on invalid IRIs and if the IRI is valid we are not in this code path anyway.

lib/oxjsonld/src/expansion.rs
```diff
@@ -49,6 +49,9 @@ pub enum JsonLdValue {
 }
 
 enum JsonLdExpansionState {
+    Initialize {
+        expand_context: JsonLdRemoteDocument,
```
Tpt, line 53: If you make this
```suggestion
        expand_context: Option<JsonLdRemoteDocument>,
```

then the initial state of the parser can always be `Initialize` and it would avoid generating the first `Element` state in two different code places (just gate `apply_expand_context` with `expand_context` being `Some`).

Tpt (commented):
Thank you!

Tpt:
I allowed myself to be bold and extracted to focused fixes from this patch #1931 #1932

PR #1904, opened 2026-08-28
Fix SPARQL slice cardinality estimates

### LINKED ISSUE #1903: SPARQL joins can panic when a subquery OFFSET exceeds LIMIT
## Description

A valid SPARQL query can panic while join ordering estimates the cardinality of a sliced subquery when `OFFSET` is greater than `LIMIT`.

The slice estimator currently computes `limit - offset`. `LIMIT` is the maximum number of rows returned after applying `OFFSET`, so the estimate should instead subtract the offset from the input cardinality and then cap the remainder at the limit.

This also means slices with an `OFFSET` but no `LIMIT` are currently overestimated because the offset is not applied at all.

## Version and platform

- Reproduced on current `main` at `ec68e3ddb2e73470ae5940c44709e041762afa41`
- The equivalent calculation is also present in the `v0.5.9` release
- macOS 26.6.2, arm64
- `rustc 1.97.1`

## Reproduction

Create a store with one triple:

```console
oxigraph update \
  --location /tmp/oxigraph-slice-repro \
  --update 'INSERT DATA { <urn:s> <urn:p> <urn:o> }'
```

Then execute a join containing a sliced subquery:

```console
oxigraph query \
  --location /tmp/oxigraph-slice-repro \
  --query 'SELECT * WHERE { { SELECT ?x WHERE { VALUES ?x { 1 2 } } OFFSET 2 LIMIT 1 } VALUES ?y { 1 } }' \
  --results-format tsv
```

## Expected behavior

The query succeeds with an empty result set.

## Actual behavior

Debug builds and builds with overflow checks enabled exit with a panic:

```text
thread 'main' panicked at lib/sparopt/src/optimizer.rs:1354:28:
attempt to subtract with overflow
```

The panic reproduced consistently. As controls, the sliced subquery succeeds when executed without a join, and changing it to `OFFSET 1 LIMIT 1` returns the expected row.

With overflow checks disabled, the subtraction wraps and supplies an incorrect cardinality estimate to join ordering.

## Relevant code

`estimate_query_expression_size` handles `QueryExpression::Slice` as:

```rust
min(inner, *limit - *offset)
```

The evaluator correctly implements the operation as `skip(offset)` followed by `take(limit)`.

mattfaltyn:
Closes #1903.

## Summary

- Apply `OFFSET` to the estimated input cardinality with saturating subtraction.
- Cap the remaining cardinality with `LIMIT` instead of subtracting `OFFSET` from `LIMIT`.
- Cover ordinary limiting, `OFFSET > LIMIT`, and OFFSET-without-LIMIT estimates.

## Why

The slice estimator used `limit - offset`, even though SPARQL applies OFFSET to the input and then LIMIT to the remaining rows. When OFFSET exceeded LIMIT, debug builds and builds with overflow checks enabled panicked during join ordering. Builds without overflow checks used a wrapped, incorrect cardinality estimate. Slices without a LIMIT also failed to account for OFFSET.

## Validation

- [x] `cargo test --workspace --exclude sparql-smith`
- [x] `cargo test -p sparopt slice_size_applies_offset_before_limit`
- [x] `cargo clippy --all-targets -- -D warnings -D clippy::all` from `lib/sparopt`
- [x] `cargo fmt -- --check`
- [x] Re-ran the issue reproduction through `oxigraph query`; it now succeeds with the expected empty result set.

No documentation changes are needed because this corrects an internal optimizer estimate without changing the SPARQL interface.

### DIFF
--- a/lib/sparopt/src/optimizer.rs
+++ b/lib/sparopt/src/optimizer.rs
@@ -1349,9 +1349,9 @@ fn estimate_query_expression_size(
             offset,
             limit,
         } => {
-            let inner = estimate_query_expression_size(inner, input_types);
+            let inner = estimate_query_expression_size(inner, input_types).saturating_sub(*offset);
             if let Some(limit) = limit {
-                min(inner, *limit - *offset)
+                min(inner, *limit)
             } else {
                 inner
             }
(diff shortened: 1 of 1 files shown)


Tpt (approved):
thanks!

PR #1906, opened 2026-08-29
Adds basic AGENTS.md with some general testing instructions

Tpt:


Tpt:
@mattfaltyn @eljeffeg I guess you are both using LLMs. Feel free to suggest extra things to add in this file.

PR #1908, opened 2026-08-29
Reject incomplete SPARQL boolean results

### LINKED ISSUE #1907: SPARQL result parsers accept truncated boolean documents
**Describe the bug**

`QueryResultsParser` accepts truncated SPARQL ASK result documents in both JSON and XML. It returns the parsed boolean before validating that the top-level JSON object or XML `<sparql>` element is closed.

This can turn an interrupted or truncated response into an authoritative `true` or `false` result instead of reporting a syntax error. The same underlying parser state is used by the slice, reader, and async reader APIs.

**To Reproduce**

1. Oxigraph `main` at `703d49be43ab166967f67ee836a36facf7dabf07`, macOS arm64, Rust 1.97.1.
2. Run this against the workspace `sparesults` crate:

```rust
use sparesults::{QueryResultsFormat, QueryResultsParser, SliceQueryResultsParserOutput};

fn check(label: &str, format: QueryResultsFormat, input: &str) {
    match QueryResultsParser::from_format(format).for_slice(input) {
        Ok(SliceQueryResultsParserOutput::Boolean(value)) => {
            println!("{label}: accepted boolean={value}")
        }
        Ok(SliceQueryResultsParserOutput::Solutions(_)) => println!("{label}: accepted solutions"),
        Err(error) => println!("{label}: rejected: {error}"),
    }
}

fn main() {
    check(
        "json-truncated",
        QueryResultsFormat::Json,
        r#"{"head":{},"boolean":true"#,
    );
    check(
        "xml-truncated",
        QueryResultsFormat::Xml,
        r#"<sparql xmlns="http://www.w3.org/2005/sparql-results#"><head/><boolean>true</boolean>"#,
    );
}
```

Actual output:

```text
json-truncated: accepted boolean=true
xml-truncated: accepted boolean=true
```

Both inputs omit the required closing delimiter for the top-level result document.

The same premature return also accepts non-whitespace content after an otherwise complete document, for example `{"head":{},"boolean":true} false` or a second element after XML `</sparql>`.

**Expected behavior**

Truncated documents and documents with trailing content should return syntax errors. A boolean result should only be returned after the complete JSON object or XML `<sparql>` element and document EOF have been validated.

**Additional context**

Complete JSON and XML ASK documents parse successfully as a control. The behavior is also present in the latest v0.5.9 release.

mattfaltyn:
## Summary

- wait for the complete JSON result object and EOF before returning an ASK boolean
- wait for `</sparql>` and EOF before returning an XML ASK boolean
- reject truncated and trailing JSON/XML documents while preserving JSON member-order flexibility

The parsers previously returned as soon as they read the JSON boolean token or XML `</boolean>` tag. This left the enclosing result document unread, so incomplete responses and trailing content were accepted as authoritative boolean results.

Closes #1907.

## Testing

- `cargo +1.98.0 fmt -- --check`
- `cargo +1.98.0 test --features rdf-12`
- `cargo +1.98.0 test --all-targets --features async-tokio,sparql-12` (`lib/sparesults`)
- `cargo +1.98.0 clippy --all-targets -- -D warnings -D clippy::all` (`lib/sparesults`)
- `cargo +1.98.0 clippy --all-targets --features async-tokio,sparql-12 -- -D warnings -D clippy::all` (`lib/sparesults`)
- `cargo +1.98.0 test -p oxigraph-testsuite --test oxigraph oxigraph_sparql_results_testsuite`

### DIFF
--- a/lib/sparesults/src/json.rs
+++ b/lib/sparesults/src/json.rs
@@ -462,7 +462,7 @@ enum JsonInnerReaderState {
     },
     AfterBindings,
     BeforeBoolean,
-    AfterRoot(bool),
+    AfterRoot,
     Ignore {
         level: usize,
         after: JsonInnerReaderStateAfterIgnore,
@@ -513,8 +513,14 @@ impl JsonInnerReader {
                         Ok(None)
                     }
                     "results" => {
-                        self.state = JsonInnerReaderState::BeforeResults;
-                        Ok(None)
+                        if self.boolean.is_some() {
+                            Err(QueryResultsSyntaxError::msg(
+                                "SPARQL JSON results must not contain both 'boolean' and 'results' keys",
+                            ))
+                        } else {
+                            self.state = JsonInnerReaderState::BeforeResults;
+                            Ok(None)
+                        }
                     }
                     "boolean" => {
                         self.state = JsonInnerReaderState::BeforeBoolean;
@@ -529,8 +535,8 @@ impl JsonInnerReader {
                     }
                 },
                 JsonEvent::EndObject => {
-                    if let Some(value) = self.boolean {
-                        self.state = JsonInnerReaderState::AfterRoot(value);
+                    if self.boolean.is_some() {
+                        self.state = JsonInnerReaderState::AfterRoot;
                         Ok(None)
                     } else {
                         Err(QueryResultsSyntaxError::msg(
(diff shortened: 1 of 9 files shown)


lib/sparesults/src/json.rs
```diff
@@ -461,6 +462,7 @@ enum JsonInnerReaderState {
     },
     AfterBindings,
     BeforeBoolean,
+    AfterRoot(bool),
```
Tpt: nit: This is weird to have the `boolean` both as a parameter of the state here and as a global state in the struct. What about keeping it only as a global state or put it in both `AfterRoot` and `InRootObject` (the later with `Option<bool>`)?

lib/sparesults/src/json.rs
Tpt, line 515: maybe check here that `self.boolean` is none and fail if it's not to prevent using both `"boolean"` and `"results"` keys?

Tpt (commented):
thank you! Two minor nits

lib/sparesults/src/json.rs
mattfaltyn: > This is weird to have the `boolean` both as a parameter of the state here and as a global state in the struct. What about keeping it only as a global state or put it in both `AfterRoot` and `InRootObject` (the later with `Option<bool>`)?

Thanks, good point. I kept the boolean only in `JsonInnerReader::boolean` and made `AfterRoot` a marker state. This is cleaner and avoids duplicated state.

lib/sparesults/src/json.rs
mattfaltyn: > maybe check here that `self.boolean` is none and fail if it is not to prevent using both `"boolean"` and `"results"` keys?

Good catch. I added that guard and a negative syntax test for a document containing both keys. Thank you.

Tpt (approved):
thank you!

PR #1916, opened 2026-08-29
Add borrowed callbacks for N-Triples and N-Quads

eljeffeg:
## Summary

- add callback-based `NTriplesParser::parse_reader` and `NQuadsParser::parse_reader` APIs yielding callback-scoped `TripleRef` and `QuadRef` values
- borrow plain lexer strings and reuse a semantic arena between statements instead of materializing owned RDF terms for every callback
- use the same line-format recognizer for the existing owned iterators and the new borrowed callbacks
- support RDF 1.2 triple terms, including nesting, while allocating only their required owned `Triple` subtrees
- preserve the existing owned iterator API, validation, recovery, and error behavior

## Motivation

Streaming consumers often encode or index parsed RDF immediately and do not need to retain an owned intermediate `Triple` or `Quad`. A callback-scoped borrowed API allows those consumers to process validated RDF directly while keeping parser memory bounded.

The higher-ranked callback lifetime prevents references into the parser's reusable buffers from escaping. Consumers that need to retain a statement can call `into_owned` inside the callback.

`TermRef::Triple` contains `&Triple`, so recursively borrowed triple terms cannot be represented by the existing `oxrdf` reference model. The borrowed line parser therefore keeps ordinary term strings in its reusable arena and temporarily owns only quoted-triple subtrees. This avoids an `oxrdf` API break and keeps the common path allocation-light.

## Validation

- `cargo clippy -p oxttl --all-targets --all-features -- -D warnings`
- `cargo clippy -p oxttl --all-targets --no-default-features -- -D warnings`
- `cargo test -p oxttl --all-features`
- `cargo test -p oxttl --no-default-features`
- `cargo test -p oxigraph-testsuite`
- one-minute `nquads`, `trig`, and `n3` fuzz runs
- `cargo doc -p oxttl --all-features --no-deps`
- borrowed callback and existing owned parser benchmarks

On the same development machine and N-Triples corpus, the shared borrowed callback path parses at roughly 405 MiB/s versus roughly 269 MiB/s for the owned iterator.

### DIFF
--- a/lib/oxttl/src/lexer.rs
+++ b/lib/oxttl/src/lexer.rs
@@ -13,7 +13,7 @@ use std::str;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum N3Token<'a> {
-    IriRef(OxString),
+    IriRef(N3String<'a>),
     PrefixedName {
         prefix: &'a str,
         local: OxStr<'a>,
@@ -21,8 +21,8 @@ pub enum N3Token<'a> {
     },
     Variable(OxString),
     BlankNodeLabel(&'a str),
-    String(OxString),
-    LongString(OxString),
+    String(N3String<'a>),
+    LongString(N3String<'a>),
     Integer(&'a str),
     Decimal(&'a str),
     Double(&'a str),
@@ -33,6 +33,41 @@ pub enum N3Token<'a> {
     },
     Punctuation(&'a str),
     PlainKeyword(&'a str),
+}
+
+pub enum N3String<'a> {
+    Borrowed(&'a str),
+    Owned(OxString),
+}
+
+impl std::fmt::Debug for N3String<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.as_str().fmt(f)
+    }
+}
+
+impl PartialEq for N3String<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_str() == other.as_str()
+    }
+}
+
+impl Eq for N3String<'_> {}
+
+impl N3String<'_> {
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Borrowed(value) => value,
+            Self::Owned(value) => value,
+        }
+    }
+
+    pub fn into_owned(self) -> OxString {
+        match self {
+            Self::Borrowed(value) => OxString::new_owned(value),
+            Self::Owned(value) => value,
+        }
+    }
 }
 
 #[derive(Eq, PartialEq)]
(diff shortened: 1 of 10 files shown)


Tpt:
Thank you for this! I need a bit of time to "digest" a bit the MR.

> The existing N-Triples benchmark corpus parses at roughly 420 MiB/s through the borrowed callback API versus 265 MiB/s through

Great to read. So, moving into this direction might be worth it.

Overall, I find it sad to not support triple terms and having duplicated parser code. I would love to avoid both of these. I will try to figure out a way to fix both of these.

eljeffeg:
Thanks for the direction. I have updated the branch to address both concerns:

- N-Triples and N-Quads now use the same generic line recognizer for owned and borrowed output; the duplicated recognizer has been removed.
- The borrowed callbacks now support RDF 1.2 triple terms, including nested triple terms.

I kept `oxrdf` source-compatible. `TermRef::Triple` currently contains `&Triple`, so the parser must give a triple term an addressable owned representation. The implementation therefore borrows ordinary term strings from a reusable arena and boxes/materializes only quoted-triple subtrees. Boxing also keeps the ordinary `TermDescriptor` small, so the common path does not inherit the size of an owned `Triple`.

I considered changing the `oxrdf` reference model for fully recursive borrowing. A direct `TermRef::Triple(TripleRef)` is recursively sized because `TripleRef` contains a `TermRef`; solving that cleanly requires an indirection or a new handle/accessor-based recursive view and would be a broader, breaking model decision. My preference is to keep that discussion separate unless you would like this parser API to depend on such a redesign.

The rebased update is formatted, Clippy-clean with and without default features, passes the `oxttl` and `oxigraph-testsuite` suites, and completed the required one-minute `nquads`, `trig`, and `n3` fuzz runs.

eljeffeg:
CI note: the completed matrix is green for Clippy, rustdoc, semver, Codecov, Codspeed, N-Quads/TriG/N3 fuzzing, sanitizer, macOS, Windows, WASI, and the remaining test jobs. The single red `test_linux_i686` job is an unrelated failure in `test_bulk_load_rollback`: RocksDB left a temporary `bulk-*.sst` file visible when the test compared directory contents. All earlier i686 library/parser steps in that job passed. GitHub does not allow this fork contributor to rerun the upstream job, so a maintainer rerun should be sufficient.

lib/oxttl/src/lexer.rs
```diff
@@ -35,6 +35,41 @@ pub enum N3Token<'a> {
     PlainKeyword(&'a str),
 }
 
+pub enum N3String<'a> {
```
Tpt: I think you can just use a `OxStr<'a>` instead. `OxStr::new` allows to build a `OxStr` from a slice.

PR #1920, opened 2026-08-31
Deduplicate merged SPARQL default graphs

### LINKED ISSUE #1919: Merged default graphs double-count shared triples
## Describe the bug

When two graphs share a triple, querying their merged default graph counts that triple once per source graph. This changes `COUNT` and `SUM`, and `INSERT ... USING` can persist an inflated aggregate.

Reproduced on `main` at `5e77d401e548ccb299eff5dc2b5f8c6a2ec3ceec` (Oxigraph `0.6.0-dev`), macOS 26.6.2 / ARM64, Rust 1.97.1, using the in-memory store. The corresponding iterator code is also present in `v0.5.10`; I have not separately run this reproduction on that release.

## To reproduce

Start with an empty store and execute:

```sparql
INSERT DATA {
  GRAPH <urn:g1> { <urn:s> <urn:p> 10 }
  GRAPH <urn:g2> { <urn:s> <urn:p> 10 }
}
```

Then query it:

```sparql
SELECT (COUNT(*) AS ?count) (SUM(?o) AS ?total)
FROM <urn:g1>
FROM <urn:g2>
WHERE { ?s <urn:p> ?o }
```

Expected: `count = 1`, `total = 10`.
Actual: `count = 2`, `total = 20`.

With the same store, this update also stores `20` instead of `10`:

```sparql
INSERT { <urn:report> <urn:total> ?total }
USING <urn:g1>
USING <urn:g2>
WHERE {
  SELECT (SUM(?o) AS ?total)
  WHERE { ?s <urn:p> ?o }
}
```

The equivalent query configured with `prepared.dataset_mut().set_default_graph_as_union()` also returns `count = 2`, `total = 20`.

Controls:
- A single `FROM <urn:g1>` correctly returns `1` and `10`.
- `SELECT (COUNT(*) AS ?count) WHERE { GRAPH ?g { ?s <urn:p> ?o } }` correctly returns `2`, because the named graph matches are separate. This behavior should remain unchanged.
- Distinct triples that happen to have the same object must remain distinct; deduplicating projected values or aggregate arguments would not fix the root cause.

## Expected semantics and likely cause

[SPARQL Query §13.2.1](https://www.w3.org/TR/sparql11-query/#specifyingDataset) defines the default graph for multiple `FROM` clauses as the RDF merge of the source graphs. The same ground triple therefore appears once in that graph.

In [`EvalDataset::internal_quads_for_pattern`](https://github.com/oxigraph/oxigraph/blob/5e77d401e548ccb299eff5dc2b5f8c6a2ec3ceec/lib/spareval/src/eval.rs#L165-L192), both the multi-graph and union-default paths erase `graph_name` and concatenate matches without deduplicating the resulting triples. This occurs before solution mappings and aggregates are evaluated.

I checked open and closed issues/PRs and their comments. #525 concerns CONSTRUCT output duplication, #1835 concerns repeated HTTP dataset parameters, and #1905 concerns GRAPH-variable scoping; this reproduction does not depend on those paths.

A fix is proposed in #1920, with regression coverage for merged graph queries, the union-default API, `USING` updates, and the Python/JavaScript bindings. It also covers legitimate duplicate join/projection results and cancellation after the first result in both merge modes. Separate named-graph matches and read-error propagation are preserved. The PR links this issue for closure on merge.

mattfaltyn:
## Problem

When default graphs are merged by multiple `FROM` or `USING` clauses, a triple shared by the source graphs is currently evaluated once per graph. This inflates query aggregates and can persist incorrect totals through `INSERT ... USING`. The union-default-graph API has the same behavior.

Closes #1919.

## Changes

- Reuse `hash_deduplicate` after clearing the source graph name in the two merged-default-graph iterator paths. Equality uses the full triple, before projection or aggregation.
- Preserve the single-graph fast path, named-graph matches, and dataset error propagation. No storage format, dependency, or configuration changes.
- Add file-based query and update regressions with four source quads forming three distinct triples: `COUNT = 3`, `SUM = 40`. Two distinct triples share the value `10`, so deduplicating values would fail the tests. The update also verifies that the source graphs remain unchanged.
- Exercise union deduplication in its API documentation example, add read-error coverage for both merge paths, and update the changelog.
- Extend the existing Python and JS binding tests with overlapping graphs. Correct the JS default-graph-list test that previously expected the duplicated triple to produce two rows.
- Verify that a self-join preserves four identical projected values from distinct matches, with and without optimization.
- Verify streaming cancellation after the first result in both merge modes, and clarify the explicit graph-list API semantics.

Merged scans now retain a hash set of the unique matching triples for the iterator's lifetime. This adds memory proportional to those matches; single-graph queries are unaffected.

## Validation

Confirmed before the fix: both file-based regressions fail (`COUNT = 4`, `SUM = 50`, persisted total `50`) and the union API example returns an extra solution.

Completed validation on `d035917d`, macOS ARM64 (Rust 1.97.1, CPython 3.13.13, Node 24.19.0):

- `cargo test --offline --locked --workspace --features rdf-12`: 650 tests/doctests passed, none failed or ignored.
- `cargo test --offline --locked -p oxigraph-testsuite`: complete file-based conformance suite passed, including the query with and without optimization.
- `cargo test --offline --locked -p spareval`: 14 unit tests and 18 doctests passed.
- `cargo +nightly-2026-04-16 fuzz run <target> --sanitizer none -- -max_total_time=60`: both required targets passed (75,843 `sparql_query_eval` cases; 21,035 `sparql_update_eval` cases).
- `uv run --locked -m unittest` in `python/tests`: 204 passed. `npm test -- --run` in `js`: 43 passed.
- Strict `cargo clippy -p spareval --all-targets -- -D warnings -D clippy::all` passed for default features and separately with `sparql-12`, `sep-0002`, and `sep-0006`.
- `cargo fmt -- --check`, `git diff --check`, JS Biome, Python Ruff, and Python type checks after CI's stub-generation step passed.
- `RUSTDOCFLAGS="-D warnings" cargo doc --offline --locked -p spareval --no-deps --all-features` passed.
- The original store reproduction now returns the corrected merged/union results and persisted total, while single-graph and separate named-graph controls remain unchanged.

Follow-ups `c0a2736d` and `68a56b15` add only tests and API documentation; the runtime fix is unchanged from the fully validated revision above. The new join regression rejects the original implementation (ten rows instead of five), and the cancellation regression rejects eager buffering. Both temporary mutations were restored before validation. The final cancellation fixture leaves a distinct logical result pending after the first row.

Follow-up targeted regression checks, both required one-minute fuzz targets (151,569 query cases; 42,599 update cases), formatting, Python lint/types, and JS Biome passed. The final evaluator test code compiles. A fresh complete local rerun was blocked by macOS native executable startup before test code, including a serial retry; that interrupted run is **not counted as passed**.

The previous revision passed all 45 GitHub checks. See Checks for CI on the current revision.

### DIFF
--- a/lib/oxigraph/src/storage/rocksdb.rs
+++ b/lib/oxigraph/src/storage/rocksdb.rs
@@ -503,6 +503,68 @@ impl<'a> RocksDbStorageReader<'a> {
         }
     }
 
+    pub fn quads_for_pattern_in_union(
+        &self,
+        subject: Option<&EncodedTerm>,
+        predicate: Option<&EncodedTerm>,
+        object: Option<&EncodedTerm>,
+        graph_names: Option<&[Option<EncodedTerm>]>,
+    ) -> RocksDbMergedDecodingQuadIterator<'a> {
+        let order = TripleOrder::from_pattern(subject, predicate, object);
+        let iters = if let Some(graph_names) = graph_names {
+            graph_names
+                .iter()
+                .map(|graph_name| {
+                    let graph_name = graph_name.as_ref().unwrap_or(&EncodedTerm::DefaultGraph);
+                    self.quads_for_pattern(subject, predicate, object, Some(graph_name))
+                })
+                .collect()
+        } else {
+            vec![
+                self.quads_for_pattern(
+                    subject,
+                    predicate,
+                    object,
+                    Some(&EncodedTerm::DefaultGraph),
+                ),
+                self.quads_for_pattern_in_named_graphs(subject, predicate, object),
+            ]
+        };
+        RocksDbMergedDecodingQuadIterator::new(iters, order)
+    }
+
+    fn quads_for_pattern_in_named_graphs(
+        &self,
+        subject: Option<&EncodedTerm>,
+        predicate: Option<&EncodedTerm>,
+        object: Option<&EncodedTerm>,
+    ) -> RocksDbChainedDecodingQuadIterator<'a> {
+        RocksDbChainedDecodingQuadIterator::new(match subject {
+            Some(subject) => match predicate {
+                Some(predicate) => match object {
+                    Some(object) => {
+                        self.spog_quads(&encode_term_triple(subject, predicate, object))
+                    }
+                    None => self.spog_quads(&encode_term_pair(subject, predicate)),
+                },
+                None => match object {
+                    Some(object) => self.ospg_quads(&encode_term_pair(object, subject)),
+                    None => self.spog_quads(&encode_term(subject)),
+                },
+            },
+            None => match predicate {
+                Some(predicate) => match object {
+                    Some(object) => self.posg_quads(&encode_term_pair(predicate, object)),
+                    None => self.posg_quads(&encode_term(predicate)),
+                },
+                None => match object {
+                    Some(object) => self.ospg_quads(&encode_term(object)),
+                    None => self.spog_quads(&[]),
+                },
+            },
+        })
+    }
+
     pub fn quads(&self) -> RocksDbChainedDecodingQuadIterator<'a> {
         RocksDbChainedDecodingQuadIterator::pair(self.dspo_quads(&[]), self.gspo_quads(&[]))
     }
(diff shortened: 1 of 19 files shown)


Tpt:
Thank you for this! Relying on `hash_deduplicate` is a bit unfortunate because it means we keep all emitted quads into memory (not great if the underlying dataset is large). I think we can be smarter by leveraging the facts that some storages are ordered (like the RocksDB one) but it won't always work. Not sure if we can do better with the current architecture query evaluation architecture...

So, my main questioning, is: is the slow down worth it to get a bit more conformant evaluation?

mattfaltyn:
> “I think we can be smarter by leveraging the facts that some storages are ordered”

Great suggestion—implemented in 2df68535. RocksDB now uses an ordered streaming merge with memory bounded by source graphs; generic and in-memory datasets keep the hash fallback. Added on-disk coverage for all index orders.

lib/spareval/src/dataset.rs
```diff
@@ -42,6 +42,23 @@ pub trait QueryableDataset<'a>: Sized + 'a {
         graph_name: Option<Option<&Self::InternalTerm>>,
     ) -> impl Iterator<Item = Result<InternalQuad<Self::InternalTerm>, Self::Error>> + use<'a, Self>;
 
+    /// Fetches quads from the RDF merge of the given graphs according to a pattern.
+    ///
+    /// `None` for `graph_names` encodes the union of all graphs. Implementations may override
+    /// this method when they can compute the merge more efficiently than hash deduplication.
+    /// Returned quads must have no graph name and must not contain duplicate triples.
+    fn internal_quads_for_pattern_in_union(
```
Tpt: Thank you! I am not sure this is the best signature for this method:
- `internal_quads_for_pattern_in_union` seems a bit unclear name to me. It actually returns triples (we don't care about the graph name and deduplicate ignoring it). What about naming it `internal_triples_for_pattern` or `merged_internal_triples_for_pattern` (the merge referring to the RDF graph merge that should be applied here).
- I would make the method return a new `InternalTriple` type instead of `InternalQuad` to highlight that we don't care about the named graph.
- I would try to provide a real default implementation here instead of returning `None`, just like the other optional methods of the trait.

What do you think about all of these? I feel with this that the method would be a bit more intuitive to users and a bit harder to wrongly implement.

If you feel it's going a bit too much, glad to take over this MR.

lib/oxigraph/src/storage/rocksdb.rs
```diff
@@ -888,6 +950,116 @@ impl Iterator for RocksDbChainedDecodingQuadIterator<'_> {
...
+}
+
+impl Iterator for RocksDbMergedDecodingQuadIterator<'_> {
+    type Item = Result<EncodedQuad, StorageError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            for (head, iter) in self.heads.iter_mut().zip(&mut self.iters) {
+                if head.is_none() {
+                    *head = iter.next().map(|quad| {
```
Tpt: hyper-nit: I guess we can detect `Some(Err())` here and avoid the loop line 1032.

Tpt (commented):
Thank you for this work! I just have an API design question.

mattfaltyn:
> “What about naming it `internal_triples_for_pattern`”
>
> “I would make the method return a new `InternalTriple` type”
>
> “I would try to provide a real default implementation here”

Great suggestions—implemented all three in 75027d28, along with the iterator error fast path. Thank you!

Tpt (approved):
Thank you!

PR #1953, opened 2026-09-17
sparopt: don't use a for-loop join for zero-length property paths

### LINKED ISSUE #1952: [sparopt] for-loop join with zero-length property paths returns extra rows
With `sep-0006` enabled (which `oxigraph` and `spareval` do by default), the optimizer can rewrite a join into a `Lateral` for-loop join when the right side is a property path whose variables are bound by the left side (`reorder_joins` in `lib/sparopt/src/optimizer.rs`). For a path that can match zero-length, that correlated evaluation returns rows that evaluating the path independently and joining does not.

Repro, no data needed (an empty dataset is enough):

```sparql
SELECT * WHERE {
  () a? ?b .
  ?b a | ^(a* / a*) ?1
}
```

With optimizations: one solution, `?b = rdf:nil`, `?1 = rdf:nil`.
With `without_optimizations()`: no solutions.

`?b` is bound to `rdf:nil` by the zero-length `a?` from the ground `()`. `rdf:nil` does not occur in the graph, so `?b a | ^(a* / a*) ?1` evaluated independently has no match: a zero-length path only matches terms of the graph when both ends are variables. The generated `Lateral` evaluates the path with `?b` already bound, so `^((a*)/(a*))` matches `rdf:nil` to itself and returns a row.

Root cause: `is_path_fit_for_for_loop_join` (`lib/sparopt/src/optimizer.rs:1175-1209`) only rejects a zero-length path when the zero-length node's immediate endpoints are bound. When the zero-length node is nested in `Seq`/`Alt`/`Inv`, the check is bypassed: the `Seq` arm uses a fresh `#intermediate#` variable and `||`, so `Seq(ZeroOrMorePath, ZeroOrMorePath)` under `Inv` is considered fit even though the far endpoint is bound.

I found this with a differential test on `spareval` (optimized vs `without_optimizations`) driven by `sparql-smith`: 3,000,000 generated queries, 4,149 of them containing `GRAPH`, all agreeing until minimization produced the query above. I'll open a fix.

tunglambk:
Fixes #1952.

`is_path_fit_for_for_loop_join` only rejected a zero-length path when the `ZeroOrMorePath`/`ZeroOrOnePath` node's own endpoints were bound. When the node is nested in a sequence or an alternative, the endpoint bound by the left side of the join isn't one of the node's endpoints, so the path was still evaluated as a for-loop join.

That changes results: a zero-length match binds the input term to itself even when it does not appear in the graph, so the correlated evaluation returns rows that evaluating the path independently and joining does not.

The new check runs on the whole path before the recursive part, so paths that cannot match zero-length still take the for-loop join. The cost is that joins where a zero-length-capable path has a bound endpoint now use a regular hash join instead.

Test: `testsuite/oxigraph-tests/sparql-optimization/zero_length_path_for_loop_join_*` (the optimizer emitted `lateral` before the fix). `cargo test -p oxigraph-testsuite --test sparql`, `cargo test -p oxigraph-testsuite --test oxigraph oxigraph_optimizer_testsuite`, `cargo test -p sparopt`, `cargo test -p spareval`, `cargo fmt --all -- --check` and `cargo clippy -p sparopt -p spareval --all-targets -- -D warnings` all pass.


tunglambk:
The CodSpeed regression is on `BSBM explore 1000 queryAndUpdate in memory without optimizations`, which runs through `SparqlEvaluator::without_optimizations()`, so `Optimizer::optimize_query_expression` is never called and this change isn't on that path. I also dumped the optimized plan for all 635 BSBM and Sparqloscope benchmark queries before and after the change and they're byte-identical, so I read the delta as measurement noise rather than a cost of the fix.

lib/sparopt/src/optimizer.rs
```diff
@@ -1177,22 +1177,41 @@ fn is_path_fit_for_for_loop_join(
...
     object: &TermPattern,
     entry_types: &VariableTypes,
+) -> bool {
+    // A zero-length path match binds the input term to itself, even if it does not
+    // appear in the graph. The for-loop join evaluates the path with the input
+    // bindings, so it would then return matches that a regular join, which evaluates
+    // the path independently, does not have.
+    if can_match_zero_length(path)
+        && (is_term_pattern_bound(subject, entry_types)
+            || is_term_pattern_bound(object, entry_types))
```
Tpt: This is a bit aggressive: if `subject` or `object` is a constant, the optimization is still safe I think

Tpt:
Thank you very much for this! I think I found a narrower path that should fix the issue while allowing to rewrite more queries: #1960 Is it fine with you if I ship this one instead?

tunglambk:
Absolutely, please ship yours — #1960 is the better change and I would rather see the narrower one land.

I checked it covers what my branch was after. On `e0f286b0` I compared optimized evaluation against `without_optimizations` on the same dataset across thirteen shapes: my branch's own query, a zero-length path with a bound endpoint, a bound endpoint plus a constant object (including the `:p*/:p*` form), a `Link` with two variable endpoints, and the nested alternative and sequence shapes. All thirteen agree, where the pre-fix commit `22864b20` had two disagreements. My branch's regression test also passes unchanged on `e0f286b0`, so there is nothing left for it to add — its only extra precision was letting a few more correct `OneOrMore` laterals through.

So I am closing this in favour of #1960. Thanks for picking it up and for taking the narrower route.

One thing worth a look: #1952 is still open and its timeline has no reference to #1960, so the merge may not have closed it.

PR #1955, opened 2026-09-17
sparopt: reorder the LATERAL right side built from a join with the left side bindings

### LINKED ISSUE #1951: Join reordering inside generated `Lateral` ignores outer bindings in `UNION` and filtered groups
# Join reordering inside generated `Lateral` ignores outer bindings in `UNION` and filtered groups

## Describe the bug

When a `JOIN` is rewritten into a `Lateral`, join reordering within a `UNION` branch (or a grouped pattern with a `FILTER`) appears to ignore variables bound by the left-hand side of the `Lateral`.

As a result, when two patterns in a branch have the same estimate, the optimizer may choose the first-written pattern even if it does not use variables already bound by the outer query. This can cause a full scan to be repeated once per left-side row.

```sparql
PREFIX ex: <http://example.com/>
SELECT * WHERE {
  ?a ex:p ?c .
  { ?s ex:q ?o . ?s ex:r ?c . }
  UNION
  { ?c ex:q ?o . }
}
```

With 1,000 `?a ex:p ?c` triples, 1,000 `ex:q` triples, and 1,000 `ex:r` triples, the query above takes about 350 ms on pyoxigraph 0.5.11. Reversing the order of the first branch:

```sparql
?s ex:r ?c .
?s ex:q ?o .
```

returns the same 1,000 solutions in about 2 ms.

The difference appears to be that `?s ex:q ?o` is scanned once for every bound value of `?c`, producing roughly 1,000,000 intermediate results, whereas evaluating `?s ex:r ?c` first restricts the scan to the already-bound `?c` values.

The same behavior occurs for grouped patterns with a `FILTER`. An explicit `LATERAL` does not exhibit the problem.

## To Reproduce

```python
import time
import pyoxigraph as ox

ex = lambda name: ox.NamedNode(f"http://example.com/{name}")
store = ox.Store()

store.extend(
    quad
    for i in range(1000)
    for quad in (
        ox.Quad(ex(f"a{i}"), ex("p"), ex(f"c{i}")),
        ox.Quad(ex(f"s{i}"), ex("q"), ex(f"o{i}")),
        ox.Quad(ex(f"s{i}"), ex("r"), ex(f"c{i}")),
    )
)

query = "PREFIX ex: <http://example.com/> SELECT * WHERE { ?a ex:p ?c . { %s } UNION { ?c ex:q ?o . } }"

for branch in (
    "?s ex:q ?o . ?s ex:r ?c .",
    "?s ex:r ?c . ?s ex:q ?o .",
):
    start = time.perf_counter()
    rows = len(list(store.query(query % branch)))
    print(f"{branch}  {rows} rows  {(time.perf_counter() - start) * 1000:.1f} ms")
```

Output:

```text
?s ex:q ?o . ?s ex:r ?c .  1000 rows  350.9 ms
?s ex:r ?c . ?s ex:q ?o .  1000 rows    2.0 ms
```

## Expected behavior

The optimizer should reorder patterns within each branch using variables bound by the left side of the generated `Lateral`, just as it already does for explicit `LATERAL` and, on `main`, for `LeftJoin`/`OPTIONAL`.

Both query orders should therefore produce similar plans and runtimes.

## Explain plan evidence

Slow order:

```text
Lateral [1000]
  QuadPattern(?a ex:p ?c) [1000]
  Union [1000]
    Lateral [1000]
      QuadPattern(?s ex:q ?o) [1000000]
      QuadPattern(?s ex:r ?c) [1000]
```

Fast order:

```text
Lateral [1000]
  QuadPattern(?a ex:p ?c) [1000]
  Union [1000]
    Lateral [1000]
      QuadPattern(?s ex:r ?c) [1000]
      QuadPattern(?s ex:q ?o) [1000]
```

The slow plan performs a full scan of `?s ex:q ?o` for each left-side row, while the fast plan evaluates the pattern that uses the already-bound `?c` first.

## Regression

The behavior appears to have been introduced by commit `6b243ab` (`sparopt: properly optimize joins recursively`, #1888).

Before that commit, both query orders produced similar runtimes. After it, the branch is reordered into a `Lateral`, but the generated `Lateral` appears not to be reordered again using the variables bound by its left side.

## Possible cause

From reading `lib/sparopt/src/optimizer.rs`:

- `Optimizer::reorder_joins` recursively reorders join children using the incoming `input_types`.
- When a child is converted into `GraphPattern::lateral(output, next)`, `next` does not appear to be reordered again using the newly available `output_types`.
- The explicit `Lateral` optimization path does call `reorder_joins(..., &left_types)`.
- #1733 added equivalent handling for `LeftJoin`, which appears to be why `OPTIONAL` is fixed on `main`.

Reordering the generated `Lateral`'s right-hand side using the variables bound by its left-hand side seems likely to eliminate the issue.

## Environment

- pyoxigraph 0.5.11
- Reproduced on 0.5.10 and current `main`
- Linux x86_64

## Related

#1884 and #1888: the empty-key join inside a UNION that 6b243ab fixed.
#1733: the same ordering fix for OPTIONAL, on main.
#1883: another per-row rescan, OPTIONAL under GROUP BY, answered as fixed by #1733.
#1418: store statistics for join reordering. Statistics would change the estimates, but not which variables the right side of the Lateral counts as bound.

kiki286:
Fixes #1951

When `reorder_joins` turns a join child into the right side of a `Lateral`, the child had only been reordered with the join input variables (since #1888 made the reordering recursive). Inside a `UNION` branch, a group with a `FILTER` or a group with a `BIND`, a pattern that does not use the variables bound by the left side can then be evaluated first and fully scanned once per left row. Explicit `LATERAL` and, since #1733, `OPTIONAL` already reorder their right side with the left side variables.

This reorders the child again with the variables bound by the left side when it becomes the right side of a `Lateral`, keeping the previous plan if that makes it unfit for a for loop join (e.g. it turns into a cartesian product).

Doing only that makes join reordering exponential in the nesting depth of such groups, because every nested child gets reordered twice per level (16 nested filtered groups: 0.4 ms -> 280 ms). So `reorder_joins` gets a `for_estimation` flag: the children plans used to take the greedy decisions are computed as before, without the second pass, and only the plan kept for each child is fully reordered. Decisions are the same as before; it stays polynomial (depth 20: 0.7 ms -> 4 ms). I'm happy to rework this part if you would prefer another way to avoid the blowup.

Tests: new `union_lateral_reordering`, `filter_lateral_reordering` and `nested_union_lateral_reordering` optimizer tests. The `nested_bind_lateral_reordering` expected output changes: `?t` is bound by the left side, so `?a :hasT ?t` is now evaluated before `?a a :A`.

On the #1951 data (in-memory store): UNION 317 ms -> 1.6 ms, filtered group 384 ms -> 1.4 ms, same solutions. WatDiv basic and BSBM explore plans are unchanged. Comparing `main` and this branch on about 1,200 benchmark queries and about 2,900 randomly generated nested queries gave identical results.


kiki286:
Force-pushed an amended commit: same fix, same plans, less planning work.

I measured the CodSpeed report locally under callgrind (4 runs per commit). The flagged `without optimizations` benchmark moved +0.11% against its own 1.08% run-to-run spread, and it does not call the optimizer at all; that benchmark is noisy because bulk loading renames blank nodes with random ids. But the run did show a real cost the report listed as untouched: BSBM business intelligence +2.9%, all of it the optimizer planning parts of the query twice.

The amended commit skips the second pass where it provably cannot give another plan, which removes 78% of that: business intelligence is now +0.65% against main, and `reorder_joins` on those 300 queries goes 44.1M (main) -> 64.0M -> 48.4M instructions. No plan changes: 0 differences over the 685 patterns those benchmarks plan and 4,198 further queries, and the optimizer test expectations are unchanged.

PR #1957, opened 2026-09-19
Add Store::snapshot, a pinned read view that takes no lock

fabio-rovai:
`Store` already snapshots internally for every read: `len`, `is_empty`,
`contains` and `quads_for_pattern` each call `self.storage.snapshot()`. So two
successive reads can observe two different states, and there is no way to ask
for one state and hold it across several reads.

`start_transaction` pins a state, but on the memory backend it costs every
writer: `Store::insert` opens a transaction of its own, and
`MemoryStorage::start_transaction` waits for an open one, so a concurrent
`insert` does not return until the reader commits. On the RocksDB backend a
writer proceeds. The same code therefore behaves differently depending on how
the store was opened, and a long read stops an in-memory store outright.

This exposes what `Storage::snapshot` already does: a read view pinned to the
state at the instant it was taken, holding no lock.

```rust
let snapshot = store.snapshot();
store.insert(quad)?;             // committed afterwards
assert_eq!(snapshot.len()?, 1);  // and not seen through the snapshot
```

`StoreSnapshot` offers `len`, `is_empty`, `contains`, `iter` and
`quads_for_pattern`, mirroring the equivalent `Store` methods, plus
`IntoIterator for &StoreSnapshot` to match `IntoIterator for &Store`.

### Implementation note

The reader is shared through an `Arc` rather than cloned per call.
`rocksdb_wrapper::Reader` owns a `*mut rocksdb_readoptions_t` and frees it in
`Drop`, so deriving `Clone` on it would double free. One reader, one `Drop`,
shared with the iterators it hands out.

### Tests

Two, in `lib/oxigraph/tests/store.rs`:

- `test_snapshot_is_pinned_to_one_state` — the view does not move under a
  commit made after it was taken, through `len`, `contains`, `iter` and
  `quads_for_pattern`, while the store itself does see it.
- `test_snapshot_does_not_block_writers` — a writer thread completes while a
  snapshot is open. This is the property `start_transaction` does not have on
  the memory backend, and it is the reason for the addition.

`cargo test -p oxigraph` passes, 105 tests. No existing behaviour changes; this
only adds a method and a type.

### Motivation

Found while building a certificate-emitting reasoner that reads a graph many
times and then has to state which graph it read. A copy of the selected data
costs ~221 bytes a triple; holding a transaction costs every writer the length
of the run; a snapshot costs neither. Happy to adjust naming or the method set
if you would rather it were shaped differently.


lib/oxigraph/src/store.rs
```diff
@@ -858,6 +895,96 @@ impl IntoIterator for &Store {
...
+/// Returned by [`Store::snapshot`]. Every read on it sees the state the store
+/// was in when it was taken, whatever has been committed since. It holds no
+/// lock, so writers are not blocked while it lives.
+///
+/// The reader is behind an `Arc` rather than cloned per call on purpose:
+/// `rocksdb_wrapper::Reader` owns a `*mut rocksdb_readoptions_t` and frees it
+/// in `Drop`, so a derived `Clone` would double free. One reader, one `Drop`,
+/// shared ownership with the iterators it hands out.
+#[derive(Clone)]
+pub struct StoreSnapshot {
```
Tpt: Nit: I would name it `Snapshot` for consistency with `Transaction`

```suggestion
pub struct Snapshot {
```

lib/oxigraph/src/store.rs
```diff
@@ -858,6 +895,96 @@ impl IntoIterator for &Store {
...
+/// A read view of a [`Store`] pinned to one state.
+///
+/// Returned by [`Store::snapshot`]. Every read on it sees the state the store
+/// was in when it was taken, whatever has been committed since. It holds no
+/// lock, so writers are not blocked while it lives.
+///
+/// The reader is behind an `Arc` rather than cloned per call on purpose:
+/// `rocksdb_wrapper::Reader` owns a `*mut rocksdb_readoptions_t` and frees it
+/// in `Drop`, so a derived `Clone` would double free. One reader, one `Drop`,
+/// shared ownership with the iterators it hands out.
```
Tpt: I think is wrong, `rocksdb_wrapper::Reader` `Clone` implementation properly copy the `*mut rocksdb_readoptions_t` type. So, no need for an `Arc` (but please correct me if it's me who is actually wrong)

lib/oxigraph/src/store.rs
```diff
@@ -339,6 +339,43 @@ impl Store {
         self.storage.snapshot().is_empty()
     }
 
+    /// Take a read snapshot of the store, pinned to the state at this instant.
+    ///
+    /// Added for open-ontologies. `Store` already snapshots internally for every
```
Tpt: Please rephrase the comment to not be about your project, the other Oxigraph users (and myself) likely do not care much

lib/oxigraph/src/store.rs
```diff
@@ -858,6 +895,96 @@ impl IntoIterator for &Store {
...
+        graph_name: Option<&GraphName>,
+    ) -> SnapshotQuadIter {
+        SnapshotQuadIter {
+            iter: self.reader.quads_for_pattern(
+                subject.map(EncodedTerm::from).as_ref(),
+                predicate.map(EncodedTerm::from).as_ref(),
+                object.map(EncodedTerm::from).as_ref(),
+                graph_name.map(EncodedTerm::from).as_ref(),
+            ),
+            reader: Arc::clone(&self.reader),
```
Tpt: I don't think keeping the snapshot object around is useful, the `rocksdb_wrapper` code already takes care of that. Hence, you can just return the existing `QuadIter` type

lib/oxigraph/src/store.rs
Tpt, line 240: to factorize code I think you can just do `self.snapshot().quads_for_pattern(subject, predicate, object, graph_name)` here (same for the other methods)

Tpt (commented):
Thank you for this! This is very welcome. Some ideas to simplify the code

lib/oxigraph/src/store.rs
fabio-rovai: Renamed to `Snapshot` in d30289b.

lib/oxigraph/src/store.rs
fabio-rovai: You are right. `impl Clone for Reader` at rocksdb_wrapper.rs copies the options with `oxrocksdb_readoptions_create_copy` and the pinned snapshot sits in an `Arc<SnapshotReader>`, so a clone is already safe. I had read the `Drop` and missed the manual `Clone`. The `Arc` is gone in d30289b: `StorageReader`, `StorageReaderKind` and `RocksDbStorageReader` now derive `Clone`, and the snapshot holds a plain `StorageReader<'static>`.

lib/oxigraph/src/store.rs
fabio-rovai: Rewritten in d30289b. The doc now only says what the method does and how it differs from `start_transaction`.

lib/oxigraph/src/store.rs
fabio-rovai: Done in d30289b. `SnapshotQuadIter` is removed and `Snapshot::quads_for_pattern` returns `QuadIter<'static>` with a cloned reader, the same shape `Store::quads_for_pattern` already had.

lib/oxigraph/src/store.rs
fabio-rovai: Done in d30289b. `quads_for_pattern`, `contains`, `len` and `is_empty` on `Store` now go through `self.snapshot()`, so the encoding of the pattern lives in one place. `cargo test -p oxigraph` is still at 105 passing, clippy clean.

Tpt (approved):
Thank you!

